PR #7, opened 2025-01-28
Add ElectricalCellRecording (ExperimentalTrace before) and import data

AurelienJaquier:
- remove label from LicenseCreate: we don't need both a name and a label
- add acronym to brain region
- new schema for subject
- convenience schema for base data model and single cell model
- new schema for file
- new schema for stimulus and trace
- new schema for mechanism


jdcourcol:
@AurelienJaquier can I have access to this document ?

app/schemas/electrical_cell_recording.py
```diff
@@ -0,0 +1,114 @@
...
+    BrainRegionRead,
+    CreationMixin,
+    IdentifiableMixin,
+    LicensedCreateMixin,
+    LicensedReadMixin,
+    SubjectRead,
+)
+
+
+class StimulusCreate(BaseModel):
```
eleftherioszisis, line 20: I haven't added those yet. Is dt the only thing that defines a stimulus? It doesn't seem to carry much information as it is structured here.

app/schemas/electrical_cell_recording.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+# Trace Model for Electrophysiology Data
+class TraceCreate(SingleCellData):
+    file: File = Field(..., title="File", description="File associated with the trace.")
+    stimuli: StimulusCreate = Field(
+        ...,
+        title="Stimuli",
+        description="List of stimuli applied to the cell with their respective time steps",
+    )
+    comment: str | None = Field(
```
eleftherioszisis, line 91: Is comment necessary? Could "description" be used instead?

app/schemas/electrical_cell_recording.py
```diff
@@ -0,0 +1,114 @@
...
+        "", title="Comment", description="Additional information or observations"
+    )
+    derivation: int | None = Field(
+        None,
+        title="Derivation ID",
+        description="The ID of the cell or model the trace was derived from",
+    )
+    image: list[File] | None = Field(
```
eleftherioszisis, line 99: Images will be a directory asset and will not be listed explicitly here.

app/schemas/electrical_cell_recording.py
eleftherioszisis: @AurelienJaquier from the ontology resource you sent me stimuli are defined as:
```
    {
      "@id": "http://bbp.epfl.ch/neurosciencegraph/ontologies/stimulustypes/A",
      "@type": "Class",
      "atlasRelease": {
        "@id": "https://bbp.epfl.ch/neurosciencegraph/data/4906ab85-694f-469d-962f-c0174e901885",
        "@type": "BrainAtlasRelease",
        "_rev": 12
      },
      "label": "A",
      "subClassOf": [
        "http://bbp.epfl.ch/neurosciencegraph/ontologies/stimulustypes/CurrentClampStimulus"
      ]
    },
```
Would it be sufficient to use an Enum with all possible stimulus types, e.g. something like:
```python
class StimulusCreate:
    id: uuid.UUID
    protocol: StimulusProtocolEnum
    dt: float
```

AurelienJaquier: Having an Enum with all possible stimuli looks good to me. However, we have to keep in mind that the users will probably use their own custom stimuli that are not in the Enum. Which means that we will have to update the Enum along the way, as the users ask us to integrate their custom protocols into the database.
I hope this is ok from the engineering side. We have planned to do the same kind of thing for other fields in other schemas, like protocol_design in morphology_method.

app/schemas/sub_cellular_model_script.py
```diff
@@ -0,0 +1,71 @@
...
+            ),
+        ),
+    ]
+    is_temperature_dependent: Annotated[
+        bool,
+        Field(
+            title="is temperature dependent",
+            description="Whether the mechanism changes depending on temperature or not",
+        ),
+    ] = True
```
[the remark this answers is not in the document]
AurelienJaquier: Most of the ion channel mod files are not temperature dependent (they are not using any q10 factor in their equations), so it would be better to keep False as the default value

app/schemas/electrical_cell_recording.py
AurelienJaquier: I had a quick chat with @darshanmandge about this. We think that it would be best to keep this comment field, in case the experimentalist have some small comments about their recordings that are not important enough to be in the description, but that are still worth keeping in the metadata.

app/schemas/sub_cellular_model_script.py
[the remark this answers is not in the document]
eleftherioszisis: Does that mean that temperature is optional if most ion channel mod files are not temperature dependent, or is it always defined regardless if the channel uses it?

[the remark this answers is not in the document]
AurelienJaquier: Temperature in the schema is optional since most of the ion channel are not temperature dependent.
When we want to run a cell though, we will have to set a temperature, but it will not affect the ion channels that are temperature independant.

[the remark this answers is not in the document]
eleftherioszisis: Sure, setting a temperature when simulating a cell, which is ignored by ion channels that are not temperature dependent makes perfect sense. 

I'll make temperature optional so that it is only specified for temperature dependent ion channels.

eleftherioszisis:
I added stimuli as separate resources with a protocol string attribute. Depending on how these stimuli are going to be used it may worth it to investigate storing them as JSONB. That will also make registering easier instead of having to create the main resource and then create N stimuli for that resource.

I looked into the possibility of creating a stimulus protocol unique table, but the names of protocols are local to the resources (have to map to the nwb file) and cannot be easily curated because the h5 files needs to be changed too. For example the list of protocols I collected from nexus are:

```python
['A', 'ADHPdepol', 'ADHPhyperpol', 'ADHPrest', 'APDrop', 'APResh',
       'APThreshold', 'APWaveform', 'C1HP1sec', 'C1_HP_0.5sec',
       'C1_HP_1sec', 'C1step_1sec', 'C1step_ag', 'C1step_highres',
       'CalOU01', 'CalOU04', 'DeHyperPol', 'Delta', 'ElecCal',
       'FirePattern', 'H10S8', 'H20S8', 'H40S8', 'HighResThResp',
       'HyperDePol', 'IDRest', 'IDRestTest', 'IDThres', 'IDdepol',
       'IDhyperpol', 'IDrest', 'IDthresh', 'IRdepol', 'IRhyperpol',
       'IRrest', 'IV', 'IV2', 'IV_Test', 'LoOffset1', 'LoOffset3',
       'NegCheops', 'NegCheops1', 'NegCheops2', 'NegCheops3',
       'NegCheops4', 'NegCheops5', 'NoiseOU3', 'NoisePP', 'NoiseSpiking',
       'OU10Hi01', 'OU10Lo01', 'OU10Me01', 'PosCheops', 'Pulse', 'RPip',
       'RSealClose', 'RSealOpen', 'Rac', 'ResetITC', 'Rin', 'S2', 'S30',
       'SAPThres1', 'SAPThres2', 'SAPThres3', 'SAPTres1', 'SAPTres2',
       'SAPTres3', 'SAPWaveform', 'SDelta', 'SIDRest', 'SIDThres',
       'SIDTres', 'SIV', 'SIne20Hz', 'SNoisePP', 'SNoiseSpiking', 'SRac',
       'SResetITC', 'SSineSpec', 'SSpikeRec', 'SSponAPs', 'SSponHold',
       'SSponNoHold', 'STesteCode', 'STrueNoise', 'SetAmpl', 'SetISI',
       'SetISITest', 'SineSpec', 'SpikeRec', 'SpikeRec_Ih',
       'SpikeRec_Kv1.1', 'SponAPs', 'SponHold25', 'SponHold3',
       'SponHold30', 'SponNoHold20', 'SponNoHold30', 'SpontAPs',
       'Spontaneous', 'SsAHP', 'StartHold', 'StartNoHold',
       'StartStandeCode', 'Step', 'SubWhiteNoise', 'TestAmpl', 'TestRheo',
       'TestSpikeRec', 'Test_eCode', 'TesteCode', 'Truenoise',
       'VacuumPulses', 'WhiteNoise', 'maria-STEP', 'pulser', 's2', 'sAHP',
       'scope', 'spuls', 'step_1', 'step_2', 'step_3']
```
Where there are many synonyms, mistypes, test protocols, etc. For that reason I opted for creating the stimuli as resources with protocol as a string and the only constraint is for those protocols to be consistent with the file inside the resource.

jdcourcol:
wow, this is really messy. I don't know how the user can then use that later on for optimization if that is not curated. @darshanmandge

darshanmandge:
We have a [mapping](https://github.com/openbraininstitute/BluePyEModel/blob/main/bluepyemodel/ecode/__init__.py) in BluePyEModel for different ecodes to specify what type of protocol it is

jdcourcol:
@darshanmandge can you make a proposal to curate that list ? and what we should have at the end of the day in the database.
Also, if the file have to be curated as well, do you have a code example for that ?

darshanmandge:
Ok. Does anyone here have file names for the obscure protocols `STrueNoise`, `spuls`, etc., which I can look into to understand what they are? I also have an ecodes description document somewhere. I will check that to

AurelienJaquier:
@darshanmandge these files gives you the names of the files using each protocol for different projects. 
[bbp_lnmce_stim_dict.json](https://github.com/user-attachments/files/19583005/bbp_lnmce_stim_dict.json)
[public_hippocampus_stim_dict.json](https://github.com/user-attachments/files/19583006/public_hippocampus_stim_dict.json)
[public_sscx_stim_dict.json](https://github.com/user-attachments/files/19583007/public_sscx_stim_dict.json)
[public_thalamus_stim_dict.json](https://github.com/user-attachments/files/19583008/public_thalamus_stim_dict.json)

app/cli/utils.py
```diff
@@ -80,10 +82,18 @@ def get_or_create_species(species, db, _cache={}):
...
-        protocol=data["label"],
+def create_stimulus(data, entity_id, project_context, db):
+    row = ElectricalRecordingProtocol(
+        name=data["label"],
+        description=data.get("definition", None),
+        dt=None,
+        stimulus_injection_type=ElectricalRecordingStimulusType.unknown,
+        stimulus_shape=ElectricalRecordingStimulusShape.unknown,
+        stimulus_start_time=None,
+        stimulus_end_time=None,
```
eleftherioszisis, line 93: For importing the data we can either infer these from the nexus resource or leave them undefined to be updated later.

app/cli/utils.py
```diff
@@ -80,10 +82,18 @@ def get_or_create_species(species, db, _cache={}):
     return sp.id
 
 
-def create_stimulus(data, entity_id, db):
-    row = ElectricalRecordingStimulus(
-        protocol=data["label"],
+def create_stimulus(data, entity_id, project_context, db):
+    row = ElectricalRecordingProtocol(
+        name=data["label"],
```
eleftherioszisis, line 87: Not sure if it's a good idea to use the name for the protocol name and restrict that with an enum instead of using a dedicated attribute. On the other hand it will be searchable.

eleftherioszisis:
@darshanmandge, @AurelienJaquier concerning the curated list of stimulus protocol names and corresponding code for changing the nib, is there an update?

jdcourcol:
> @darshanmandge, @AurelienJaquier concerning the curated list of stimulus protocol names and corresponding code for changing the nwb, is there an update?

@darshanmandge consider that as a priority as it is in the critical path to deliver entitycore

app/db/types.py
```diff
@@ -61,3 +62,33 @@ class SingleNeuronSimulationStatus(StrEnum):
...
+
+
+class ElectricalRecordingType(StrEnum):
+    intracellular = auto()
+    extracellular = auto()
+    both = auto()
+    unknown = auto()
+
+
+class ElectricalRecordingStimulusType(StrEnum):
```
jdcourcol, line 79: should we use an ontology here, like https://ontobee.org/ontology/OBI?iri=http://purl.obolibrary.org/obo/OBI_0002182

app/db/model.py
```diff
@@ -349,6 +347,24 @@ class Entity(TimestampMixin, Root):
     }
 
 
+class Subject(SpeciesMixin, Entity):
+    __tablename__ = "subject"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    age: Mapped[int | None]
```
jdcourcol, line 353: age is special. I did not tackle that one as it can be P14, or weeks, or whatever.

app/db/model.py
```diff
@@ -514,12 +530,49 @@ class Role(LegacyMixin, TimestampMixin, Base):
...
+    __tablename__ = "electrical_cell_recording"
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
     name: Mapped[str] = mapped_column(index=True)
-    description: Mapped[str]
-    __mapper_args__ = {"polymorphic_identity": "single_cell_experimental_trace"}  # noqa: RUF012
+    description: Mapped[str] = mapped_column(default="")
+    recordingType: Mapped[ElectricalRecordingType]
+    recordingLocation: Mapped[STRING_LIST]
+    ljp: Mapped[float] = mapped_column(default=0.0)
+    comment: Mapped[str] = mapped_column(default="")
```
jdcourcol, line 568: which should put a comment explaining what is a comment here. or be more explicit.

app/db/model.py
```diff
@@ -514,12 +530,49 @@ class Role(LegacyMixin, TimestampMixin, Base):
...
+    Entity,
+):
+    __tablename__ = "electrical_cell_recording"
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
     name: Mapped[str] = mapped_column(index=True)
-    description: Mapped[str]
-    __mapper_args__ = {"polymorphic_identity": "single_cell_experimental_trace"}  # noqa: RUF012
+    description: Mapped[str] = mapped_column(default="")
+    recordingType: Mapped[ElectricalRecordingType]
+    recordingLocation: Mapped[STRING_LIST]
```
jdcourcol, line 566: do we have a format for that ?

eleftherioszisis: I believe recordingLocation can be anything that can be parsed by NEURON, for example ["soma[0]_0.5"].

app/db/model.py
eleftherioszisis: Do we need the comment in general? Isn't description enough?

eleftherioszisis:
@darshanmandge , @AurelienJaquier , is this good to go or do we want to add anything else for this iteration?

app/db/model.py
```diff
@@ -509,11 +513,47 @@ class Role(LegacyMixin, Identifiable):
...
 ):
-    __tablename__ = EntityType.single_cell_experimental_trace.value
+    __tablename__ = EntityType.electrical_cell_recording.value
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    recording_type: Mapped[ElectricalRecordingType]
+    recording_origin: Mapped[ElectricalRecordingOrigin]
+    recording_location: Mapped[STRING_LIST]
+    ljp: Mapped[float] = mapped_column(default=0.0)
+    comment: Mapped[str] = mapped_column(default="")
+
```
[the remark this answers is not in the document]
jdcourcol: the rest can come later.

AurelienJaquier:
This is good to be merged from my point of view

eleftherioszisis:
Is this good to go @jdcourcol ?

jdcourcol:
go ahead

jdcourcol (approved):

PR #9, opened 2025-01-28
basic files API

mgeplf:


docs/API.md
```diff
@@ -115,17 +115,70 @@ GET /brain-regions/{id}: could do this, but not sure it's useful: the full
...
+The later is things like circuit data, which can be large, and may not include only individual files, rather a directory.
+
+## Semantics
+
+* An entity (eg: morphology), can be associated with multiple files
+* A file is associated with only one Entity, this also means that a `Folder` owns all the contents, and no other Entity points to anything within the folder
+* A file has a type associated with it; at first approximation these are MIME types, however there are many vendor specific ones that will have to be handled
+* Once registered, a file is immutable; it cannot be changed or replaced
+* A file inherits its Authorization from the entity to which it is associated
+
```
jdcourcol, line 142: I don't know where to put that comment. I understand an entity can have several asset. The asset pointing to the entity. However, we don't put any information about the relationship between the asset and the entity.
Say I have a morphology and 3 assets which are images. One is XY plane view, one is YZ plane view,...
With the current model, I have 3 assets, mimetype being image pointing to the morphology. We are missing a "property/metadata" that says "xy_plane" or something that helps.

jdcourcol: May be what that means is that the provenance model should consider assets and not entity. ?

docs/API.md
```diff
@@ -115,17 +115,70 @@ GET /brain-regions/{id}: could do this, but not sure it's useful: the full
...
+The former is likely to be smaller files (such as morphologies, ephys traces, images and reports \[ie: pdf\], etc).
+The later is things like circuit data, which can be large, and may not include only individual files, rather a directory.
+
+## Semantics
+
+* An entity (eg: morphology), can be associated with multiple files
+* A file is associated with only one Entity, this also means that a `Folder` owns all the contents, and no other Entity points to anything within the folder
+* A file has a type associated with it; at first approximation these are MIME types, however there are many vendor specific ones that will have to be handled
+* Once registered, a file is immutable; it cannot be changed or replaced
```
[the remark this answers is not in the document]
jdcourcol: I think I am ok with having versions at the file level and not at the entity level.

docs/API.md
```diff
@@ -115,17 +115,70 @@ GET /brain-regions/{id}: could do this, but not sure it's useful: the full
...
+1) Downloading them via a web API
+2) Mounting them in some fashion on the cluster
+
+The former is likely to be smaller files (such as morphologies, ephys traces, images and reports \[ie: pdf\], etc).
+The later is things like circuit data, which can be large, and may not include only individual files, rather a directory.
+
+## Semantics
+
+* An entity (eg: morphology), can be associated with multiple files
+* A file is associated with only one Entity, this also means that a `Folder` owns all the contents, and no other Entity points to anything within the folder
```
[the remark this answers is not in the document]
jdcourcol: I suggest we don't support versioning of entity but we support version of assets:
- entity properties can change
- asset are kept and point to the entity, with a version identifier.
- assets have a "is_version" relationship between them
- we support only incr. versions. 0->1->2->3...
- provenance links can point to asset

docs/API.md
GianlucaFicarelli: 1. type of relationship between the asset and the entity
Yes this is a limitation of the current KG and it usually requires to iterate over all the distributions until finding the expected content-type, but this strategy doesn't work if there are multiple files with the same content-type.
Implementation detail: In the new tables there is a `meta` jsonb column that can be used to store arbitrary metadata.
It seems hard to define a relationships that works for any type of entity, but if we can formalize the possible types of relations, we could also add a dedicated attribute/column.

2. relationships between assets in the same entity
I see the assets just as the artifacts of the entity, and I don't expect relations between the artifacts.
For example, to create a morphology I could generate the ASC file locally, convert it to SWC, validate them locally, and only after that create the entity and upload the files altogether, and in this case there isn't any specific relation between them.
In general, things can become complicated if multiple files are used to generate multiple files, and we want to track all the relations.

PR #11, opened 2025-01-30
remove `/reconstruction_morphology/q/` endpoint

mgeplf:
all GET queries for morphologies, including facet and `?search=term` should match API docs

### DIFF
--- a/app/routers/morphology.py
+++ b/app/routers/morphology.py
@@ -75,8 +75,8 @@ def morphology_query(
     db: SessionDep,
     morphology_filter: Annotated[MorphologyFilter, FilterDepends(MorphologyFilter)],
     search: str | None = None,
-    skip: int = 0,
-    limit: int = 10,
+    page: int = 0,
+    page_size: int = 10,
 ):
     name_to_table = {
         "species": Species,
@@ -87,8 +87,8 @@ def morphology_query(
         query = db.query(ReconstructionMorphology)
         query = morphology_filter.filter(query)
         response = ListResponse[ReconstructionMorphologyRead](
-            data=morphology_filter.sort(query).offset(skip).limit(limit).all(),
-            pagination=Pagination(page=0, limit=0, total=0),
+            data=morphology_filter.sort(query).offset(page * page_size).limit(page_size).all(),
+            pagination=Pagination(page=page, page_size=page_size, total_items=query.count()),
         )
     else:
         facets = {}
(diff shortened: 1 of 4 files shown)


app/routers/types.py
```diff
@@ -0,0 +1,17 @@
...
+M = TypeVar("M", bound=BaseModel)
+
+
+class Pagination(BaseModel):
+    page: int
+    limit: int
+    total: int
+
+
+class ListResponse(BaseModel, Generic[M]):
```
GianlucaFicarelli, line 14: Minor, but the new syntax in python 3.12 can be:
```suggestion
class ListResponse[M: BaseModel](BaseModel):
```
and you don't even need to define the TypeVar above.
Ref: https://docs.python.org/3/whatsnew/3.12.html#pep-695-type-parameter-syntax

app/routers/morphology.py
```diff
@@ -69,62 +69,61 @@ def create_reconstruction_morphology(
...
-        .all()
-    )
-    res = {
-        "data": [ReconstructionMorphologyRead.model_validate(rm) for rm in rms],
-        "facets": facets,
-    }
-    return JSONResponse(content=jsonable_encoder(res))
+        response = ListResponse[ReconstructionMorphologyRead](
+            data=[ReconstructionMorphologyRead.model_validate(rm) for rm in rms],
+            pagination=Pagination(page=0, limit=0, total=0),
```
GianlucaFicarelli, line 126: Are the pagination parameters still placeholders?

app/routers/types.py
mgeplf: neat

app/routers/morphology.py
mgeplf: 🤦 they shouldn't be, but they might need to be in the short term - I don't like taking these shortcuts though.

app/routers/morphology.py
```diff
@@ -87,15 +87,15 @@ def morphology_query(
...
+            pagination=Pagination(page=page, page_size=page_size, total_items=query.count()),
         )
     else:
-        facets = {}
+        facets: Facets = {}
         for ty, table in name_to_table.items():
             types = aliased(table)
             facet_q = (
-                db.query(types.name, func.count().label("count"))  # type: ignore[attr-defined]
+                db.query(types.name, func.count().label("total"))  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
mgeplf: completely agree; I'll add a TODO.

app/routers/types.py
```diff
@@ -0,0 +1,20 @@
+from typing import TypeVar
+
+from pydantic import BaseModel
+
+M = TypeVar("M", bound=BaseModel)
```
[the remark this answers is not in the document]
mgeplf: yup

app/routers/types.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Pagination(BaseModel):
+    page: int
+    page_size: int
+    total_items: int
+
+
+type Facets = dict[str, dict[str, int]]
```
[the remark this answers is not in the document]
mgeplf: yup

GianlucaFicarelli (approved):

PR #12, opened 2025-01-30
add basic /brain-regions/ API that returns the non-hierarchy hierarchy

mgeplf:
* this hierarchy is the on that is currently used on the frontend, but this will be replaced by a more svelte one in the future


mgeplf:
One should be able to get the hierarchy posted in the teams channel using `http://127.0.0.1:8000/brain-region/`; I started to add more features, but mainly to keep the tests running; big questions remain about who can modify the brain region hierarchy, IMO.

app/routers/brain_region.py
```diff
@@ -9,17 +13,63 @@
...
+        response = JSONResponse(content=[tuple(r) for r in get_region_tree(db, start_id=997)])
+    else:
+        # return the old style DKE hierarchy
+        response = Response(content=HIERARCHY, media_type="application/json")
+    return response
+
+
+@router.get("/{id_}", response_model=BrainRegionRead)
+def read_reconstruction_morphology(db: SessionDep, id_: int):
+    rm = db.query(BrainRegion).filter(BrainRegion.id == id_).first()
```
GianlucaFicarelli, line 70: With the new syntax it would be something like:
```suggestion
    import sqlalchemy as sa
    query = sa.select(BrainRegion).filter(BrainRegion.id == id_)
    rm = db.scalars(query).first()
```

app/db/model.py
```diff
@@ -94,9 +96,11 @@ class BrainLocation(Base):
 
 class BrainRegion(TimestampMixin, Base):
     __tablename__ = "brain_region"
+
     id: Mapped[int] = mapped_column(primary_key=True, index=True)
-    ontology_id: Mapped[str] = mapped_column(unique=True, index=True, nullable=False)
     name: Mapped[str] = mapped_column(unique=True, index=True, nullable=False)
+    acronym: Mapped[str] = mapped_column(unique=True, index=True, nullable=False)
+    children: list[int] = Column(ARRAY(Integer), nullable=True)  # type: ignore[assignment]
```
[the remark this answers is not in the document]
mgeplf: seems yes

app/routers/brain_region.py
```diff
@@ -9,17 +13,63 @@
...
-    db_brain_region = BrainRegion(ontology_id=brain_region.ontology_id, name=brain_region.name)
+    db_brain_region = BrainRegion(**brain_region.model_dump())
     db.add(db_brain_region)
     db.commit()
     db.refresh(db_brain_region)
     return db_brain_region
+
+
+@router.get("/")
+def get(db: SessionDep, flat: str | None = None) -> Response:
```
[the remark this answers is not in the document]
mgeplf: yes

app/routers/brain_region.py
mgeplf: I think we should switch all those over at once, rather than doing it piecemeal; I've make a ticket: https://github.com/openbraininstitute/entitycore/issues/13

GianlucaFicarelli: Agreed, I commented this one because it was a new query, but it can be converted later 👍

app/routers/brain_region.py
```diff
@@ -9,17 +13,63 @@
...
-    db_brain_region = BrainRegion(ontology_id=brain_region.ontology_id, name=brain_region.name)
+    db_brain_region = BrainRegion(**brain_region.model_dump())
     db.add(db_brain_region)
     db.commit()
     db.refresh(db_brain_region)
     return db_brain_region
+
+
+@router.get("/")
+def get(db: SessionDep, flat: bool | None = False) -> Response:  # noqa: FBT001, FBT002
```
GianlucaFicarelli, line 56: Minor, but it doesn't need to be optional, as in `flat: bool  = False`

bilalesi (approved):

app/db/model.py
```diff
@@ -94,9 +95,12 @@ class BrainLocation(Base):
 
 class BrainRegion(TimestampMixin, Base):
     __tablename__ = "brain_region"
-    id: Mapped[int] = mapped_column(primary_key=True, index=True)
-    ontology_id: Mapped[str] = mapped_column(unique=True, index=True, nullable=False)
+
+    # See https://github.com/openbraininstitute/core-web-app/blob/cd89893db3fe08a1d2e5ba90235ef6d8c7be6484/src/types/ontologies.ts#L7
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True, index=True)
```
GianlucaFicarelli, line 100: If we change the id from int to bigint, to contain the ontology_id, then all the foreign keys are converted as well.
Cannot we keep the ontology_id column with the values provided externally?
What are the pros and cons of the 2 solutions?
Also, I'm not sure to understand the id at the link above, because it's defined as str and not as int.

app/cli/import-data.py
```diff
@@ -100,20 +108,35 @@ def import_licenses(data, db):
 
 
 def _import_annotation_body(data, db_type_, db):
+    # aa = [(d.get('label', ''), d.get('prefLabel', ''), d.get('definition', '')) for d in data]
+    # breakpoint() # XXX BREAKPOINT
```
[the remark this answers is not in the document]
mgeplf: grr, I need to reenable my precommit hook for that, sorry

app/db/model.py
mgeplf: I'll make the comment better, it's more to say what the shape of the frontend is using.

We could keep the ontology_id, but since it's used as the primary key in nearly every operation, I think it's easier to use it as the `id`; there is the downside of the space being slightly larger, but I don't think that's a big worry.
The simplicity of having it as the id means that one can use the hiearchy.json, and just use the id there as the lookup for the GET endpoint (ie: `/brain-region/{id}`) as opposed to having either two endpoints, or a param (ie: `/brain-region/{id}?by_ontology=True`.

That's my reasoning, anyways.

GianlucaFicarelli: Fair enough, especially for the GET endpoint `/brain-region/{id}`.
Out of scope for this PR, but if I'm not wrong the hierarchy file in nexus was changing during time (using tags).
I wonder if we are reasonably confident that future changes will be backward compatible, and we don't require versioning for this?
If needing versioning, we'll need some changes to the brain_region table, and it could be easier to use an internal id for the FK

mgeplf: That's something I briefly thought about, but since this PR was just to add the `.json` file, and it grew bigger, I was hoping to defer that to later...:(

mgeplf:
FYI the test failure was on GH's end, I just reran it and it worked.

GianlucaFicarelli (approved):

app/db/model.py
eleftherioszisis: Just a note, it's not only one hierarchy changing, but we may also have different hierarchies corresponding to different species (e.g. human, mouse) that may have an overlap of regions but with different corresponding parcellation ids as they would belong to a different voxelized atlas.

PR #14, opened 2025-01-31
Support assets

GianlucaFicarelli:
Still WIP

- tables
```
asset
asset_entity
```
- endpoints
```
GET    /{entity_type}/{entity_id}/assets
POST   /{entity_type}/{entity_id}/assets
GET    /{entity_type}/{entity_id}/assets/{asset_id}
DELETE /{entity_type}/{entity_id}/assets/{asset_id}
GET    /{entity_type}/{entity_id}/assets/{asset_id}/download
POST   /{entity_type}/{entity_id}/assets/upload/initiate
POST   /{entity_type}/{entity_id}/assets/upload/complete
```
- added error handlers
- logic split in:
    - routers: should contain only input validation and the logic to present the result
    - service: should contain the business logic
    - repository: should contain the low level queries. This is handled through the class RepositoryGroup that's injected as a fastapi dependency, so it can be passed in place of of the db dependency.

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1,5 +1,6 @@
 from datetime import datetime
-from typing import Annotated, ClassVar
+from typing import Any, ClassVar
+from uuid import UUID
 
 from sqlalchemy import (
     DateTime,
@@ -7,43 +8,18 @@ from sqlalchemy import (
     MetaData,
     UniqueConstraint,
     func,
-    or_,
 )
-from sqlalchemy.dialects.postgresql import TSVECTOR
+from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR
 from sqlalchemy.orm import DeclarativeBase, Mapped, declared_attr, mapped_column, relationship
-from sqlalchemy.types import VARCHAR, TypeDecorator
 
-
-class StringListType(TypeDecorator):
-    impl = VARCHAR
-    cache_ok = True
-
-    def process_bind_param(self, value, dialect):  # noqa: ARG002, PLR6301
-        if value is not None:
-            return ",".join(value)
-        return None
-
-    def process_result_value(self, value, dialect):  # noqa: ARG002, PLR6301
-        if value is not None:
-            return value.split(",")
-        return None
-
-    @staticmethod
-    def is_equal(column, value):
-        return func.strpos(column, value) > 0
-
-    @staticmethod
-    def in_(column, values):
-        return or_(*[StringList.is_equal(column, value) for value in values])
-
-
-StringList = Annotated[StringListType, "StringList"]
+from app.db.types import AssetStatus, EntityType, StringList, StringListType
 
 
 class Base(DeclarativeBase):
     type_annotation_map: ClassVar[dict] = {
         datetime: DateTime(timezone=True),
         StringList: StringListType,
+        dict[str, Any]: JSONB,
     }
     # See https://alembic.sqlalchemy.org/en/latest/naming.html
     metadata = MetaData(
@@ -512,3 +488,25 @@ class ExperimentalSynapsesPerConnection(LocationMixin, SpeciesMixin, LicensedMix
     name: Mapped[str] = mapped_column(unique=False, index=True, nullable=False)
     description: Mapped[str] = mapped_column(unique=False, index=False, nullable=False)
     __mapper_args__ = {"polymorphic_identity": "experimental_synapses_per_connection"}  # noqa: RUF012
+
+
+class Asset(TimestampMixin, Base):
+    """Asset table."""
+
+    __tablename__ = "asset"
+    id: Mapped[int] = mapped_column(primary_key=True)
+    uuid: Mapped[UUID] = mapped_column(index=True, unique=True)  # for external access
+    status: Mapped[AssetStatus] = mapped_column(nullable=False)
+    path: Mapped[str]
+    meta: Mapped[dict[str, Any]]  # not used yet. can be useful?
+    # TODO: consider other attributes
+
+
+class AssetEntity(Base):
+    """Asset-Entity many-to-many table."""
+
+    __tablename__ = "asset_entity"
+    id: Mapped[int] = mapped_column(primary_key=True)
+    asset_id: Mapped[int] = mapped_column(ForeignKey("asset.id", ondelete="CASCADE"), index=True)
+    entity_id: Mapped[int] = mapped_column(index=True)  # cannot be ForeignKey
+    entity_type: Mapped[EntityType]  # needed to look up the correct entity table
--- a/app/db/types.py
+++ b/app/db/types.py
@@ -1,0 +1,47 @@
+from enum import auto
+from typing import Annotated
+
+from sqlalchemy import func, or_
+from sqlalchemy.types import VARCHAR, TypeDecorator
+
+from app.utils.enum import HyphenStrEnum
+
+
+class StringListType(TypeDecorator):
+    impl = VARCHAR
+    cache_ok = True
+
+    def process_bind_param(self, value, dialect):  # noqa: ARG002, PLR6301
+        if value is not None:
+            return ",".join(value)
+        return None
+
+    def process_result_value(self, value, dialect):  # noqa: ARG002, PLR6301
+        if value is not None:
+            return value.split(",")
+        return None
+
+    @staticmethod
+    def is_equal(column, value):
+        return func.strpos(column, value) > 0
+
+    @staticmethod
+    def in_(column, values):
+        return or_(*[StringList.is_equal(column, value) for value in values])
+
+
+StringList = Annotated[StringListType, "StringList"]
+
+
+class EntityType(HyphenStrEnum):
+    # name (underscore separated): for table names
+    # value (hyphen separated): for endpoints
+    reconstruction_morphology = auto()
+    species = auto()
+    strain = auto()
+    experimental_bouton_density = auto()
+
+
+class AssetStatus(HyphenStrEnum):
+    CREATED = auto()
+    DELETED = auto()
--- a/app/errors.py
+++ b/app/errors.py
@@ -1,0 +1,66 @@
+"""Api exceptions."""
+
+import dataclasses
+from collections.abc import Iterator
+from contextlib import contextmanager
+from enum import auto
+from http import HTTPStatus
+from typing import Any
+
+from sqlalchemy.exc import NoResultFound
+
+from app.utils.enum import UpperStrEnum
+
+
+class ApiErrorCode(UpperStrEnum):
+    """API Error codes."""
+
+    INVALID_REQUEST = auto()
+    ENTITY_NOT_FOUND = auto()
+    ENTITY_FORBIDDEN = auto()
+
+
+@dataclasses.dataclass(kw_only=True)
+class ApiError(Exception):
+    """API Error."""
+
+    message: str
+    error_code: ApiErrorCode
+    http_status_code: HTTPStatus = HTTPStatus.BAD_REQUEST
+    details: Any = None
+
+    def __repr__(self) -> str:
+        """Return the repr of the error."""
+        return (
+            f"{self.__class__.__name__}("
+            f"message={self.message!r}, "
+            f"error_code={self.error_code}, "
+            f"http_status_code={self.http_status_code}, "
+            f"details={self.details!r})"
+        )
+
+    def __str__(self) -> str:
+        """Return the str representation."""
+        return (
+            f"message={self.message!r} "
+            f"error_code={self.error_code} "
+            f"http_status_code={self.http_status_code} "
+            f"details={self.details!r}"
+        )
+
+
+class EventError(Exception):
+    """Raised when failing to process a message in the queue."""
+
+
+@contextmanager
+def ensure_result(error_message: str) -> Iterator[None]:
+    """Context manager that raises ApiError when no results are found after executing a query."""
+    try:
+        yield
+    except NoResultFound as err:
+        raise ApiError(
+            message=error_message,
+            error_code=ApiErrorCode.ENTITY_NOT_FOUND,
+            http_status_code=HTTPStatus.NOT_FOUND,
+        ) from err
--- a/app/repository/asset.py
+++ b/app/repository/asset.py
@@ -1,0 +1,77 @@
+"""Asset repository module."""
+
+from uuid import UUID
+
+import sqlalchemy as sa
+from sqlalchemy import Row, Sequence
+
+from app.db.model import Asset, AssetEntity
+from app.db.types import AssetStatus, EntityType
+from app.repository.base import BaseRepository
+
+
+class AssetRepository(BaseRepository):
+    """AssetRepository."""
+
+    def get_entity_assets(
+        self,
+        entity_type: EntityType,
+        entity_id: int,
+    ) -> Sequence[Row]:
+        """."""
+        query = (
+            sa.select(
+                Asset.uuid,
+                Asset.path,
+                Asset.status,
+            )
+            .select_from(Asset)
+            .join(AssetEntity)
+            .where(
+                AssetEntity.entity_type == entity_type,
+                AssetEntity.entity_id == entity_id,
+            )
+        )
+        return self.db.execute(query).all()
+
+    def get_entity_asset(
+        self,
+        entity_type: EntityType,
+        entity_id: int,
+        asset_id: UUID,
+    ) -> Row | None:
+        query = (
+            sa.select(
+                Asset.uuid,
+                Asset.path,
+                Asset.status,
+            )
+            .select_from(Asset)
+            .join(AssetEntity)
+            .where(
+                AssetEntity.entity_type == entity_type,
+                AssetEntity.entity_id == entity_id,
+                Asset.uuid == asset_id,
+            )
+        )
+        return self.db.execute(query).one_or_none()
+
+    def update_entity_asset_status(
+        self,
+        entity_type: EntityType,
+        entity_id: int,
+        asset_id: UUID,
+        asset_status: AssetStatus,
+    ) -> Asset | None:
+        query = (
+            sa.update(Asset)
+            .values(status=asset_status)
+            .where(
+                AssetEntity.asset_id == Asset.id,
+                AssetEntity.entity_type == entity_type,
+                AssetEntity.entity_id == entity_id,
+                Asset.uuid == asset_id,
+            )
+            .returning(Asset)
+        )
+        return self.db.execute(query).scalar_one_or_none()
--- a/app/repository/asset.py
+++ b/app/repository/asset.py
@@ -3,7 +3,7 @@
 from uuid import UUID
 
 import sqlalchemy as sa
-from sqlalchemy import Row, Sequence
+from sqlalchemy import Sequence
 
 from app.db.model import Asset, AssetEntity
 from app.db.types import AssetStatus, EntityType
@@ -17,36 +17,26 @@ class AssetRepository(BaseRepository):
         self,
         entity_type: EntityType,
         entity_id: int,
-    ) -> Sequence[Row]:
+    ) -> Sequence[Asset]:
         """."""
         query = (
-            sa.select(
-                Asset.uuid,
-                Asset.path,
-                Asset.status,
-            )
-            .select_from(Asset)
+            sa.select(Asset)
             .join(AssetEntity)
             .where(
                 AssetEntity.entity_type == entity_type,
                 AssetEntity.entity_id == entity_id,
             )
         )
-        return self.db.execute(query).all()
+        return self.db.execute(query).scalars().all()
 
     def get_entity_asset(
         self,
         entity_type: EntityType,
         entity_id: int,
         asset_id: UUID,
-    ) -> Row | None:
+    ) -> Asset:
         query = (
-            sa.select(
-                Asset.uuid,
-                Asset.path,
-                Asset.status,
-            )
-            .select_from(Asset)
+            sa.select(Asset)
             .join(AssetEntity)
             .where(
                 AssetEntity.entity_type == entity_type,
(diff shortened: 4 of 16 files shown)


app/errors.py
```diff
@@ -0,0 +1,66 @@
...
+
+from app.utils.enum import UpperStrEnum
+
+
+class ApiErrorCode(UpperStrEnum):
+    """API Error codes."""
+
+    INVALID_REQUEST = auto()
+    ENTITY_NOT_FOUND = auto()
+    ENTITY_FORBIDDEN = auto()
```
mgeplf, line 20: will this be user visable? I think `not found` should be returned for resources that the user doesn't have access to

app/routers/asset.py
```diff
@@ -0,0 +1,96 @@
...
+    return FileResponse(file_path)
+
+
+@router.delete("/{entity_type}/{entity_id}/assets/{asset_id}")
+def delete_entity_asset(
+    repos: RepoGroupDep, entity_type: EntityType, entity_id: int, asset_id: UUID
+) -> AssetRead:
+    """Delete the data and metadata of an assets associated with a specific entity."""
+    # TODO:
+    #  - should we keep the asset record, but mark it as deleted?
```
mgeplf, line 74: good question; we have to consider this problem for the whole API, but for files it's more important, since they can be so large.
For public resources, I think he answer is that they can't be unpublished, as Bilal pointed out, but I suspect there will be exceptional cases where they will have to be removed....so marking as deleted seems to be the way to go.

app/routers/asset.py
```diff
@@ -0,0 +1,96 @@
...
+    #  - should we keep or delete the asset in the datastore?
+    #  - what if the user want to re-upload the same file that was deleted?
```
mgeplf, line 76: what is the definition of `same file`?  even if a file isn't deleted, we probably need to consider the case of same names?

app/routers/asset.py
```diff
@@ -0,0 +1,96 @@
...
+        entity_id=entity_id,
+        proj_id=proj_id,
+        asset_id=asset_id,
+    )
+
+
+@router.post("/{entity_type}/{entity_id}/assets/upload/initiate")
+def initiate_entity_asset_upload(repos: RepoGroupDep, entity_type: EntityType, entity_id: int):
+    """Generate a signed URL with expiration that can be used to upload the file directly to S3."""
+    raise NotImplementedError
```
mgeplf, line 90: if a file/folder  is being `registered`, ie: it already exists in the s3 store, but entitycore needs to become aware of it, is this the entrypoint?

mgeplf (commented):
the points in the PR all make sense to me, I had a quick read through the code, and it all seems to be sensical.  I think there is still the `register from workflow` that needs to be considered, and what the "layout" of the s3 backing store is, since we haven't completely thought this through, imo.

app/errors.py
GianlucaFicarelli: In case of error, the idea is to return a response that contains:
- the http status (for example 404 NOT FOUND)
- a payload with more details for the type of error, specifically:
  - a message (human readable)
  - an application error code (it provides the client more details about the type of error, in case the http status code is too generic)
  - details (not sure yet if needed, it can contain some data to be used by the client/UI, depending on the type of error)

app/routers/asset.py
GianlucaFicarelli: For reference, note also that it's possible to use versioning-enabled bucket on S3, so that a file deleted can be recovered (until the marker is deleted as well)
https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html

app/routers/asset.py
GianlucaFicarelli: I'm thinking to the case where an entity (a morphology for example) is associated to a file (morph.swc for example) and the user wants to overwrite the file morph.swc (for example because the old one was wrong).
If this use case is needed, we can do it in one step (one request), or two steps (delete + post/put).
The file to be replaced can be identified by its uuid, or by its path/filename (we can have a unique key on the path for example)

app/routers/asset.py
GianlucaFicarelli: I think we need a different endpoint for linking resources already on S3, I'll add it later

bilalesi:
suggestion: one thing we can also consider for assets type, is to use the `magic number` to determine the type, not the extension from the filename, this way it's secure (some cases where the user alter `.exe` file to `.pdf` for instance) and very deterministic, except for specific files as `.swc` which is basically text file so in this case we can check the magic number and get the extension from filename

app/routers/asset.py
GianlucaFicarelli: Thinking more about possible use cases, bbp-workflow should be able to move and register the assets using these endpoints (because the assets that were created in scratch need to be moved into a directory on S3 that should be provided by the service).

I'm not sure that we need to allow to register arbitrary files already on S3, because the service cannot control their location in this way. Do we?

app/routers/asset.py
mgeplf: Hmm, but the "owning" entity is staying the same?
I'm not sure replacement like that should be allowed; immutability would suggest that since the bad morph was uploaded, it can't be undone.  I think of it like a `git push --force`; and that we can't have such things - bad data needs a separate new entity to replace it.
Does that make sense?

app/routers/asset.py
mgeplf: IMO, as long as there is a way for bbp-workflow to register folders it produces without having to reupload large amounts of data, then it's fine.

app/routers/asset.py
GianlucaFicarelli: The owning entity would still be the same.
However, if modification should not be allowed, we can forbid replacing a file.
But then should we also forbid deleting specific files of the entity?
We cannot forbid adding new files, though, unless we introduce a concept of released/frozen entity.
In general, when we consider immutability, do we want to apply it to entities, files, or both?

app/schemas/base.py
```diff
@@ -1,6 +1,11 @@
 from datetime import datetime
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import UUID4, BaseModel, ConfigDict
+
+
+class ProjectContext(BaseModel):
+    virtual_lab_id: UUID4
+    project_id: UUID4
```
GianlucaFicarelli, line 8: This is a placeholder, if will be removed after https://github.com/openbraininstitute/entitycore/pull/10 is merged

app/dependencies/auth.py
```diff
@@ -0,0 +1,7 @@
+from typing import Annotated
+
+from fastapi import Header
+
+from app.schemas.base import ProjectContext
+
+ProjectContextHeader = Annotated[ProjectContext, Header()]
```
GianlucaFicarelli, line 7: This is a placeholder, if will be removed after https://github.com/openbraininstitute/entitycore/pull/10 is merged

app/service/asset.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+    # if (
+    #         not result.authorized_public
+    #         and result.authorized_project_id != project_context.project_id
+    # ):
+    #     raise ApiError(
+    #         message="Entity forbidden",
+    #         error_code=ApiErrorCode.ENTITY_FORBIDDEN,
+    #         http_status_code=HTTPStatus.FORBIDDEN,
+    #     )
```
GianlucaFicarelli, line 31: This should be uncommented after https://github.com/openbraininstitute/entitycore/pull/10 is merged

app/utils/s3.py
```diff
@@ -0,0 +1,122 @@
...
+    filename: str,
+    is_public: bool,
+) -> str:
+    """Return the path used to store the file on S3."""
+    vlab_id = str(vlab_id)
+    proj_id = str(proj_id)
+    entity_id_mod = f"{entity_id % 0xFFFF:04x}"
+    path = f"assets/{entity_type}/{entity_id_mod}/{entity_id}/{filename}"
+    if not is_public:
+        path = f"private/{vlab_id[:4]}/{vlab_id}/{proj_id}/{path}"
```
GianlucaFicarelli, line 28: The S3 paths should still be discussed

GianlucaFicarelli (commented):
@mgeplf this is ready for another review. Still missing, but I think they can go into a separate PR:
- finalization of private/public paths on S3
- upload and download of big files through delegation
- registration of directories (it could probably use the same or a similar mechanism, except that it doesn't need to generate a presigned url)

Some code will need minor changes after merging https://github.com/openbraininstitute/entitycore/pull/10

alembic/versions/20250210_171207_91dfdcef513d_.py
```diff
@@ -0,0 +1,78 @@
+"""empty message
```
mgeplf, line 1: I not reviewing this :)

app/config.py
```diff
@@ -32,6 +33,13 @@ class Settings(BaseSettings):
     LOG_CATCH: bool = True
     LOG_STANDARD_LOGGER: dict[str, str] = {"root": "INFO"}
 
+    S3_PRIVATE_BUCKET_NAME: str = "obi-private"
+    S3_PUBLIC_BUCKET_NAME: str = "obi-public"
+    S3_MULTIPART_THRESHOLD: int = 5 * 1024**2  # bytes
```
mgeplf, line 38: how'd you pick this?

app/db/model.py
```diff
@@ -1,50 +1,26 @@
...
-
-StringList = Annotated[StringListType, "StringList"]
+from app.db.types import BIGINT, AssetStatus, StringList, StringListType
 
 
 class Base(DeclarativeBase):
     type_annotation_map: ClassVar[dict] = {
         datetime: DateTime(timezone=True),
         StringList: StringListType,
+        dict[str, Any]: JSONB,
```
mgeplf, line 23: can we add a type alias for `dict[str, Any]` so it's easier to search for its usage? Even calling it JSONSerialize would be helpful imo.

app/db/model.py
```diff
@@ -516,3 +492,29 @@ class ExperimentalSynapsesPerConnection(LocationMixin, SpeciesMixin, LicensedMix
...
+    __tablename__ = "asset"
+    id: Mapped[BIGINT] = mapped_column(primary_key=True)
+    status: Mapped[AssetStatus] = mapped_column(nullable=False)
+    path: Mapped[str] = mapped_column(nullable=False)  # relative path
+    fullpath: Mapped[str] = mapped_column(nullable=False)  # full path on S3
+    bucket_name: Mapped[str]
+    is_directory: Mapped[bool]
+    content_type: Mapped[str]
+    size: Mapped[BIGINT]
+    meta: Mapped[dict[str, Any]]  # not used yet. can be useful?
```
mgeplf, line 509: I suspect you're right that it will be useful, so I guess we keep it for now, even if it's not used - don't want to many schema changes :)

app/repository/asset.py
```diff
@@ -0,0 +1,96 @@
...
+            sa.select(Asset)
+            .join(Root, Root.id == Asset.entity_id)
+            .where(
+                Asset.entity_id == entity_id,
+                Asset.id == asset_id,
+                Asset.status != AssetStatus.DELETED,
+                Root.type == entity_type,
+            )
+        )
+        return self.db.execute(query).scalar_one()
```
mgeplf, line 49: how does this interact w/ the error handling you added yesterday?
ie: you use, `with ensure_result(error_message="..."):` in a GET.
Here is the expectation that the sqlalchemy exception will be auto converted?

app/repository/asset.py
```diff
@@ -0,0 +1,96 @@
...
+        entity_id: int,
+        asset_id: int,
+        asset_status: AssetStatus,
+    ) -> Asset:
+        """Update the status of the given asset.
+
+        Raise an error if any of the following is true:
+
+        - the asset doesn't exist for the given entity
+        - the status is already set with the same requested value
```
mgeplf, line 82: to confirm, we're relying on postgres transactions to prevent multi access?
do we have to worry about check if it's `AssetStatus.DELETED` before doing updates?

docs/assets.md
```diff
@@ -0,0 +1,124 @@
...
+- (not specific to assets) Entities must have a unique id that can be used in citations. Examples:
+    - `obi:reconstruction-morphology/27` (exposes the entity type and the internal numeric id)
+    - `obi:27` (the id is globally unique even without the entity type)
+
+## Asset management
+
+### Asset creation
+
+- Upload for small files can be done directly through the API (to be checked what limit to consider)
+- Upload for big files should be done through delegation as proposed originally in separate documents.
```
mgeplf, line 30: I think we need to say where the delegation doc is, or copy it here, or say that it will be copied :)

app/routers/asset.py
```diff
@@ -0,0 +1,192 @@
...
+    validate_filesize,
+)
+
+router = APIRouter(
+    prefix="",
+    tags=["assets"],
+)
+
+
+@router.get("/{entity_type}/{entity_id}/assets")
```
mgeplf, line 33: is there a reason why they are being namespaced byt `{entity_type}`?

mgeplf (commented):
First pass just to understand things, and ask a couple questions

app/db/model.py
GianlucaFicarelli: good idea

app/repository/asset.py
GianlucaFicarelli: yes, `scalar_one` raises a NoResultFound exception that can be intercepted at the service level with  ensure_result

app/routers/asset.py
GianlucaFicarelli: This matches the endpoints for the entities (and we need to ensure that we'll use the same naming convention including the hyphen separator).
For example, to get an entity we call `GET /reconstruction-morphology/1`
and to get the assets we call `GET /reconstruction-morphology/1/assets`
(at the moment it's reconstruction_morphology instead of reconstruction-morphology)

app/repository/asset.py
GianlucaFicarelli: This query would fail if we try to set the status to DELETED but the status is already DELETED, because of `scalar_one()`, and concurrency safety should be ensured by transactions.

If we want to prevent updating the status if it's DELETED (like transitioning from DELETED to CREATED), then some changes are needed, but we don't have this use case at the moment.

alembic/versions/20250210_171207_91dfdcef513d_.py
GianlucaFicarelli: We'll probably can consolidate all the migration files in one, before starting to deploy the service

app/db/types.py
```diff
@@ -28,3 +33,25 @@ def in_(column, values):
...
+JSONDICT = Annotated[dict[str, Any], mapped_column(JSONB)]
+
+
+class EntityType(StrEnum):
+    """Entity types that are directly exposed through the API.
+
+    For each entry:
+
+    - name (underscore separated): used for table names
+    - value (hyphen separated): used for endpoints ### FIXME
```
mgeplf, line 46: I don't think the fixme is needed; kebab case seems right on the call side

app/repository/entity.py
```diff
@@ -0,0 +1,75 @@
...
+            table = Base.metadata.tables[name]
+        except KeyError:
+            err = f"Table {name} not found"
+            raise RuntimeError(err) from None
+        if name not in cls._descendants:
+            err = f"Table {name} is not a subclass of entity"
+            raise TypeError(err)
+        return table
+
+    def get_entity(
```
mgeplf, line 35: do you think it would make sense to include the `project_id` check filter here? instead of having to call it at time of use?  maybe there an be an `get_entity_without_auth` for the cases where project_id isn't a filter, so we can grep for it to audit?

app/repository/entity.py
```diff
@@ -0,0 +1,75 @@
...
+
+from app.db.model import Base, Entity
+from app.db.types import EntityType
+from app.repository.base import BaseRepository
+
+
+class EntityRepository(BaseRepository):
+    """EntityRepository."""
+
+    _descendants = frozenset(
```
mgeplf, line 17: fancy, I'm surprised that sqlalchemy doesn't have something similar. I looked, but didn't find anything :(

app/repository/group.py
```diff
@@ -0,0 +1,38 @@
...
+    """Group of repositories sharing the same database session."""
+
+    def __init__(
+        self,
+        db: Session,
+        entity_repo_class: type[EntityRepository] = EntityRepository,
+        asset_repo_class: type[AssetRepository] = AssetRepository,
+    ) -> None:
+        """Init the repository group."""
+        self._db = db
```
mgeplf, line 21: aside: do you know how having copies of the `db` work? does it spin up a separate connection?

app/routers/asset.py
```diff
@@ -0,0 +1,192 @@
...
+def upload_entity_asset(
+    *,
+    repos: RepoGroupDep,
+    project_context: VerifiedProjectContextHeader,
+    s3_client: S3ClientDep,
+    entity_type: EntityType,
+    entity_id: int,
+    file: UploadFile,
+    meta: Annotated[dict | None, Form()] = None,
+    is_public: Annotated[bool, Form()] = False,
```
mgeplf, line 80: this isn't based on whether the entity is public?

app/routers/asset.py
```diff
@@ -0,0 +1,192 @@
...
+    project_context: VerifiedProjectContextHeader,
+    s3_client: S3ClientDep,
+    entity_type: EntityType,
+    entity_id: int,
+    asset_id: int,
+) -> AssetRead:
+    """Delete an assets associated with a specific entity.
+
+    The asset record is not deleted from the database, but its status is changed.
+    The file is actually deleted from S3, unless using a versioning-enabled bucket.
```
mgeplf, line 159: maybe we shouldn't delete it from s3? if the files are small, I'm not sure how expensive keeping them will be; I guess we can't just set them as private?

app/service/asset.py
```diff
@@ -0,0 +1,108 @@
...
+            http_status_code=HTTPStatus.NOT_FOUND,
+        )
+    # if (
+    #         not result.authorized_public
+    #         and result.authorized_project_id != project_context.project_id
+    # ):
+    #     raise ApiError(
+    #         message="Entity forbidden",
+    #         error_code=ApiErrorCode.ENTITY_FORBIDDEN,
+    #         http_status_code=HTTPStatus.FORBIDDEN,
```
mgeplf, line 30: I wonder if it should be `NOT_FOUND`; so that we don't leak existence of assets?

mgeplf (approved):

app/db/types.py
GianlucaFicarelli: The FIXME was indeed intended for deciding the case.
At the moment the entity endpoints are exposed with snake_case (reconstruction_morphology).
In the asset endpoints it's possible to use the enum value in kebab-case (reconstruction-morphology) but it's not used yet (so it still matches the entity endpoints).

app/repository/entity.py
GianlucaFicarelli: Yeah, same. There is `inspect`, but I didn't find a way to get the class from the table name or from the Table instance, unless iterating over the mappers.

app/service/asset.py
GianlucaFicarelli: I see the reason, but all the entity and assets ids are incremental integers, so they are not really hidden anyway.
However, the endpoint requires both the entity id and the asset id, and this can be slightly more difficult to guess.
It's a compromise between secrecy and clarity of the response.
Btw I also need to uncomment this code.

app/routers/asset.py
GianlucaFicarelli: If we allow mutability, and we delete the file, then it's possible to upload a new file with the same relative path only if we delete the old one.
This is because with the current file structure, the absolute path (s3 key) doesn't depend on the asset id, but only on the virtual_lab_id/project_id/entity_id.
This can be handy if we mount S3 and want to see all the files belonging to the same entity in the same prefix/directory.
If this is not required, we can store each file with a different prefix, as it's already done with Nexus.

app/routers/asset.py
GianlucaFicarelli: Right, since we require the entity_id, it can be inferred 👍

app/repository/group.py
GianlucaFicarelli: There is a pool of connections with the size depending on the config and instantiated at https://github.com/openbraininstitute/entitycore/blob/main/app/db/session.py#L60
Then db is a new Session created for each call to the endpoint, using the connections in the underlying pool.
If there aren't connections available, it should wait. We can do more testing to check the behaviour under heavy load, though.

app/repository/entity.py
GianlucaFicarelli: yes it sounds reasonable, I'll have a look.

app/repository/entity.py
GianlucaFicarelli: btw I'm still not completely sure if we want to restrict assets only to entities, or if inheriting from root is enough.
We can change it if needed.

app/repository/group.py
mgeplf: agreed, that will come later, I guess :(

app/repository/entity.py
mgeplf: Ok.

app/service/asset.py
mgeplf: Ok, works for me.

app/db/types.py
GianlucaFicarelli: agreed, we can make it uniform in a separate PR

app/routers/asset.py
GianlucaFicarelli: Done

app/routers/asset.py
GianlucaFicarelli: For the record, discussed, we can investigate and enable versioning enabled buckets on S3 to be able to retrieve or restore a deleted file.
See also https://docs.aws.amazon.com/AmazonS3/latest/userguide/RetrievingObjectVersions.html

mgeplf (approved):

PR #16, opened 2025-02-03
Improve import

mgeplf:
* make sure paths are sorted so the import is deterministic.

### DIFF
--- a/app/cli/import-data.py
+++ b/app/cli/import-data.py
@@ -256,7 +256,7 @@ def import_agents(data_list, db):
                     print(e)
 
 
-def import_single_neuron_simulation(data, db):
+def import_single_neuron_simulation(data, db, file_path):
     possible_data = [elem for elem in data if "SingleNeuronSimulation" in elem["@type"]]
     if not possible_data:
         return
@@ -298,7 +298,7 @@ def get_or_create_annotation(annotation_, reconstruction_morphology_id, db):
     return db_annotation.id
 
 
-def import_analysis_software_source_code(data, db):
+def import_analysis_software_source_code(data, db, file_path):
     possible_data = [data for data in data if data["@type"] == "AnalysisSoftwareSourceCode"]
     if not possible_data:
         return
@@ -327,7 +327,7 @@ def import_analysis_software_source_code(data, db):
             db.commit()
 
 
-def import_me_models(data, db):
+def import_me_models(data, db, file_path):
     def is_memodel(data):
         types = data["@type"]
         if isinstance(types, list):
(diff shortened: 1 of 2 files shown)


app/cli/import-data.py
```diff
@@ -104,6 +105,10 @@ def _import_annotation_body(data, db_type_, db):
             class_elem = curate.curate_etype(class_elem)
         db_elem = db.query(db_type_).filter(db_type_.pref_label == class_elem["label"]).first()
         if db_elem:
+            if (
+                db_elem.definition != class_elem.get("definition", "")
+            ) or db_elem.alt_label != class_elem.get("prefLabel", ""):
+                breakpoint()  # XXX BREAKPOINT
```
GianlucaFicarelli, line 111: Do we need this breakpoint, or is it a leftover?

app/cli/import-data.py
```diff
@@ -498,60 +505,89 @@ def import_morphologies(data_list, db):
...
+            .first()
+        )
+        if mfa:
+            already_registered += len(annotation)
             continue
+
+        if len(annotation) > 1:
+            duplicate_annotation += len(annotation) - 1
+
+        data = annotation[0]
```
GianlucaFicarelli, line 570: Cannot a morphology have multiple annotations (not duplicated), and wouldn't this ignore them?
For example https://openbluebrain.com/web/fusion/bbp/mmb-point-neuron-framework-model/resources/https%3A%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fmmb-point-neuron-framework-model%2F9358ec13-8ec2-4176-ab75-6f7a23f7ea4a#links
has 5 NeuronMorphologyFeatureAnnotation and they don't look all duplicates

app/cli/import-data.py
mgeplf: I'll remove it

app/cli/import-data.py
mgeplf: yeah, we'll need clarity on that from @jdcourcol, because he was skipping them for now, and I assume we probably want them.
This PR helps to show what is actually happening, imo, and should make it easier to chase down what we want to do.

GianlucaFicarelli: Sounds good, can we add a comment in the code?

GianlucaFicarelli (approved):

app/cli/import-data.py
jdcourcol: I swept it under the carpet. I needed time to investigate why this was happening.

mgeplf: I'll add a comment.
I was confused by the unique constraint on the annotation:
https://github.com/openbraininstitute/entitycore/blob/cda71cf6c6d03e6408cb3aed5fd78c019ff12491/app/db/model.py#L411

@jdcourcol the example @GianlucaFicarelli found is good, perhaps that will help w/ the investigation.

jdcourcol: with respect to the unique constraint, my expectation was to have only one NeuronMorphologyFeatureAnnotation  per morphology.

Issue #18, opened 2025-02-04
add missing properties to `/reconstruction-morphology`endpoint


bilalesi:
1. missing properties from the current `/reconstruction-morphologies`:

      - mtype
      - contributors
      - facets
      - distributions

2. currently `species` is an object, probably it should be a `list`.

mgeplf:
I've started on this:
* for the mtypes, I need to make the importer aware, and wire them up; I have started that
* I will have to do the same for the contributors
* the facets should be calculated when a search term is included, but I assume you want them all the time?
* distributions will come after we have file support

Finally, wrt species, I think I'm missing why they should be a list?

bilalesi:
> the facets should be calculated when a search term is included, but I assume you want them all the time?

 yep exactly, the facets should be in the response all time

bilalesi:
This is related to the comment in teams, 
> `mtype` should be included in search query params

bilalesi:
> contributions done here: [#28](https://github.com/openbraininstitute/entitycore/pull/28)

thanks @mgeplf 
i will check it soon and let you know

PR #23, opened 2025-02-05
Add MType annotation to morphologies

mgeplf:
* Need for front end, see #18


mgeplf:
The semantics of this have likely changed; the mtype annotations can be multiple...I'm putting this in WIP.

PR #27, opened 2025-02-05
Import thalamic morphologies

jdcourcol:
add morphologies with "NeuronMorphology" and no "ReconstructedNeuronMorphology" to be imported. Fix curation issues caused by these morphologies.


app/cli/curate.py
```diff
@@ -22,6 +22,8 @@ def curate_role(role):
 def curate_annotation_body(annotation_body):
     if "Mtype" in annotation_body["@type"]:
         annotation_body["@type"] = ["MType", "AnnotationBody"]
+    if annotation_body.get("@id", "") == "nsg:InhibitoryNeuron":
```
mgeplf, line 25: I'm wondering if this is fixed by this:
https://github.com/openbraininstitute/entitycore/pull/23/files#diff-158c27f0c6fc0ed8e45d7d3f6db2490b70231328e7381ee41f028a9ed1439bf7R149

jdcourcol: it is a bit different. The label that is used is "placeholder model" instead of "inhibitory neuron".

mgeplf (approved):
thanks

Issue #32, opened 2025-02-10
Import of data should set the correct creation_date/update_date not the one at the point of import


mgeplf:
As noticed by @bilalesi

PR #34, opened 2025-02-11
Add correct creation and update data based on data

mgeplf:
Added for:
* Agents
* Licenses
* AnnotationBody
* AnalysisSoftwareSourceCode
* MEModel
* EModel
* Mesh
* Traces
* Morphology
* MorphologyAnnotation
* Experimental Densities

This should fix https://github.com/openbraininstitute/entitycore/issues/32


Makefile
```diff
@@ -68,6 +68,11 @@ run-local: ## Run the application locally
 	uv run -m alembic upgrade head
 	uv run -m app run --host $(UVICORN_HOST) --port $(UVICORN_PORT) --reload
 
+import:  ## Run the import on a database, assumes mba_hierarchy.json and out are in the current dir
+	uv run -m alembic upgrade head
+	uv run -m app.cli.import-data hierarchy mba_hierarchy.json
+	uv run -m app.cli.import-data run ./out
+
```
GianlucaFicarelli, line 75: The import target is already defined above, so this shouldn't be added (it's probably the result of some merge)

app/cli/curate.py
```diff
@@ -77,15 +88,88 @@ def curate_etype(data):
...
+            "@type": "Person",
+            "givenName": "h",
+            "familyName": "arikris",
+            "_createdAt": datetime.datetime.now(datetime.UTC).isoformat(),
+            "_updatedAt": datetime.datetime.now(datetime.UTC).isoformat(),
+        },
+        {
+            "@id": "https://bbp.epfl.ch/nexus/v1/realms/bbp/users/ricardi",
+            "@type": "Person",
+            "givenName": "Nicolo",
```
GianlucaFicarelli, line 127:
```suggestion
            "givenName": "Niccolò",
```

app/cli/curate.py
```diff
@@ -77,15 +88,88 @@ def curate_etype(data):
...
+            "@id": "https://bbp.epfl.ch/neurosciencegraph/data/licenses/97521f71-605d-4f42-8f1b-c37e742a30b",
+            "label": "undefined",
+            "description": "undefined",
+            "_createdAt": datetime.datetime.now(datetime.UTC).isoformat(),
+            "_updatedAt": datetime.datetime.now(datetime.UTC).isoformat(),
+        },
+        {
+            "@id": "https://bbp.epfl.ch/neurosciencegraph/data/licenses/97521f71-605d-4f42-8f1b-c37e742a30bf",
+            "label": "undefined",
+            "description": "undefined",
```
GianlucaFicarelli, line 171: Maybe `SSCX Portal Data Licence final v1.0`? This is present in public/sscx/provEntity.json but the licenses are not imported from this file.

app/cli/curate.py
```diff
@@ -77,15 +88,88 @@ def curate_etype(data):
...
+            "_createdAt": datetime.datetime.now(datetime.UTC).isoformat(),
+            "_updatedAt": datetime.datetime.now(datetime.UTC).isoformat(),
+        },
+    ]
+
+
+def default_licenses():
+    return [
+        {
+            "@id": "https://bbp.epfl.ch/neurosciencegraph/data/licenses/97521f71-605d-4f42-8f1b-c37e742a30b",
```
GianlucaFicarelli, line 162: I see that this code is just moved from a different location, but the id is the same id as below, but without the trailing f. Also it doesn't seem used, so I think it's safe to remove it.

app/cli/import-data.py
```diff
@@ -457,26 +434,32 @@ def import_traces(data_list, db, file_path, project_id):
...
             license_id = utils.get_license_mixin(data, db)
             species_id, strain_id = utils.get_species_mixin(data, db)
+            createdAt, updatedAt = utils.get_created_and_updated(data)
+
             db_item = SingleCellExperimentalTrace(
                 legacy_id=[data.get("@id", None)],
-                name=name,
-                description=description,
+                name=data.get("name", None),
+                description=data.get("description", None),
```
GianlucaFicarelli, line 446: name and description cannot be None in SingleCellExperimentalTrace

app/cli/import-data.py
```diff
@@ -492,25 +475,29 @@ def import_morphologies(data_list, db, file_path, project_id):
...
             license_id = utils.get_license_mixin(data, db)
             species_id, strain_id = utils.get_species_mixin(data, db)
+            createdAt, updatedAt = utils.get_created_and_updated(data)
+
             db_reconstruction_morphology = ReconstructionMorphology(
                 legacy_id=[data.get("@id", None)],
-                name=name,
-                description=description,
+                name=data.get("name", None),
+                description=data.get("description", None),
```
GianlucaFicarelli, line 486: name and description cannot be None in ReconstructionMorphology

app/cli/import-data.py
```diff
@@ -674,20 +671,25 @@ def _do_import(db, input_dir, project_id):
     all_files = sorted(glob.glob(os.path.join(input_dir, "*", "*", "*.json")))
 
     print("importing agents")
+    import_agents(curate.default_agents(), db)
     for file_path in all_files:
         with open(file_path) as f:
             data = json.load(f)
-            import_agents(data, db)
+            possible_data = [
+                d for d in data if {"Person", "Organization"} & set(d.get("@type", {}))
```
GianlucaFicarelli, line 679: Minor, I find it more readable with:
```suggestion
                d for d in data if d.get("@type") in {"Person", "Organization"}
```

GianlucaFicarelli (commented):
LGTM but I added a few comments mainly related to code that was moved around

app/cli/curate.py
mgeplf: yup, very good eye.

app/cli/curate.py
mgeplf: nice find, I think I'll try and figure out why that isn't being detected as a license.

app/cli/import-data.py
mgeplf: `@type` can be a a list so I need to check for any overlap

GianlucaFicarelli (approved):

Issue #35, opened 2025-02-11
add a proper error propagation system


mgeplf:
Currently there are some code paths that raise exceptions for errors, but these should be wrapped and made clearer.

EX:
https://github.com/openbraininstitute/entitycore/pull/10#discussion_r1949083472

or the calls to `db.commit()` if a `sqlalchemy.exc.IntegrityError` happens.

PR #38, opened 2025-02-12
Add error handling and health and version endpoints

GianlucaFicarelli:
The error handling code has been split from https://github.com/openbraininstitute/entitycore/pull/14

### DIFF
--- a/app/application.py
+++ b/app/application.py
@@ -93,4 +93,8 @@ app.add_middleware(
 )
 app.include_router(
     router,
+    responses={
+        404: {"description": "Not found", "model": ErrorResponse},
+        422: {"description": "Validation Error", "model": ErrorResponse},
+    },
 )
--- a/app/application.py
+++ b/app/application.py
@@ -2,15 +2,22 @@ import asyncio
 import os
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
+from http import HTTPStatus
 from typing import Any
 
 from fastapi import FastAPI
+from fastapi.encoders import jsonable_encoder
+from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
+from starlette.requests import Request
+from starlette.responses import Response
 
 from app.config import settings
 from app.db.session import configure_database_session_manager
+from app.errors import ApiError, ApiErrorCode
 from app.logger import L
 from app.routers import router
+from app.schemas.api import ErrorResponse
 
 
 @asynccontextmanager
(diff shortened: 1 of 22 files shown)


app/application.py
```diff
@@ -33,10 +40,48 @@ async def lifespan(_: FastAPI) -> AsyncIterator[dict[str, Any]]:
         L.info("Stopping application")
 
 
+async def api_error_handler(request: Request, exception: ApiError) -> Response:
```
mgeplf, line 43: does it matter that these are async, but nearly everything else is currently sync?

app/application.py
```diff
@@ -33,10 +40,48 @@ async def lifespan(_: FastAPI) -> AsyncIterator[dict[str, Any]]:
...
+    """Handle API errors to be returned to the client."""
+    err_content = ErrorResponse(
+        message=exception.message,
+        error_code=exception.error_code,
+        details=exception.details,
+    )
+    L.warning("API error in {} {}: {}", request.method, request.url, err_content)
```
mgeplf, line 50: does this mean we shouldn't log errors on what is raising the error?

app/routers/morphology.py
```diff
@@ -33,10 +33,8 @@
     response_model=ReconstructionMorphologyExpand | ReconstructionMorphologyRead,
 )
 def read_reconstruction_morphology(db: SessionDep, rm_id: int, expand: str | None = None):
-    rm = db.query(ReconstructionMorphology).filter(ReconstructionMorphology.id == rm_id).first()
-
-    if rm is None:
-        raise HTTPException(status_code=404, detail="ReconstructionMorphology not found")
+    with ensure_result(error_message="ReconstructionMorphology not found"):
+        rm = db.query(ReconstructionMorphology).filter(ReconstructionMorphology.id == rm_id).one()
```
mgeplf, line 37: nice, I like it

mgeplf (approved):
LGreatTM

app/application.py
GianlucaFicarelli: Error handlers can be defined sync or async, but I experimented a bit to be sure:
1. error_handler async, endpoint async: the code in the error handler and the code in the endpoint are executed in the main thread
2. error_handler async, endpoint sync: the code in the error handler is executed in the main thread, the code in the endpoint is executed in a separate thread
4. error_handler sync, endpoint async: the code in the error handler is executed in a separate thread, the code in the endpoint is executed in the main thread
5. error_handler sync, endpoint sync: the code in the error handler and the code in the endpoint are executed in the same thread (that's not the main thread)

So I'd say that the usual recommendations apply also for the error handlers as for the endpoints: if it's blocking it should be defined as sync, if it's not blocking it can be defined as async (so it's not executed in a separate thread)

app/application.py
GianlucaFicarelli: ApiError should be used for errors to be notified to the client, that should be a sort of "expected" errors.
For this reason I don't think we have to log the full traceback, although the code that raises the ApiError can log more information if needed.
Do you think we should log more information here?

GianlucaFicarelli:
Fixes https://github.com/openbraininstitute/entitycore/issues/35

PR #39, opened 2025-02-12
Update facet output

mgeplf:
As asked for here:
https://github.com/openbraininstitute/entitycore/pull/28#issuecomment-2650115514


app/routers/morphology.py
```diff
@@ -86,23 +86,23 @@ def create_reconstruction_morphology(
...
-            .group_by(types)  # type: ignore[arg-type]
+            query.with_only_columns(table, func.count().label("count"))
+            .group_by(table)  # type: ignore[arg-type]
+            .order_by(table.name)  # type: ignore[attr-defined]
         )
-        facets[ty] = [Facet(id=r.id, label=r.name, count=count) for r, count in facet_q.all()]
+        facets[name] = [
+            Facet(id=row.id, label=row.name, count=count)
+            for row, count in db.execute(facet_q).all()
+            if row is not None  # exclude null rows
```
mgeplf, line 104: any idea why there are null rows?

GianlucaFicarelli: There can be null rows with strains, because strain_id in SpeciesMixin can be null.
We could decide to include the null rows in the response if we want that the sum of counts match the number of morphologies, but there isn't an id in that case.

mgeplf: ah, thanks, I don't think including them would be good

GianlucaFicarelli (approved):

PR #41, opened 2025-02-13
Add indexes and execute analyze

GianlucaFicarelli:
- Add indexes to all the foreign keys, since they can be used for joins
- Add index on creation_date, since it's used for ordering
- Automatically execute analyze after importing data, to update the statistics used by the query planner


app/db/model.py
```diff
@@ -285,10 +291,10 @@ class Entity(TimestampMixin, Root):
     # TODO: keep the _ ? put on agent ?
     createdBy = relationship("Agent", uselist=False, foreign_keys="Entity.createdBy_id")
     # TODO: move to mandatory
-    createdBy_id: Mapped[int] = mapped_column(ForeignKey("agent.id"), nullable=True)
+    createdBy_id: Mapped[int] = mapped_column(ForeignKey("agent.id"), index=True, nullable=True)
```
mgeplf, line 294: not sure the `agent` indices here and elsewhere will be needed; what's the use case? I'm not sure how much we'll be searching by who created/updated the Annotation/Entity.

app/db/model.py
```diff
@@ -377,7 +383,9 @@ class Mesh(DistributionMixin, Entity):
         nullable=False,
         autoincrement=True,
     )
-    brain_region_id: Mapped[int] = mapped_column(ForeignKey("brain_region.id"), nullable=False)
+    brain_region_id: Mapped[int] = mapped_column(
+        ForeignKey("brain_region.id"), index=True, nullable=False
```
mgeplf, line 387: I also wonder about this one.

mgeplf (commented):
not opposed to the indices I flagged, I'm just not sure they will be used; do you have `analyze` data that shows them helping?

app/db/model.py
GianlucaFicarelli: It could be used if we need to show the agent pref_label in the UI. is it possible?
Note also that we can remove unused indexes later, based on postgresql statistics.

mgeplf: I usually try and prefetch/preload things, but yeah, maybe.
Ok, forgiveness rather than permission :)

mgeplf (approved):

PR #42, opened 2025-02-13
add project_id for `Entity` during import

mgeplf:


Makefile
```diff
@@ -42,6 +42,12 @@ lint:  ## Run linters
 build:  ## Build the Docker image
 	docker compose --progress=plain build app
 
+import:  ## Run the import on a database, assumes mba_hierarchy.json and out are in the current dir
+	@$(call load_env,test-local)
```
GianlucaFicarelli, line 46: Nice addition, but shouldn't we import into the db specified with run-local instead of test-local?
We should also call:
```
	docker compose up --wait db
```

app/cli/import-data.py
```diff
@@ -738,13 +752,18 @@ def analyze():
 
 @cli.command()
 @click.argument("input-dir", type=REQUIRED_PATH_DIR)
-def run(input_dir):
+@click.option(
+    "--project-id",
+    type=str,
+    help="The UUID4 `project-id` under which the entities will be registered",
```
GianlucaFicarelli, line 758: Just wondering, what project should we use when importing data for real?

mgeplf: I do not have a good answer for that; hence why I made it a parameter.
I assume we'll make a purpose project of legacy data, or multiple projects.

Makefile
mgeplf: I guess I've just been calling this w/ my custom env file all the time...I can change it to whatever you want.

GianlucaFicarelli: yeah ok, I'm used to run migration against the local db exposed by run-local

GianlucaFicarelli (approved):

PR #44, opened 2025-02-14
add `APP_DISABLE_AUTH` env var that when set to true, disables the Bearer token check

mgeplf:
* it also allows using any project-id


mgeplf:
this feels hacky, if you have a better idea, let me know

GianlucaFicarelli (approved):

PR #52, opened 2025-02-24
add tests for pagination

mgeplf:


mgeplf:
This fixes https://github.com/openbraininstitute/entitycore/pull/52 for the `GET /{route}/` endpoints

app/routers/types.py
```diff
@@ -19,5 +30,5 @@ class Facet(BaseModel):
 
 class ListResponse[M: BaseModel](BaseModel):
     data: list[M]
```
GianlucaFicarelli, line 32: Are we good with `data`? This can be debatable and I cannot see a general consensus also in common 3rd parti APis so I'm fine with that, but if we want to change it it's better to do it now.

app/dependencies/__init__.py
```diff
@@ -0,0 +1,12 @@
+from typing import Annotated
+
+from fastapi import Depends
+
+from app.routers.types import PaginationRequest
+
+
+def page_validation(page: int | None = None, page_size: int | None = None) -> PaginationRequest:
+    return PaginationRequest(page=page or 0, page_size=page_size or 10)
```
GianlucaFicarelli, line 9: It's more common and intuitive for most users to have the first page=1 instead of 0.
We could add a default and validation >=1 to the PaginationRequest model.
We may decide to add also a max limit to page_size, but we can leave it for later when we have a better idea of reasonable values.

app/routers/types.py
```diff
@@ -1,13 +1,24 @@
...
+from pydantic import BaseModel, ConfigDict
 
 
-class Pagination(BaseModel):
+class PaginationRequest(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+
+    page: int
+    page_size: int
+
```
GianlucaFicarelli, line 9: We could add:
- validation
- default values
- optional but probably useful, a property returning `(page - 1) * page_size` (that's the `offset`)

app/routers/contribution.py
```diff
@@ -50,9 +70,10 @@ def read_contribution(id_: int, project_context: VerifiedProjectContextHeader, d
 def create_contribution(
     contribution: ContributionCreate, project_context: VerifiedProjectContextHeader, db: SessionDep
 ):
-    if not constrain_entity_query_to_project(
-        db.query(Entity).filter(Entity.id == contribution.entity_id), project_context.project_id
-    ).first():
+    stmt = constrain_entity_query_to_project(
+        sa.select(Entity).filter(Entity.id == contribution.entity_id), project_context.project_id
+    ).with_only_columns(sa.func.count())
+    if not db.execute(stmt).scalar_one_or_none():
```
GianlucaFicarelli, line 76: Minor, we could use scalar_one instead of scalar_one_or_none because `count(*)` should always return one row

app/routers/types.py
mgeplf: I swear I had this conversation with someone, but I can't find it at the moment.  We didn't find a consensus, maybe it was with @bilalesi ?

app/dependencies/__init__.py
mgeplf: Good point, I'll switch to 1 based.

app/routers/contribution.py
mgeplf: will do

app/routers/types.py
mgeplf: Shall I make a ticket and we do it in another PR?

GianlucaFicarelli: As you prefer, I commented here because this PR is adding `pagination_request.page * pagination_request.page_size` to most of the queries, and it can be replaced with `pagination_request.offset`

mgeplf: I was in the process of writing a function to transform the query, but that works too.
I'll do that.

mgeplf: good suggestion; makes it much nicer

GianlucaFicarelli (approved):

app/routers/types.py
GianlucaFicarelli: For comparison, we use these different formats:
In entitycore:
```json
{
  "data": [],
  "pagination": {
    "page": 0,
    "page_size": 0,
    "total_items": 0
  },
  "facets": {
    "additionalProp1": [
      {
        "id": 0,
        "label": "string",
        "count": 0,
        "type": "string"
      }
    ]
  }
}
```

In virtual-lab-api:
```json
{
  "message": "string",
  "data": {
    "results": [],
    "total": 0,
    "page": 0,
    "page_size": 0
  }
}
```
accounting-service:
```json
{
  "message": "string",
  "data": {
    "items": [],
    "meta": {
      "page": 1,
      "page_size": 1,
      "total_pages": 0,
      "total_items": 0
    },
    "links": {
      "self": "string",
      "prev": "string",
      "next": "string",
      "first": "string",
      "last": "string"
    }
  }
}
```

Issue #20, opened 2025-02-04
add `cell-composition-summary` file/endpoint


bilalesi:
this can be served as file (same as brain region) at the beginning 

1. fusion: https://openbluebrain.com/web/fusion/bbp/atlasdatasetrelease/resources/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fatlasdatasetrelease%2F6fd68be0-985a-4425-9460-91d3d932d72c
2. file url: https://openbluebrain.com/api/nexus/v1/files/bbp/atlasdatasetrelease/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fatlasdatasetrelease%2F6fd68be0-985a-4425-9460-91d3d932d72c
3. <img src="https://github.com/user-attachments/assets/0bd13d60-43da-44f2-b10b-cc3c48e794c2" data-canonical-src="https://github.com/user-attachments/assets/0bd13d60-43da-44f2-b10b-cc3c48e794c2" width="300" height="400" />

PR #53, opened 2025-02-24
add cell-composition-summary

mgeplf:
Fixes https://github.com/openbraininstitute/entitycore/issues/20

### DIFF
--- a/app/routers/cell_composition.py
+++ b/app/routers/cell_composition.py
@@ -1,0 +1,20 @@
+from pathlib import Path
+
+from fastapi import APIRouter, Response
+
+COMPOSITION_SUMMARY = (
+    (Path(__file__).parent.parent / "static/cellCompositionSummary_payload_prod.json").open().read()
+)
+
+
+router = APIRouter(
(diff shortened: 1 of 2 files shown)


bilalesi (approved):
LGTM

app/routers/cell_composition.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
+
+from fastapi import APIRouter, Response
+
+COMPOSITION_SUMMARY = (
+    (Path(__file__).parent.parent / "static/cellCompositionSummary_payload_prod.json").open().read()
```
GianlucaFicarelli, line 6: To avoid explicitly opening and closing the file, we can use:
```suggestion
    (Path(__file__).parent.parent / "static" / "cellCompositionSummary_payload_prod.json").read_text()
```
We could also json load/dump to be sure that the file contains valid json, but that's not so important since we control the file, and this is temporary.

mgeplf: True, but I hope this goes away before we have problems like that.

Issue #50, opened 2025-02-21
`BrainLocation` is currently a relationship, it probably can just be embedded in the model


mgeplf:
It seems unlikely that they will be reused, since it's a triple of floats.

eleftherioszisis:
Registering a payload it is currently counter-intuitive with regards to the brain location. Although it is an identifiable it is expected to be registered as `brain_location` instead of `brain_location_id`

For example the following morphology:
```
{'name': 'my-morph', 'description': 'A morphology', 'species_id': 1, 'strain_id': 1, 'brain_region_id': 68, 'brain_location_id': 1, 'legacy_id': 'temp'}
```

Fails to be registered with:
```
{"error_code":"INVALID_REQUEST","message":"Validation error","details":[{"type":"missing","loc":["body","brain_location"],"msg":"Field required"}]}
```

Whereas in the table itself:
```
            Column             |       Type        | Collation | Nullable | Default
-------------------------------+-------------------+-----------+----------+---------
 id                            | integer           |           | not null |
 description                   | character varying |           | not null |
 name                          | character varying |           | not null |
 morphology_description_vector | tsvector          |           |          |
 license_id                    | integer           |           |          |
 brain_location_id             | integer           |           |          |
 brain_region_id               | bigint            |           | not null |
 species_id                    | integer           |           | not null |
 strain_id                     | integer           |           |          |
```

Would it be possible to make all identifiables to be treated the same way? It breaks the general logic of serialization of identifiables from the client side where if a property X has an id it is expected to be registered as `X_id`.

It is also confusing as there is hidden logic in how the attributes are expected in a payload.

mgeplf:
All that would go away when I change `BrainLocation` to not be a relationship.
JDC is fine w/ that.

@eleftherioszisis; what would you name the attributes?

eleftherioszisis:
The ones you suggested in teams will do: `brain_location_[x|y|z]`

mgeplf:
Fixed in #56

PR #56, opened 2025-02-25
Remove `BrainLocation` table

mgeplf:
* Only the SEU `ReconstructionMorphology` the `coordinatesInBrainAtlas`
* the `JSONB` is used since it's not clear what the contents of location should be, beyond a triple of floats.  The reference atlas is probably required, too, for instance
* See https://github.com/openbraininstitute/entitycore/issues/50 for more details


alembic/versions/20250225_163455_48378d76d7b0_.py
```diff
@@ -0,0 +1,259 @@
...
+    op.create_table(
+        "brain_location",
+        sa.Column("id", sa.INTEGER(), autoincrement=True, nullable=False),
+        sa.Column("x", sa.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
+        sa.Column("y", sa.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
+        sa.Column("z", sa.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
+        sa.PrimaryKeyConstraint("id", name="pk_brain_location"),
+    )
+    op.create_index("ix_brain_location_id", "brain_location", ["id"], unique=False)
+    # ### end Alembic commands ###
```
GianlucaFicarelli, line 259: For confirmation: we are removing the column brain_location_id from:

- emodel
- experimental_bouton_density
- experimental_neuron_density
- experimental_synapses_per_connection
- memodel
- reconstruction_morphology
- single_cell_experimental_trace
- single_neuron_simulation
- single_neuron_synaptome

but we are replacing it with location only in reconstruction_morphology, because the coordinates are never used in the other entities?


> the JSONB is used since it's not clear what the contents of location should be, beyond a triple of floats. The reference atlas is probably required, too, for instance

I see that the AtlasRelease is present in the original data. Should it be imported as well?

app/db/model.py
```diff
@@ -421,6 +413,8 @@ class ReconstructionMorphology(LicensedMixin, LocationMixin, SpeciesMixin, Entit
     morphology_description_vector: Mapped[str] = mapped_column(TSVECTOR, nullable=True)
     morphology_feature_annotation = relationship("MorphologyFeatureAnnotation", uselist=False)
 
+    location: Mapped[PointLocation] = mapped_column(nullable=True)
```
GianlucaFicarelli, line 416: I wonder if `location` is clear enough to represent what was called "coordinatesInBrainAtlas" 🤔

alembic/versions/20250225_163455_48378d76d7b0_.py
mgeplf: > For confirmation: we are removing the column brain_location_id from:

Correct.  Having a table for a set of three floats as a relation seems unlikely to see reuse, IMO.

> but we are replacing it with location only in reconstruction_morphology, because the coordinates are never used in the other entities?

That was how I decided, and looking at the tables: I don't think there is a precise location for an `emodel`, a density (how could you have density at a point in space?), maybe `single_cell_experimental_trace` could be a point in space, but I suspect that would be more complicated than a single location, etc.

> I see that the AtlasRelease is present in the original data. Should it be imported as well?

Good question, we're currently not versioning the BrainRegions, which is probably where the `AtlasRelease` concept will show up, however that's still up in the air AFAIK.

app/db/model.py
mgeplf: I pondered that a bit, had `brain_location` at first, but didn't really like that - which other location could a morphology be in, but in a brain?

alembic/versions/20250225_163455_48378d76d7b0_.py
mgeplf: @jdcourcol should probably look at this, and give a :+1: / :-1:.

jdcourcol: I agree w/ removing the BrainLocation table and having the properties only on morph for now.

app/db/model.py
GianlucaFicarelli: Yeah ok, a similar atLocation was used to identify the file location, but it was a different context.
It should be enough in a given context.

GianlucaFicarelli (approved):

PR #57, opened 2025-02-25
Import distributions

GianlucaFicarelli:


app/repository/asset.py
```diff
@@ -50,17 +50,19 @@ def get_entity_asset(
 
     def create_entity_asset(self, entity_id: int, asset: AssetCreate) -> Asset:
         """Create an asset associated with the given entity."""
+        sha256_digest = bytes.fromhex(asset.sha256_digest) if asset.sha256_digest else None
```
mgeplf, line 53: are we trusting that the sha256 is created correctly by the client?

mgeplf (approved):

app/cli/import-data.py
```diff
@@ -823,5 +844,41 @@ def recurse(i):
...
+        query = sa.select(Asset.full_path, Asset.sha256_digest)
+        rows = db.execute(query).all()
+        for row in tqdm(rows):
+            dst = Path(row.full_path)
+            digest = row.sha256_digest.hex()
+            ignored.pop(digest, None)
+            if not dst.exists():
+                src = src_paths[digest]
+                Path(row.full_path).parent.mkdir(parents=True, exist_ok=True)
+                shutil.copy2(src, row.full_path)
```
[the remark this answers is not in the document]
GianlucaFicarelli: It should work, I suppose that the S3 sync work as well

app/repository/asset.py
GianlucaFicarelli: The sha256 is not sent by the client, but created on the server side here https://github.com/openbraininstitute/entitycore/pull/57/files#diff-1ca290196a1ef0caea07c7de99937f605797d2b7bbdcdcf750a18cc372afabdfR91

mgeplf: k, just wanted to be 100% sure. Thanks.

app/cli/import-data.py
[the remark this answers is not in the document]
mgeplf: not necessary at the moment, but it's a thought

mgeplf (approved):

PR #61, opened 2025-03-03
Rename asset endpoints to kebab-case

GianlucaFicarelli:


app/db/types.py
```diff
@@ -60,13 +60,13 @@ def process_result_value(self, value, dialect):  # noqa: ARG002, PLR6301
 JSONDICT = Annotated[dict[str, Any], mapped_column(JSONB)]
 
 
-class EntityType(StrEnum):
-    """Entity types that are directly exposed through the API.
+class EntityWithAssets(HyphenStrEnum):
+    """Entity with assets that are directly exposed through the API.
```
GianlucaFicarelli, line 64: For discussion: this enum is renamed to restrict the asset endpoints to accepts only some types of entities.
Alternatively, we could keep the name EntityType, and allow any entity (that still needs to be added to the enum)

mgeplf: I don't quite follow why one would restrict the asset by the entity? I think I just need an example use case.

GianlucaFicarelli: Yeah I don't have an actual use case, I was just considering that we may want to be able to apply some limitations.
I'll revert the renaming, and we'll see if any restriction will be needed in the future.

mgeplf: Ah, ok, I thought it was something I'm missing. Thanks.

PR #62, opened 2025-03-03
Add MType classification to morphologies

mgeplf:
* ReconstructionMorphology may have many mtypes in the future, so there
  is a `MTypeClassification` table make the connections to the `MTypeClass`
* These entries also hold who created the classification
* Currently using a "simple" mtype classification; a simple list of potential classes


app/routers/legacy/model/utils.py
```diff
@@ -64,10 +64,10 @@
         "pref_label": "label",
         "definition": "definition",
     },
-    MTypeAnnotationBody: {
-        "pref_label": "label",
-        "definition": "definition",
-    },
+    # MTypeAnnotationBody: {
```
mgeplf, line 67: I just commented these out, since I don't think it'll be used, but in case it is, we know it needs to be converted to the new scheme

app/cli/import-data.py
```diff
@@ -48,29 +49,33 @@ def get_or_create_annotation_body(annotation_body, db):
...
-        if db_annotation is MTypeAnnotationBody:
+        if annotation_type is MTypeClass:
             msg = f"Missing mtype in annotation body {annotation_body}"
             raise ValueError(msg)
-        ab = db_annotation(pref_label=annotation_body["label"])
+        ab = annotation_type(pref_label=annotation_body["label"])
         db.add(ab)
         db.commit()
-    return ab.id
+        print(f"Added Annotation: {ab}")
```
mgeplf, line 77: this was debug, I'll delete this in the review

tests/test_mtype.py
```diff
@@ -0,0 +1,100 @@
...
+    assert "mtypes" in data
+    mtypes = data["mtypes"]
+    assert len(mtypes) == 2
+
+    for mtype in mtypes:
+        assert "update_date" in mtype
+        del mtype["update_date"]
+
+        assert "creation_date" in mtype
+        del mtype["creation_date"]
```
GianlucaFicarelli, line 86: Instead of deleting the keys, it's possible to use ANY from unittest.mock (this would be just an existence test, but we could improve it to match patterns or types)

app/db/model.py
```diff
@@ -210,22 +210,33 @@ class AnnotationBody(LegacyMixin, TimestampMixin, Base):
...
+    __tablename__ = "mtype_classification"
+    id: Mapped[int] = mapped_column(
+        primary_key=True,
+        index=True,
+        nullable=False,
+        autoincrement=True,
+    )
+    createdBy_id: Mapped[int] = mapped_column(ForeignKey("agent.id"), index=True, nullable=True)
+    updatedBy_id: Mapped[int] = mapped_column(ForeignKey("agent.id"), index=True, nullable=True)
+    entity_id: Mapped[int] = mapped_column(ForeignKey("entity.id"), index=True, nullable=False)
```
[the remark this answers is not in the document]
mgeplf: good question, I don't know, so let's not do it for now.

app/filters/morphology.py
```diff
@@ -17,6 +23,7 @@ class MorphologyFilter(FilterWithAliases):
     brain_region_id: int | None = None
     species_id__in: list[int] | None = None
 
+    mtypes: MTypeClassFilter | None = FilterDepends(with_prefix("mtypes", MTypeClassFilter))
```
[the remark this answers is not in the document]
mgeplf: Ok, I'll change it; it's a bit of a misnomer, b/c they'll be a multiplicity at some point.

GianlucaFicarelli (approved):

PR #69, opened 2025-03-05
Return the contribution ids when retrieving a morphology

GianlucaFicarelli:
Other:
- rename the filter contributor* to contribution*

### DIFF
--- a/app/schemas/contribution.py
+++ b/app/schemas/contribution.py
@@ -28,6 +28,6 @@ class ContributionRead(ContributionBase, CreationMixin):
     entity: EntityRead
 
 
-class ContributorRead(ContributionBase):
+class ContributionReadWithoutEntity(ContributionBase, CreationMixin):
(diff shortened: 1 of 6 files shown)


app/schemas/contribution.py
```diff
@@ -28,6 +28,6 @@ class ContributionRead(ContributionBase, CreationMixin):
     entity: EntityRead
 
 
-class ContributorRead(ContributionBase):
+class ContributionReadWithoutEntity(ContributionBase, CreationMixin):
```
mgeplf, line 31: why on earth is the id on CreationMixin? strange

mgeplf (approved):

app/schemas/contribution.py
GianlucaFicarelli: It seemed strange to me as well

eleftherioszisis (approved):

PR #73, opened 2025-03-06
Add sorting for mtype to morphology endpoint

mgeplf:
* Requested feature
* Had to further subclass fastapi_filter since it didn't take the sub-filters into account


mgeplf:
this was at best half-brained, more like 1/4 brained.

PR #75, opened 2025-03-06
Make facets optional in the morphology endpoint

GianlucaFicarelli:
Some clients don't need the facets (for example entitysdk), so we can avoid N quite complex queries (one for each facet) in those cases.
We could decide what can be the default value for the parameter `with_facets`.
By setting `with_facets=False` by default, we should be more confident that the facets are requested only when needed.
However, this PR has `with_facets=True` for backward compatibility.

This PR doesn't need to be merged yet, because I don't see a big difference in response times when the search is executed locally, but it can be used for simple benchmarks:
- ~360 ms for http://127.0.0.1:8000/reconstruction-morphology?page_size=1000&with_facets=true
- ~310 ms for http://127.0.0.1:8000/reconstruction-morphology?page_size=1000&with_facets=false

### DIFF
--- a/app/routers/morphology.py
+++ b/app/routers/morphology.py
@@ -144,7 +144,7 @@ def morphology_query(
     pagination_request: PaginationQuery,
     morphology_filter: Annotated[MorphologyFilter, FilterDepends(MorphologyFilter)],
     search: str | None = None,
-    with_facets: bool = True,
+    with_facets: bool = False,
 ) -> ListResponse[ReconstructionMorphologyRead]:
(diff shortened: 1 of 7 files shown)


mgeplf:
`By setting with_facets=False by default` would be my vote; I don't think we need to maintain backwards compat yet.

GianlucaFicarelli:
I've set `with_facets=False` by default and fixed tests.
At the moment the morphology endpoint is the only one returning facets, but we should add that parameter to other endpoints when/if we add facets to them.
I'm going to merge unless there is any remaining concern.

PR #79, opened 2025-03-07
Return allowed query params

GianlucaFicarelli:
Note that the params that are excluded from the json schema are intentionally not mentioned in the error response, assuming that they are internal or hidden for a reason.

### DIFF
--- a/app/dependencies/common.py
+++ b/app/dependencies/common.py
@@ -20,9 +20,15 @@ def forbid_extra_query_params(
     but this option is intended for internal use, and it might be removed in the future.
     """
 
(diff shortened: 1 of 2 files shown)


GianlucaFicarelli:
> I'm so used to not having to handle this, since unused query params are usually ignored :)

The issue is that we can have a lot of parameters for filtering. If the client sends a wrong parameter, or assumes that it exists when it doesn't, and we don't check that, then he will be surprised than filtering doesn't work, for example.

mgeplf:
Yeah, I get the use case, I just figure the caller normally handles that; it's fine.

PR #83, opened 2025-03-10
Order morphologies by mtype

g-bar:
Since mike is off a took a stab at this one :).

Usage:

`curl http://localhost:8000/reconstruction-morphology?page=1&page_size=100&order_by=mtype&mtype__order_by=pref_label`


g-bar:
Nevermind, seemed to be working but it messed up with the pagination. Then I made some changes and now it's not working at all.  I'm closing for now. This feature is tricky.

GianlucaFicarelli:
> Nevermind, seemed to be working but it messed up with the pagination. Then I made some changes and now it's not working at all. I'm closing for now. This feature is tricky.

Thanks for looking into it, we can revisit it later if we need.
Just a note about the params, at first sight I'd prefer to be able to pass something like `order_by=mtype.pref_label`  instead of `order_by=mtype&mtype__order_by=pref_label`, but this can be discussed later as well.

PR #90, opened 2025-03-12
Replace StringList with ARRAY(VARCHAR)

GianlucaFicarelli:
Note: it requires `make destroy`, optionally followed by `make import`.


eleftherioszisis (approved):

mgeplf:
Nice, I think the StringList thing was a holdover from when JDC was doing things w/ sqlite?

PR #91, opened 2025-03-14
Add SingleNeuronSimulation

eleftherioszisis:
* Add router for single-neuron-simulation for creating/reading single cell simulations
* Add read/create schemas
* Add simple schemas for MEModel for the needs of SIngleNeuronSimulation

Questions/Observations:

**1. SingleNeuronSimulation should be an Activity**
SingleNeuronSimulation serves the purpose of an activity which should have a layout of:
- used (config used for this activity)
- generated (result artifact of this activity)

However, it directly links an me_model, and has config + results in the attached file. While this is fine given that the single cell simulation is not a workflow task, if in the future we want to use it in the workflow context, it will not follow the expected Activity layout.

**2. injectionRecording is a list, but a singular entry in the example resource.**
In the existing model:
```
    injectionLocation: Mapped[StringList] = mapped_column(default="")
    recordingLocation: Mapped[StringList] = mapped_column(default=[])
```
In the example resource:
```
  "injectionLocation": "soma[0]",
  "recordingLocation": ["soma[0]_0.5"],
```
Are there use cases for multiple injection locations?


jdcourcol:
yes, injection location could be a list indeed.

jdcourcol:
> 1. SingleNeuronSimulation should be an Activity
yeah, we should rework that you are right.

app/schemas/simulation.py
```diff
@@ -0,0 +1,36 @@
...
+class SingleNeuronSimulationCreate(
+    SingleNeuronSimulationBase,
+    AuthorizationOptionalPublicMixin,
+):
+    me_model_id: int
+    brain_region_id: int
+
+
+class SingleNeuronSimulationRead(
+    SingleNeuronSimulationCreate,
```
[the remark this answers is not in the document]
eleftherioszisis: Indeed we don't want the ids. Fixed it.

app/routers/single_neuron_simulation.py
```diff
@@ -0,0 +1,58 @@
...
+    SingleNeuronSimulationRead,
+)
+
+router = APIRouter(
+    prefix="/single-neuron-simulation",
+    tags=["single-neuron-simulation"],
+)
+
+
+@router.get(
```
jdcourcol, line 21: There is no route without id ?

eleftherioszisis: > There is no route without id ?

Not yet. Should it return a paginated result of all entities? Do we also need filtering support?

jdcourcol: yes we need pagination, filtering and fulltext search. I think you can take inspiration for @g-bar 's PR

app/filters/single_neuron_simulation.py
```diff
@@ -0,0 +1,27 @@
...
+
+class SingleNeuronSimulationFilter(
+    CustomFilter,
+    CreationFilterMixin,
+    ContributionFilterMixin,
+):
+    name__ilike: str | None = None
+    brain_region_id: int | None = None
+
+    me_model: MEModelFilter | None = FilterDepends(with_prefix("me_type", MEModelFilter))
```
bilalesi, line 21: `status` is good option too, in bluenaas we have this status: 
"started", "failure", "success",
but it's open to change in the next version of bluenaas

bilalesi:
the `single_neuron_simulation` model is missing the `results` and `config` properties, are we planning to save those as `json` field or as `Assets`, 
for the config is too small we can make it even as columns in the db
for the results sometimes it's get bigger depends on the `number of recording` and `number of amplitudes`, `duration`
(for 1 recording and 5 amplitudes "the default config" is around 10mb)

eleftherioszisis:
> the `single_neuron_simulation` model is missing the `results` and `config` properties, are we planning to save those as `json` field or as `Assets` ?, for the config is too small we can make it even as columns in the db for the results sometimes it's get bigger depends on the `number of recording` and `number of amplitudes`, `duration` (for 1 recording and 5 amplitudes "the default config" is around 10mb)

I wonder why the config does not also specify the me-model that is going to be used. Do we need two different components that act as configuration inputs?

An [Activity schema](https://www.w3.org/TR/prov-o/#Activity) would usually be defined with a "used" property with the used configuration that parameterizes the simulation activity, and a "generated" that is the activity's produced entity, i.e. the results.

app/filters/single_neuron_simulation.py
eleftherioszisis: Would it be ok if we use the same status enum as workflow activities?
https://github.com/openbraininstitute/entity-management/blob/980fb8e341e828bab860a802fcf30201b34c3c69/entity_management/core.py#L321

bilalesi: yes, I am okey with those two but then we need to change `bluenaas`

eleftherioszisis: ok, I will leave them as in blueness for now.

bilalesi:
> I wonder why the config does not also specify the me-model that is going to be used. Do we need two different components that act as configuration inputs?

the me-model was saved in the resource before
```
"used": {
    "@id": "https://bbp.epfl.ch/data/bbp/mmb-point-neuron-framework-model/1b89d6a2-3d6e-4530-a4af-6752412987ea",
    "@type": "MEModel"
  }
```

the reason why we did not save it in the config back then was due we did not have the full picture on how the final configuration will look like, it was open to change for the fields name, what we should kept and not, so DKE asks to save it as json and then will add those in the resource schema when we have the final version

app/routers/single_neuron_simulation.py
```diff
@@ -0,0 +1,188 @@
...
+    project_context: VerifiedProjectContextHeader,
+):
+    with ensure_result(error_message="SingleNeuronSimulation not found"):
+        query = (
+            constrain_to_accessible_entities(
+                sa.select(SingleNeuronSimulation),
+                project_context.project_id,
+            )
+            .filter(SingleNeuronSimulation.id == id_)
+            .options(joinedload(SingleNeuronSimulation.me_model))
```
g-bar, line 51: I think we're missing the joinedloed with BrainRegion since it's included in the SingleNeuronSimulationRead schema.

eleftherioszisis:
> the `single_neuron_simulation` model is missing the `results` and `config` properties, are we planning to save those as `json` field or as `Assets` ?, for the config is too small we can make it even as columns in the db for the results sometimes it's get bigger depends on the `number of recording` and `number of amplitudes`, `duration` (for 1 recording and 5 amplitudes "the default config" is around 10mb)

Would it work to store it as an asset file similar to how it was stored before?

app/schemas/simulation.py
```diff
@@ -0,0 +1,40 @@
...
+)
+from app.schemas.me_model import MEModelRead
+
+
+class SingleNeuronSimulationBase(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    name: str
+    description: str
+    seed: int
+    status: str
```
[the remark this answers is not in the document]
eleftherioszisis: Yeah, I added a SingleNeuronSimulationStatus

g-bar:
I think we're missing a trigger to populate the description_vector when name, description are updated. But we can merge and I'll add it in my PR since I already have a function do it. https://github.com/openbraininstitute/entitycore/pull/101/files#diff-a7e62c6cce5db0921b3561ed08af4c0dbce2830301041b753a0db854bc164a2b

Also maybe we can add more tests for the list view later on.

g-bar (approved):

PR #94, opened 2025-03-17
Draft: Emodel routes

g-bar:


app/cli/import_data.py
```diff
@@ -389,6 +412,48 @@ def ingest(db, project_context, data_list):
...
+                None,
+            )
+
+            morphology = exemplar_morphology_id and utils._find_by_legacy_id(
+                exemplar_morphology_id, ReconstructionMorphology, db
+            )
+
+            assert morphology
+
+            # TODO: Import as Asset?
```
jdcourcol, line 447: This is storing things like that in a list:
grafik

@GianlucaFicarelli @eleftherioszisis  would you have any input ?

jdcourcol: There is the "implicit knowledge than the next image in the list is the stimulus:
grafik

eleftherioszisis:
We also need a the emodel script and mechanisms that are hidden within the followed workflow, right?

app/cli/import_data.py
jdcourcol: there are  217 items in that particular list.

jdcourcol: the model is 
For each StimulusType (StartHold), there a certain number of repetition id (1,2,3,...), for each repetition there are 2 images, one for the stimulation trace and one for the response trace.

jdcourcol: I don't know if we can easily re-generate these images.

GianlucaFicarelli: For Emodel,

1. a possibility can be to store these images as assets as well, although they are not distributions in Nexus.
We could add a column "attribute" (or any better name) into the asset table, that could be used to identify the relationship that should be populated in the entity model.

emodel:
- id (PK)
- stimulus (1-to-1 relationship)
- images (1-to-many relationship for the other images)

asset:
- id (PK)
- entity_id (FK to entity)
- attribute (e.g. "stimulus", "image", it could be a free string or an enum, or a reference to another table but probably it cannot be fully dynamic)

Need to check if this plays nicely with automatically loading the relationship in sqlalchemy.
If there are metadata, they can be stored in the asset metadata (the schema is not enforced in postgresql but it can be in pydantic)

2. Another approach could be to add another table with one record for each image, and the records in the asset table should reference those records.

For SingleCellExperimentalTrace, it could be something similar.

--

I'm not sure that we have these images in the exported data, and I cannot see the digest in the emodel entity.

For example, the emodel https://bbp.epfl.ch/data/bbp/mmb-point-neuron-framework-model/67fc3989-335c-4ae4-aad3-f74884d82c9f
contains the image https://bbp.epfl.ch/data/bbp/mmb-point-neuron-framework-model/b66e600c-fdfe-4d2d-9429-044d840b5dcd
but I cannot find the digest of the file anywhere in the dumped metadata.

Looking at Nexus, the resources are:
- emodel https://openbluebrain.com/api/nexus/v1/resources/bbp/mmb-point-neuron-framework-model/_/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fmmb-point-neuron-framework-model%2F67fc3989-335c-4ae4-aad3-f74884d82c9f
- image: https://openbluebrain.com/api/nexus/v1/resources/bbp/mmb-point-neuron-framework-model/_/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fmmb-point-neuron-framework-model%2Fb66e600c-fdfe-4d2d-9429-044d840b5dcd
and the image contains the digest.

jdcourcol: > I'm not sure that we have these images in the exported data, and I cannot see the digest in the emodel entity.

I think I was looking for ContentUrl, and this thing is not following this pattern unfortunately.

jdcourcol: @GianlucaFicarelli @eleftherioszisis  . I feel like we could model this in a different way:
- let's say that this "asset" is a directory in S3 (like a circuit)
- this directory contains a json file, with the list of stimuli and the relative path of the image file. (like a circuit config somehow)
- we have an API to fetch particular file (knowing the S3 directory + local path)

That would be more generic as it would fit the bill for this use case and circuits for instance (say I want to get only the node file from a circuit)

It would make the database schema less ugly than the one used in nexus.

jdcourcol: That is not how I see it.
I see the flow like that:
- Client find an asset of type "trace images" with a url U representing S3 directory where a json file (dict.json)+ images are stored
- Client ask for the json file by providing  U  + dict.json name
- Client open the dict.json file it just fetched. In that json file he can see a dictionary like:
{
stimulus:{
type: foo,
repetitions:[
{stim: img1.png,
resp: img2.png
....
- client ask for (U, img1.png) and get the img file

So instead of having one call to retrieve files from entity core, we have a call to retrieve files when the asset is a directory.

entity core does not know anything about the list of stimuli and local path

jdcourcol:
> We also need a the emodel script and mechanisms that are hidden within the followed workflow, right?

yes, as mentioned in the chat, it would be better to have direct properties from the e-model to the actual script and mechanisms.

app/cli/import_data.py
g-bar: The thumbnail is used in the list view. So we would have to fetch all the json files.

jdcourcol: @g-bar That is not the same image. In the list view, the thumbnail is generated by thumbnail service.
In the detail page, we use the above.
In the detail page + interactive view, we use some binary file for that.

jdcourcol: sorry ! I switch to Ephys traces ! That is where we use the thumbnail service.

jdcourcol: My comment above with the dict.json applied to e-phys traces. It is the same pattern here too.

However, here it seems it is a bunch of metrics generated for the e-model (currentscape, scores...).

With respect to the thumbnail, I can imagine than instead of fetching the thumbnail from the distribution each time, we cache it with the thumbnail manager.

app/filters/emodel.py
```diff
@@ -0,0 +1,46 @@
...
+class EModelFilter(CustomFilter):
+    id: int | None = None
+    creation_date__lte: datetime | None = None
+    creation_date__gte: datetime | None = None
+    update_date__lte: datetime | None = None
+    update_date__gte: datetime | None = None
+    name__ilike: str | None = None
+    brain_location_id: int | None = None
+    brain_region_id: int | None = None
+    species_id__in: list[int] | None = None
```
GianlucaFicarelli, line 26: This should be fixed here and also in MorphologyFilter (although not part of this PR):
- species_id__in should be uuid now
- brain_location_id doesn't exist anymore, so we can remove the filter

app/routers/emodel.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        db.add(db_em)
+        db.commit()
+        db.refresh(db_em)
+
+        query = emodel_joinedloads(sa.select(EModel).filter(EModel.id == db_em.id))
+        return db.execute(query).unique().scalar_one()
+
+    except InternalError as err:
```
GianlucaFicarelli, line 99: I wonder if there is any way to be more specific about the exception to catch?

app/schemas/emodel.py
```diff
@@ -0,0 +1,51 @@
...
+    score: float
+    seed: int
+
+
+class EModelCreate(EModelBase, AuthorizationOptionalPublicMixin):
+    species_id: uuid.UUID
+    strain_id: uuid.UUID | None
+    brain_region_id: int
+    exemplar_morphology_id: uuid.UUID
+    legacy_id: str | None
```
GianlucaFicarelli, line 36: Not specific to this PR, but I'm not sure that we need to expose legacy_id in the creation endpoint, since the legacy data are populated via the import script.
Side note, we store the legacy_id as a list, because there can be more than one legacy id in some cases.

app/schemas/types.py
```diff
@@ -15,6 +17,20 @@ class PaginationRequest(BaseModel):
...
+    def paginated_rows(self, db: Session, q: Query | sa.Select):
+        q = self.paginate_query(q)
+        return db.execute(q)
+
+    def pagination(self, db: Session, q: sa.Select, distinct_col: InstrumentedAttribute):
+        total_items = db.execute(
+            q.with_only_columns(sa.func.count(sa.func.distinct(distinct_col)).label("count"))
+        ).scalar_one()
+
+        return PaginationResponse(page=self.page, page_size=self.page_size, total_items=total_items)
```
GianlucaFicarelli, line 32: I'm not convinced that's a good idea to put the query logic in methods inside the schema 🤔

app/routers/common.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __call__(self, q: sa.Select, vector_col: InstrumentedAttribute):
+        if not self.search:
+            return q
+
+        return q.where(vector_col.match(self.search))
+
+
+FacetsDep = Annotated[_FacetsDep, Depends()]
+SearchDep = Annotated[Search, Depends()]
```
GianlucaFicarelli, line 78: I like that this code has been refactored out and made it common.
However, we should move it to a different location, because it doesn't belong to the routing logic, but this can be done later in another PR.

app/schemas/types.py
```diff
@@ -27,7 +43,7 @@ class PaginationResponse(BaseModel):
 class Facet(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 
-    id: uuid.UUID
+    id: uuid.UUID | int
```
GianlucaFicarelli, line 46: The `int` type is needed for the brain_region, right? We were discussing with @eleftherioszisis that it would be cleaner if also the brain_region is using ids as the other entities. No changes needed for this PR, though.

app/routers/emodel.py
g-bar: The err.orig is a long string like so:

```py


"authorized_project_id mismatch or entity is not public
CONTEXT:  PL/pgSQL function unauthorized_private_reference_function_emodel_exemplar_morphol() line 9 at RAISE"
```

So the only way is to parse the string and find the "authorized_project_id mismatch or entity is not public" from it. Maybe we can come up with some InternalError codes.

app/schemas/emodel.py
g-bar: While we migrate some data will have to be fetched from nexus (for example the EModelConfiguration), so we need to make it available at least while we migrate.

But I'm not sure if it should be 'legacy_id' or 'legacy_self'?  What is the difference between those?

app/schemas/types.py
g-bar: What I liked about this pattern is that it binds the query params to the functionality that needs it. Since the page_size and page query params are only needed by this query, no need to pass them around, they only live inside this class.

Pydantic schemas have to be used with Depends to be able to used as a query params, so we can think of this more of a dependency that adds those query params rather than a schema. 

Maybe we could rename it to PaginationDep and move it somewhere else?

app/schemas/emodel.py
g-bar: Ok, removed it from the schemas.

app/routers/common.py
g-bar: Cool, I'll add it in a new PR.

app/schemas/types.py
g-bar: Indeed it's for the brain_region.

app/filters/emodel.py
g-bar: Changed.

app/db/triggers.py
```diff
@@ -1,16 +1,104 @@
...
+        "emodel_description_vector",
+        "description_vector",
+        ["description", "name"],
+    ),
+    description_vector_trigger(
+        SingleNeuronSimulation,
+        "single_neuron_simulation_description_vector",
+        "description_vector",
+        ["description", "name"],
+    ),
```
g-bar, line 97: I added a trigger here @eleftherioszisis

app/routers/emodel.py
GianlucaFicarelli: Interesting, thanks for looking into that 👍

app/db/triggers.py
eleftherioszisis: Perfect, thanks!

app/filters/emodel.py
```diff
@@ -0,0 +1,45 @@
...
+    AgentFilter,
+    ETypeClassFilter,
+    MTypeClassFilter,
+    SpeciesFilter,
+)
+from app.filters.morphology import MorphologyFilter
+
+
+class EModelFilter(CustomFilter):
+    creation_date__lte: datetime | None = None
```
GianlucaFicarelli, line 19: I see that `id` was removed from the filter in the last commits.
We could consider adding `id__in` to be able to retrieve a list of entities by id in one request.
However, we would be probably limited by the length of the query string at some point, so I'm not sure it's worth to add it now.

g-bar: Indeed for id we have the /emodels/id route.

Is there a use case if id__in? At least on the frontend there isn't. Can't hurt to add it though.

.gitignore
```diff
@@ -8,3 +8,8 @@ coverage.xml
 .idea
 *.egg-info
 tmp
+out.tar
+.DS_Store
+mba_hierarchy.json
+notes.md
+out.tar
```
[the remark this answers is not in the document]
g-bar: I'll remove the double out.tar. 

For notes.md I like to keep a notes.md file in every repo to have some notes only applicable to that repo instead of having them all jumbled in the same place. I don't think it clashes with anything?

app/schemas/types.py
GianlucaFicarelli: If we are going to do some refactoring I'd like to try to separate the logic in router (view), service (business logic), repository (db access). Also @eleftherioszisis is willing to do some refactoring to reduce code duplication, so it can be useful if we coordinate together directly.

g-bar: Cool, reverted this change for now.

app/filters/emodel.py
GianlucaFicarelli: I think we can wait, we'll add and test it if needed.

.gitignore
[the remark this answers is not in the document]
GianlucaFicarelli: We don't have a standard naming for local uncommitted notes. It looks a bit weird to me to list the file here, but it's a minor thing and maybe it could be used by others as well, we can keep it

g-bar:
> Looks good and clean to me 👍

Thanks!

PR #98, opened 2025-03-20
add legacy_id and assets to morphology endpoint response

bilalesi:
1. include `legacy_id` in the morphology (get-one, get-all) endpoints
2. include `assets` in the morphology (get-one, get-all) endpoints


app/routers/morphology.py
```diff
@@ -72,6 +74,8 @@ def read_reconstruction_morphology(
             .options(joinedload(ReconstructionMorphology.license))
             .options(joinedload(ReconstructionMorphology.species))
             .options(joinedload(ReconstructionMorphology.strain))
+            .outerjoin(Asset, Asset.entity_id == ReconstructionMorphology.id)
+            .options(contains_eager(ReconstructionMorphology.assets))
```
g-bar, line 78: Minor but I think joinedload on Morphology.assets defaults to left outer join, so we could just do that.

> Joined eager loading is the most fundamental style of eager loading in the ORM. It works by connecting a JOIN (by default a LEFT OUTER join) to the SELECT statement 


`Joined eager loading is the most fundamental style of eager loading in the ORM. It works by connecting a JOIN (by default a LEFT OUTER join) to the SELECT statement `

GianlucaFicarelli: After some local testing it seems that for one-to-many relationships, selectinload is more performant than joinedload.
```
page_size=1000 morphologies, overall time measured on the client:
joinedload: 1.2-1.4 seconds
selectinload: 0.6 seconds
selectinload also in contributions/agent/role: 0.45 seconds
```
I pushed the changes also for list endpoint and also for the single morphology endpoint, although in this latter case the difference is minimal.

g-bar: Ah nice!

app/routers/morphology.py
```diff
@@ -165,11 +180,16 @@ def morphology_query(
...
         .options(joinedload(ReconstructionMorphology.species, innerjoin=True))
         .options(joinedload(ReconstructionMorphology.strain))
-        .options(joinedload(ReconstructionMorphology.contributions).joinedload(Contribution.agent))
-        .options(joinedload(ReconstructionMorphology.contributions).joinedload(Contribution.role))
+        .options(
+            selectinload(ReconstructionMorphology.contributions).selectinload(Contribution.agent)
+        )
+        .options(
+            selectinload(ReconstructionMorphology.contributions).selectinload(Contribution.role)
+        )
```
g-bar, line 188: @GianlucaFicarelli I think the selectinload is ok for the single result view, but for the list view it will result in 
N+1 queries. Since it will do a query to get the ids for each morphology, and then "select where id in list of ids" for each morphology.

I'm not sure would need to check.

GianlucaFicarelli: "selectinload" will execute only a single additional query like "select ... where id in (...)".
The ids are chunked by 500 items (it's hardcoded in sqlalachemy [here](https://github.com/sqlalchemy/sqlalchemy/blob/dabd77992d785cad89ed110acd2f648a454fb7ae/lib/sqlalchemy/orm/strategies.py#L2977)), to avoid lists of ids with too many items

GianlucaFicarelli (approved):

PR #101, opened 2025-03-20
MEModel

g-bar:
Implements MEModel routes inc facets, filters, search and pagination.


app/db/model.py
```diff
@@ -340,36 +408,54 @@ class Mesh(LocationMixin, Entity):
...
+        Enum(ValidationStatus, name="me_model_validation_status"),
+        nullable=False,
+        default=ValidationStatus.created,
+    )
+
+    mmodel_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("reconstruction_morphology.id"), nullable=False
+    )
+
+    mmodel = relationship("ReconstructionMorphology", foreign_keys=[mmodel_id], uselist=False)
```
eleftherioszisis, line 437: In the past, we used to refer to mmodel as the inputs to synthesis. However, we may want to change the meaning of an mmodel and use it like here.

g-bar: Yeah I wasn't sure how to name it, also EModel will change to ElectricalCellModel, so I'm not sure. Maybe just
morphology_id and electrical_cell_model_id? Any thoughts @jdcourcol ?

eleftherioszisis: I would suggest going with morphology_id for now.

app/schemas/me_model.py
```diff
@@ -1,23 +1,47 @@
...
 ):
+    id: uuid.UUID
+    species: SpeciesRead
+    strain: StrainRead | None
     brain_region: BrainRegionRead
+    contributions: list[ContributionReadWithoutEntity] | None
+    mtypes: list[MTypeClassRead] | None
+    etypes: list[ETypeClassRead] | None
+    mmodel: MModel
+    emodel: EModelBase
```
GianlucaFicarelli, line 47: MModel and EModelBase don't contain the id, but I think we should return it

tests/test_memodel.py
```diff
@@ -0,0 +1,553 @@
...
+
+from fastapi.testclient import TestClient
+from sqlalchemy.orm import Session
+
+from app.db.model import EModel, ReconstructionMorphology
+
+from .conftest import CreateIds, MEModels
+from .utils import (
+    PROJECT_HEADERS,
+    UNRELATED_PROJECT_ID,
```
GianlucaFicarelli, line 12: Now there are different fixtures for clients with different permissions, so we don't need to manually specify the headers in the tests

app/routers/memodel.py
```diff
@@ -0,0 +1,178 @@
...
+            sa.select(MEModel), user_context.project_id
+        ).filter(MEModel.id == id_)
+
+        query = memodel_joinedloads(query)
+
+        return db.execute(query).unique().scalar_one()
+
+
+@router.post("", response_model=MEModelRead)
+def create_emodel(
```
GianlucaFicarelli, line 72: Minor
```suggestion
def create_memodel(
```

app/routers/memodel.py
```diff
@@ -0,0 +1,178 @@
...
+            authorized_public=memodel.authorized_public,
+            validation_status=memodel.validation_status,
+        )
+
+        db.add(db_em)
+        db.commit()
+        db.refresh(db_em)
+
+        query = memodel_joinedloads(sa.select(MEModel).filter(MEModel.id == db_em.id))
+        return db.execute(query).unique().scalar_one()
```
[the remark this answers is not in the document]
g-bar: This is not allowed, the trigger raises exception if the reference is not public.
https://github.com/openbraininstitute/entitycore/blob/79fbc0cb65fc0065af28b48dfd25aabcb2e62d71/app/db/triggers.py#L50

I can add a test.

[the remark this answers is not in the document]
GianlucaFicarelli: The trigger checks `e1.authorized_public = TRUE OR e1.authorized_project_id = e2.authorized_project_id`
but I was able to create a public memodel while referencing a private emodel using the same project-id.
(e1.authorized_project_id = e2.authorized_project_id in that case)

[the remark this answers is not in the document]
g-bar: Ah that's true, thanks for catching it. It should be AND I think. I'll fix it and add a test.

app/db/model.py
g-bar: I changed it to morphology_id, kept emode_id , though.

app/filters/memodel.py
```diff
@@ -0,0 +1,38 @@
...
+from app.filters.morphology import MorphologyFilter
+
+
+class MEModelFilter(CustomFilter):
+    creation_date__lte: datetime | None = None
+    creation_date__gte: datetime | None = None
+    update_date__lte: datetime | None = None
+    update_date__gte: datetime | None = None
+    name__ilike: str | None = None
+    brain_location_id: int | None = None
```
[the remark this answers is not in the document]
g-bar: Changed here and in MorphologyFilter.

app/filters/memodel.py
```diff
@@ -0,0 +1,38 @@
...
+    brain_region_id: int | None = None
+    species_id__in: list[int] | None = None
+    valdation_status: ValidationStatus | None = None
```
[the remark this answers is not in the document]
g-bar: Fixed.

app/routers/memodel.py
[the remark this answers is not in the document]
g-bar: Updated the trigger to the above and added tests.

app/schemas/me_model.py
g-bar: Added.

tests/test_memodel.py
g-bar: Removed them.

app/routers/memodel.py
```diff
@@ -0,0 +1,178 @@
...
+        joinedload(MEModel.species),
+        joinedload(MEModel.strain),
+        joinedload(MEModel.emodel),
+        joinedload(MEModel.mmodel),
+        joinedload(MEModel.brain_region),
+        joinedload(MEModel.contributions).joinedload(Contribution.agent),
+        joinedload(MEModel.contributions).joinedload(Contribution.role),
+        joinedload(MEModel.mtypes),
+        joinedload(MEModel.etypes),
+        raiseload("*"),
```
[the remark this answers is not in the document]
g-bar: Yes, that is consistent with SQLAlchemy's [guidelines](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#what-kind-of-loading-to-use).

So I think entity to contributions, assets selectinload, but I think contributions to agent, role should be joinedload since those are many-to-one.

`selectinload(MEModel.contributions).joinedload(Contribution.role)`

app/schemas/emodel.py
```diff
@@ -16,7 +16,7 @@
 
 
 class ExemplarMorphology(CreationMixin, ReconstructionMorphologyBase):
-    pass
+    id: uuid.UUID
```
GianlucaFicarelli, line 19: It's just one field but we can inherit from IdentifiableMixin

app/schemas/me_model.py
```diff
@@ -13,7 +13,11 @@
     StrainRead,
 )
 from app.schemas.contribution import ContributionReadWithoutEntity
-from app.schemas.emodel import EModelBase, ExemplarMorphology as MModel
+from app.schemas.emodel import EModelBase, ExemplarMorphology as Morphology
+
+
+class EModel(EModelBase, CreationMixin):
+    id: uuid.UUID
```
GianlucaFicarelli, line 20: same here, it's just one field but we can inherit from IdentifiableMixin

GianlucaFicarelli (approved):
LGTM!

g-bar:
Thanks @GianlucaFicarelli !

PR #103, opened 2025-03-23
Add SingleNeuronSynaptome & SingleNeuronSynaptomeSimulation

eleftherioszisis:
* Add SingleNeuronSynaptome routers for read_one/create_one/read_many
* Add SingleNeuronSynaptomeSimulation routers for read_one/create_one/read_many
* Refactor router logic for read_one/create_one/read_many into app/routers/common.py for reuse


app/db/model.py
```diff
@@ -496,6 +495,22 @@ class SingleNeuronSimulation(LocationMixin, Entity):
     __mapper_args__ = {"polymorphic_identity": "single_neuron_simulation"}  # noqa: RUF012
 
 
+class SingleNeuronSynaptomeSimulation(DescriptionVectorMixin, LocationMixin, Entity):
+    __tablename__ = "single_neuron_synaptome_simulation"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    description: Mapped[str] = mapped_column(default="")
+    name: Mapped[str] = mapped_column(default="")
```
jdcourcol, line 502: Can we make name and seed mandatory (nullable = False and no default value )?

app/db/model.py
```diff
@@ -496,6 +495,22 @@ class SingleNeuronSimulation(LocationMixin, Entity):
...
+    seed: Mapped[int] = mapped_column(default=-1)
+    injectionLocation: Mapped[STRING_LIST] = mapped_column(default=[])
+    recordingLocation: Mapped[STRING_LIST] = mapped_column(default=[])
+    status: Mapped[SingleNeuronSimulationStatus] = mapped_column()
```
jdcourcol, line 506: same for status, should be mandatory

app/db/model.py
```diff
@@ -496,6 +495,22 @@ class SingleNeuronSimulation(LocationMixin, Entity):
...
+    synaptome_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("single_neuron_synaptome.id"), index=True
```
jdcourcol, line 508: can we make it mandatory ?

app/db/model.py
GianlucaFicarelli: sqlalchemy infers the nullability of the field from the type annotation (that is `str` and not `str | None` in this case), unless it's explicitly set with `nullable=`, so it's already mandatory. For confirmation, the generated migration file contains actually:
```
        sa.Column("name", sa.String(), nullable=False),
        sa.Column("seed", sa.Integer(), nullable=False),
```

app/db/model.py
eleftherioszisis: It already is. If I try creating a row without it, I get:
```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "synaptome_id" of relation "single_neuron_synaptome_simulation" violates not-null constraint
```

app/db/model.py
eleftherioszisis: Done.

app/db/model.py
jdcourcol: >sqlalchemy infers the nullability of the field from the type annotation
nice, thank you

app/routers/common.py
```diff
@@ -1,17 +1,55 @@
...
 
 
+def router_read_one(
+    *,
+    id_: uuid.UUID,
+    db: Session,
+    db_model_class: type[Entity],
+    authorized_project_id: uuid.UUID | None,
+    response_schema_class: type[BaseModel],
+    operations: list[Any],
```
g-bar, line 22: I think this can be typed as
```py
from sqlalchemy.sql.base import ExecutableOption

operations: list[ExecutableOption]
```

app/routers/common.py
```diff
@@ -1,17 +1,55 @@
...
+        ).filter(db_model_class.id == id_)
+        for operation in operations:
+            query = query.options(operation)
+
+        row = db.execute(query).unique().scalar_one()
+
+    return response_schema_class.model_validate(row)
+
+
+def router_create_one(
```
g-bar, line 37: Wonder if we can do the same for the query endpoints, they seem to follow the same pattern.

app/routers/common.py
```diff
@@ -1,17 +1,55 @@
...
+    db_model_class: type[Entity],
+    json_model: BaseModel,
+    authorized_project_id: uuid.UUID | None,
+    response_schema_class: type[BaseModel],
+):
+    data = json_model.model_dump() | {"authorized_project_id": authorized_project_id}
+    row = db_model_class(**data)
+    db.add(row)
+    db.commit()
+    db.refresh(row)
```
g-bar, line 49: Wonder if we can add this context manager here: https://github.com/openbraininstitute/entitycore/blob/09a39b74a14d9cabcb576198b0bb5dba30ad4433/app/routers/memodel.py#L77 when (if?) my MEModel PR gets merged?.

eleftherioszisis: Sure. I will wait for MEModel for this PR so that I can replace the placeholder MEModel definitions.

app/db/model.py
eleftherioszisis: @jdcourcol concerning the seed, may I remove the default too? I assume that every simulation has a seed defined for reproducibility.

jdcourcol: yes, no default and mandatory

app/routers/common.py
eleftherioszisis: Thanks, added.

app/routers/common.py
eleftherioszisis: I did a kind of dirty refactoring to minimize the code duplication but probably needs to be revisited.

app/routers/common.py
```diff
@@ -76,3 +81,106 @@ def __call__(self, q: sa.Select, vector_col: InstrumentedAttribute):
...
+        aliases=aliases,
+    )
+    filter_query = with_search(
+        filter_query,
+        db_model_class.description_vector,
+    )
+    distinct_ids_subquery = (
+        filter_model.sort(filter_query)
+        .with_only_columns(db_model_class)
+        .distinct()
```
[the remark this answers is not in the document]
eleftherioszisis: Yeah, I would suggest to first move more routers behind the common methods so that they are covered by the tests, and then we can try improving/refactoring them.

app/routers/common.py
eleftherioszisis: Is this context manager supposed to be applied to all create routers? In other words, does it make sense to pass it as an argument or it should always be enforced in the create route?

g-bar: So far we only use ensure_authorized_references those that don't use it would not raise error should be fine. But I'm not sure if we might need a different one in the future.

eleftherioszisis: Sure.

app/routers/common.py
```diff
@@ -76,3 +81,106 @@ def __call__(self, q: sa.Select, vector_col: InstrumentedAttribute):
...
+    response_schema_class: type[ListResponse],
+    name_to_facet_query_params: dict[str, FacetQueryParams],
+    filter_model: CustomFilter,
+):
+    filter_query = constrain_to_accessible_entities(
+        sa.select(db_model_class),
+        project_id=user_context.project_id,
+    )
+    for method_name, model_cls, condition in filter_query_operations:
+        filter_query = getattr(filter_query, method_name)(model_cls, condition)
```
g-bar, line 144: I think this would be clearer if instead of this list we passed a functoin that accepts the query as arg and returns the query with ops applied. Like so we avoid the for loop and the getattr. Maybe also for the data_query.

eleftherioszisis: Sure, changed it for both.

g-bar (approved):
LGTM

PR #105, opened 2025-03-24
Improve and add authentication to all the endpoints

GianlucaFicarelli:
- Add UserContext
- Add caching to reduce calls to KeyCloak
- Reuse the same instance of httpx.Client
- Add authentication to all the endpoints except /docs /health /version
- Allow to read public entities even when project-id isn't passed
- Introduce the role of service admin: service/entitycore/admin (needs updates to keycloak groups)
- Allow to write global resources only to the service admin role

More details in docs/authentication.md

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -1,0 +1,61 @@
+# Authentication
+
+## General Rules
+
+- All endpoints require authentication with a valid token, except for the [unauthenticated endpoints](#unauthenticated-endpoints).
+- For authenticated endpoints:
+  - If `virtual-lab-id` and `project-id` are provided in the request headers, they are validated with Keycloak using the given token.
+  - If `virtual-lab-id` is provided without `project-id`, it is verified.
+  - If `project-id` is provided without `virtual-lab-id`, authentication fails.
+- Endpoints for [global resources](#endpoints-for-global-resources):
+  - Do not require `virtual-lab-id` or `project-id`.
+  - **Read** endpoints are accessible to all authenticated users.
+  - **Write** endpoints require the user to be a member of the [service admin group](#service-admin-group).
+- Endpoints for [project resources](#endpoints-for-project-resources):
+  - **Read** endpoints do not require `virtual-lab-id` or `project-id`, but:
+    - If neither is provided, only public resources are returned.
+    - If both are provided, both public and private resources are returned.
+  - **Write** endpoints require both `virtual-lab-id` and `project-id`.
+
+## Service Admin Group
+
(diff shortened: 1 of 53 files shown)


eleftherioszisis (approved):
Looks great!

docs/authentication.md
```diff
@@ -0,0 +1,61 @@
...
+  - **Write** endpoints require both `virtual-lab-id` and `project-id`.
+
+## Service Admin Group
+
+To call endpoints that modify global resources, the user must belong to a special Keycloak group: `/service/entitycore/admin`.
+
+## Caching
+
+Requests to Keycloak are cached using a Time-aware Least Recently Used (TLRU) cache.
+Similar to a TTL cache, this method assigns an expiration time to each item. However, in a TLRU cache, expiration time is determined individually for each item based on the token's expiration time, capped at the configured maximum TTL.
```
g-bar, line 27: Do we need to handle the case where the user logs out or their token gets revoked?

GianlucaFicarelli: An alternative could be to subscribe to Keycloak admin events, and process the deletion of tokens.

Side note, the cache is local the the service instance. If we deploy multiple instances:
- the user info are cached separately (and each instance needs to call Keycloak once for each token/vlab-id/proj-id)
- we need to invalidate the local cache in each of them

However, I think we can postpone cache improvements for later.

GianlucaFicarelli: Opened https://github.com/openbraininstitute/entitycore/issues/106

g-bar: Sounds good for now.

tests/conftest.py
```diff
@@ -51,19 +64,122 @@ def _create_buckets(s3):
...
 
 
+@pytest.fixture
+def client_no_auth(session_client, _override_check_user_info):
+    """Return a web client instance, not authenticated."""
+    return session_client
+
+
+@pytest.fixture
+def client_1(client_no_auth):
```
eleftherioszisis, line 156: Wouldn't `client_admin` and `client_user` or something along those lines be more clear?

GianlucaFicarelli: Yes it's probably clearer, although the main difference is that they specify different project-ids.
An alternative could be to define `client_1` and `client_2` and `client_admin`, where 1 and 2 aren't admin, but maybe we don't need this level of differentiation.

eleftherioszisis: Sure, that works too.

eleftherioszisis (approved):

PR #107, opened 2025-03-27
Allow downloading directory files

eleftherioszisis:
Files can be downloaded from directory assets with:
```
GET /{entity_type}/{entity_id}/assets/{asset_id}/download?asset_path=file.ext
```
`asset_path` is mandatory when the asset is a directory

### DIFF
--- a/app/routers/asset.py
+++ b/app/routers/asset.py
@@ -1,6 +1,7 @@
 """Generic asset routes."""
 
 import uuid
+from http import HTTPStatus
 from pathlib import Path
 from typing import Annotated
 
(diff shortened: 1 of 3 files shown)


app/routers/asset.py
```diff
@@ -127,8 +129,29 @@ def download_entity_asset(
...
+            raise ApiError(message=msg, error_code=ApiErrorCode.INVALID_REQUEST)
+        full_path = str(Path(asset.full_path, asset_path))
+    else:
+        if asset_path:
+            msg = "asset_path is only applicable when asset is a directory"
+            raise ApiError(
+                message=msg,
+                error_code=ApiErrorCode.INVALID_REQUEST,
+            )
+        full_path = asset.full_path
```
GianlucaFicarelli, line 150: Instead of returning http status code 400, we could return a more specific code, for example:
 
- if the asset is a directory and asset_path is not specified: 409 Conflict (the client thinks it's a file)
- if the asset is a file and asset_path is specified: 409 Conflict (the client thinks it's a directory)
- if asset_path isn't well formed: 422 Unprocessable Content (as a validation error)

Alternatively, we could return 422 for all of them, although the first two aren't exactly validation errors.
If we change it, we could update also the http status code returned by `upload_entity_asset` to make it consistent.

eleftherioszisis: I updated the errors. Is it better now?

GianlucaFicarelli (approved):

PR #109, opened 2025-03-31
Common endpoints

g-bar:


app/routers/common.py
```diff
@@ -1,139 +1,94 @@
...
+        if apply_operations:
+            query = apply_operations(query)
 
         row = db.execute(query).unique().scalar_one()
 
     return response_schema_class.model_validate(row)
 
 
-def router_create_one(
+def router_create_one[T: BaseModel](
```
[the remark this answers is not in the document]
g-bar: I tried parametrizing by [M: DeclarativeBase] too, but noticed that sq alchemy cannot infer the return type of db.execute so it wasn't very useful. 

What this one does is allows mypy to infer the type of the result on the caller to be 
ListReponse[EModelRead] for example instead of ListResponse[BaseModel]

app/routers/common.py
```diff
@@ -1,139 +1,94 @@
...
-    apply_filter_query_operations: Callable,
-    apply_data_query_operations: Callable,
+    with_search: Search,
+    facets: WithFacets,
+    aliases: Aliases,
+    apply_filter_query_operations: ApplyOperations,
+    apply_data_query_operations: ApplyOperations,
     pagination_request: PaginationQuery,
-    response_schema_class: type[ListResponse],
+    response_schema_class: type[ListResponse[T]],
```
eleftherioszisis, line 91: The Generic is so that mypy is happy with this, isn't it?

g-bar: Commented above

> What this one does is allows mypy to infer the type of the result on the caller to be 
> ListReponse[EModelRead] for example instead of ListResponse[BaseModel]

app/routers/common.py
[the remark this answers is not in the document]
eleftherioszisis: Yeah, I had issues with making that work too and ended up not specifying a subtype. Would it be ok to limit it to `read_many`, where it is needed for ListResponse?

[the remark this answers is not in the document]
g-bar: It's helful in the other ones too. Without the generic if you look at the inferred type on the emodel read one handler you get just BaseModel, with the generic you get EModelRead.

Capture d’écran 2025-03-31 à 13 52 26

Capture d’écran 2025-03-31 à 13 52 50

app/routers/common.py
```diff
@@ -174,7 +128,7 @@ def router_read_many(
     ).scalar_one()
 
     response = response_schema_class(
-        data=data,
+        data=cast("list[T]", data),
```
GianlucaFicarelli, line 131: Is the cast needed also if we add `.all()` to the query above?

app/routers/common.py
```diff
@@ -1,139 +1,94 @@
 import uuid
```
GianlucaFicarelli, line 1: I would move this file to app/service/common.py, because it's business logic that can be reused outside of the routers. The separation into 2 levels (routers/service) should be enough for the moment.

app/routers/common.py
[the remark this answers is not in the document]
eleftherioszisis: Ah I see. ok then, nvm :)

app/routers/memodel.py
```diff
@@ -58,45 +57,34 @@ def memodel_joinedloads(select: Select):
...
-        query = memodel_joinedloads(sa.select(MEModel).filter(MEModel.id == db_em.id))
-        return db.execute(query).unique().scalar_one()
+    return router_create_one(
+        db=db,
+        db_model_class=MEModel,
+        authorized_project_id=user_context.project_id,
+        response_schema_class=MEModelRead,
+        json_model=memodel,
+        apply_operations=memodel_joinedloads,
+        context_manager=ensure_authorized_references(
```
eleftherioszisis, line 84: In the end is this something that should be applied to all routers, or specific only to this one?

app/routers/memodel.py
```diff
@@ -47,8 +46,8 @@ def memodel_joinedloads(select: Select):
         joinedload(MEModel.emodel),
         joinedload(MEModel.morphology),
         joinedload(MEModel.brain_region),
-        selectinload(MEModel.contributions).joinedload(Contribution.agent),
-        selectinload(MEModel.contributions).joinedload(Contribution.role),
+        joinedload(MEModel.contributions).joinedload(Contribution.agent),
+        joinedload(MEModel.contributions).joinedload(Contribution.role),
```
GianlucaFicarelli, line 50: Why this has been changed from selectinload to joinedload?

app/routers/memodel.py
g-bar: I think for those that reference other entities, (i.e. those that use this trigger https://github.com/openbraininstitute/entitycore/blob/5df282596ea6b57a55fe18396590a71d3da1a660/app/db/triggers.py#L73) 
We could apply it everywhere since those that don't use the trigger wouldn't raise that exception, but I'm thinking it's helpful to parametrize if later some other routers use a different context_manager.

app/dependencies/common.py
```diff
@@ -55,4 +59,69 @@ def traverse(dependant: Dependant, params: dict[str, bool]) -> dict[str, bool]:
         )
 
 
+class FacetQueryParams(TypedDict):
+    id: InstrumentedAttribute[uuid.UUID] | InstrumentedAttribute[int]
+    label: InstrumentedAttribute[str]
+    type: NotRequired[InstrumentedAttribute[str]]
+
+
+def _get_facets(
```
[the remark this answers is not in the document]
GianlucaFicarelli: Thinking mroe about it, maybe it's better if we keep "service" for the intermediate level, and move the functions interacting with the queries in a separate module. "repository" is used for queries, but it follows a different pattern.
Maybe we can add a new module "query" or "queries", thoughts?

app/routers/common.py
```diff
@@ -147,20 +102,19 @@ def router_read_many(
...
-        filter_query,
-        db_model_class.description_vector,
-    )
-    distinct_ids_subquery = (
+
+    if issubclass(db_model_class, DescriptionVectorMixin):
+        filter_query = with_search(
+            filter_query,
+            db_model_class.description_vector,  # type:ignore[attr-defined]
+        )
```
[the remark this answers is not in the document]
g-bar: Yes, much better.

app/dependencies/common.py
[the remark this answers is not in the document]
g-bar: Queries sounds good to me, although not a strong preference. Maybe you can help with this change? I did this because you mentioned in a previous PR to move this away from app/routers/common.py , but I haven't put as much thought into it.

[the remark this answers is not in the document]
GianlucaFicarelli: Sure, let's keep those changes for a separate PR, I'll open it after this one is merged.

app/routers/memodel.py
g-bar: This one is used by the read_one endpoint, I defined another one below for the list endpoint which uses `selectin`.

My thinking is that for non-list view joinedload is ok for one-to-many, because one of the drawbacks of joinedload is that it will multiply the total rows that it fetches in a MxN way where M is the fetched selected rows and N the number of rows to join per record.

For one record you just have 1XN, so it's OK, otherwise you would do two queries one to fetch the MEModel and one to fetch the contributions in a  "select contributions where id in [list of ids]" fashion but the list of ids only contains 1 item.

GianlucaFicarelli: Yeah ok, we should verify if there is a difference in performance. If it's the same, I would lean towards using the same options in both the endpoints just for simplicity, but let's keep it as is for the moment.

app/routers/common.py
GianlucaFicarelli: To be done in a separate PR (see other comment)

eleftherioszisis (approved):

PR #111, opened 2025-04-01
Move query logic to service and queries modules

GianlucaFicarelli:
- Move the logic from routers to service/queries, to be able to reuse the functions.
- Add Identifiable abstract class to the db model, so it can be used for any class that has an id but it's not an entity
- Changes to routers/common.py that has been moved to service/common.py, to be able to use it also for Identifiables


app/queries/common.py
```diff
@@ -118,13 +118,15 @@ def router_read_many[T: BaseModel](
     ).scalar_one()
 
     response = response_schema_class(
-        data=cast("list[T]", data),
+        data=data,
```
g-bar, line 121: Nice, in main it shows an error without the cast I wonder what made the difference?

GianlucaFicarelli: Yeah, I noticed that it wasn't needed anymore, but I cannot rollback to a version with the error 🤔 
According to pycharm, `data` is `ScalarResult[Row | RowMapping]`
and `data.all()` would be `Sequence[Row | RowMapping]`
But using `reveal_type` with mypy, it just says `Any`.

app/db/model.py
```diff
@@ -83,16 +87,14 @@ class BrainRegion(TimestampMixin, Base):
     children: Mapped[list[int] | None] = mapped_column(ARRAY(BigInteger))
 
 
-class Species(TimestampMixin, Base):
+class Species(TimestampMixin, Identifiable):
```
eleftherioszisis, line 90: Could TimestampMixin be part of Identifiable? I would be in favor of all resources without exception to have timstamp properties.

app/service/brain_region.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def read_one(db: SessionDep, id_: int) -> BrainRegionRead:
+    with ensure_result(error_message="Brain region not found"):
+        stmt = sa.select(BrainRegion).filter(BrainRegion.id == id_)
+        row = db.execute(stmt).scalar_one()
+    return BrainRegionRead.model_validate(row)
+
+
+def create(brain_region: BrainRegionCreate, db: SessionDep) -> BrainRegionRead:
```
eleftherioszisis, line 54: Just for symmetry, could this be renamed to create_one? Who knows maybe a create_many will exist in the future :D

app/db/model.py
GianlucaFicarelli: Adding TimestampMixin to Identifiable would cause some changes:

- `creation_date` and `update_date` would be added to the `root` table
- `creation_date` and `update_date` would be added to `measurement` and `measurement_serie_element` (they are identifiable but they don't inherit from root)
- `creation_date` and `update_date` would be removed from all the tables inheriting from `root` (that are only `entity` and `agent`)

Since it seems that root is always already joined in the queries, I don't think that the change would affect performances, but I would keep this change for a separate PR because this one doesn't modify the database.

app/db/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    ExperimentalBoutonDensity,
+    ExperimentalNeuronDensity,
+    ExperimentalSynapsesPerConnection,
+    Mesh,
+    ReconstructionMorphology,
+    SingleCellExperimentalTrace,
+)
+from app.db.types import EntityType
+
+ENTITY_TYPE_TO_CLASS: dict[EntityType, type[Entity]] = {
```
eleftherioszisis, line 13: Is there possibly a way to avoid having more maps and places that need to be updated in general when we add a new db/schema/router path?

app/db/model.py
eleftherioszisis: Indeed, it can go in a separate PR. Concerning the measurements, is there a good reason for them not to be timestamped?

GianlucaFicarelli: AFAICT the timestamps of `measurement_serie_element` should all be the same for a specific `measurement`,  and all the timestamps of `measurements` should be the same for a specific `morphology_feature_annotation`, so it would be a waste of space to add a timestamp to every record.
Moreover, I think that these tables can become big, also changing the id from int to uuid had some impact on the size.
Maybe we could rethink how these data are stored in the db, depending on how they are used (do we need to search/filter on them? I would expect that we just need to store and retrieve them altogether for a specific annotation)

app/db/utils.py
GianlucaFicarelli: Yeah I'm not keen to add other maps either. In the asset router we need to get the table class from the endpoint name. Alternatively, this mapping can be automatically generated with this:
```
ENTITY_TYPE_TO_CLASS = {entity_type: Base.metadata.tables[entity_type.name] for entity_type in EntityType}
```
but it needs to stay in a different file than types.py as it is now, to avoid circular imports.

GianlucaFicarelli: Or I can avoid adding the mapping, and get it dynamically using Base.metadata.tables, maybe this is better

eleftherioszisis: Unless the map is needed in its entirety up-front I would favor lazily getting it from Base.metadata.tables.

app/db/model.py
g-bar: Do we need root at all?
Agent, Entity inherit from it through polymorphism but it adds no fields.

and it just causes more joins and the need for aliases some times.

What is its purpose ?, seems like a remnant form the legacy endpoints but those will be gone are some points, won’t they?

app/queries/common.py
g-bar: I wonder if we could switch to pyright to improve type inference and type safety overall. 
See: https://github.com/openbraininstitute/entitycore/pull/113

app/routers/emodel.py
```diff
@@ -1,147 +1,12 @@
...
-        apply_filter_query_operations=filter_query_operations,
-        apply_data_query_operations=load,
-        pagination_request=pagination_request,
-        response_schema_class=ListResponse[EModelRead],
-        name_to_facet_query_params=name_to_facet_query_params,
-        filter_model=emodel_filter,
-    )
+read_many = router.get("")(app.service.emodel.read_many)
+read_one = router.get("/{id_}")(app.service.emodel.read_one)
+create = router.post("")(app.service.emodel.create)
```
g-bar, line 12: Not sure I’m a fan of  these very thin wrappers with no logic that do little and add boilerplate. Perhaps we have one layer too much? 
I understand  the need for separation of concerns but  I wonder if we’re leaning too much into over-modularisation.

I had a similar  discussion with Bilal over at virtual-lab-manager, so it’s entirely  possible that I’m the one with the flawed perspective here, so feel free to ignore if my comments if people like it this way.

app/queries/common.py
```diff
@@ -1,102 +1,100 @@
...
+    data = json_model.model_dump()
+    if issubclass(db_model_class, Entity):
+        data |= {"authorized_project_id": authorized_project_id}
+    row = db_model_class(**data)
+    with (
+        ensure_uniqueness(error_message=f"{db_model_class.__name__} already exists"),
+        ensure_authorized_references(
+            f"One of the entities referenced by {db_model_class.__name__} "
+            f"is not public or not owned by the user"
+        ),
```
g-bar, line 56: Better this way but we lose the custom message which is useful, any chance we could the message parametrizable, theoretically it's also possible to the the fields tha referenece other entitys from the db_model_class with `sqlalchemy.inspect` but that feels like overkill here.

app/routers/emodel.py
eleftherioszisis: I had similar feelings and as a matter of fact I had a brief discussion with @GianlucaFicarelli that we can completely generate this default set of routes automatically so that we can focus only on the service side. But that would mean things would be come a bit more implicit, with which I am fine if they reduce the amount of trivial work that needs to be done when adding new logic.

app/queries/common.py
GianlucaFicarelli: Even with the custom message we don't know which one of the referenced entities causes the issue, so I don't see a lot of added value (the list of referenced entities depend only on the table definition, so it could be static, or dynamically generated with inspect)

g-bar: Yeah good point.

app/db/model.py
GianlucaFicarelli: It was useful to ensure a global unique id when the ids were integers, but now they are uuids. It could still be useful:
- if we want to retrieve a resource by id without knowing if it's an agent or an entity, but do we need it?
- if there is any table that needs a foreign key with root, regardless if it's an agent or an entity, but I don't see any at the moment.

Btw this is for discussion, but it can be changed in a separate PR as well.

app/queries/common.py
g-bar: Shoud be also possible to write the hook so that it raises an error message with the offending reference, but that seems like overkill, so let's keep it as is for now.

app/db/utils.py
GianlucaFicarelli: Actually, `Base.metadata.tables` gives a Table, not the class that we need.
The simplest way I found to generate the dict dynamically is:
```python
{
    EntityType[mapper.class_.__tablename__]: mapper.class_
}
```
This can go inside `service/entity.py`, since it's used only there at the moment.
Feel free to suggest a simplest way :)

g-bar (approved):

eleftherioszisis (approved):

PR #112, opened 2025-04-01
Add experimental graphql endpoint

GianlucaFicarelli:
Add an experimental graphql endpoint, reusing the existing service logic.

- Accessible at http://127.0.0.1:8000/graphql
- Only species and morphologies can be retrieved and modified.
- Graphql would have its own error handling according to the specs, but it is temporarily bypassed

This is a showcase, not to be merged.


g-bar:
Really cool, thanks for the showcase.

PR #113, opened 2025-04-02
Pyright

g-bar:
I have noticed that Mypy is not very poweful in terms of type inference it seems to give up and resort to Any in cases where 
[pyright](https://github.com/microsoft/pyright) is able to infer the types.

Also we could improve our types by making use of the fact that `Select` is generic.

With the above  I was able to parametrize the `router_read_many` et al. functions by `M: Identifiable` and pyright is able to 
infer the type of the data all the way down to `ScalarResult[M@router_read_many]`

I ran pyright on the `app` folder and theare are just 43 errors, so the switch could be done without too much effort.

What do you think @GianlucaFicarelli  @eleftherioszisis  should I create a ticket to switch to Pyright?


app/queries/common.py
```diff
@@ -117,8 +117,12 @@ def router_read_many[T: BaseModel](
...
     ).scalar_one()
 
-    response = response_schema_class(
-        data=data,
+    # reveal_type(data) # information: Type of "data" is "ScalarResult[M@router_read_many]"
+
+    response = ListResponse(
+        data=[
+            response_schema_class.model_validate(row) for row in data
+        ],  # Manual validation to please pyright
```
g-bar, line 125: Pyright is able to correctly infer the types here.

GianlucaFicarelli (commented):
I'm not against switching to pyright if it provides more detailed checks, unless this forces us to add more "ignore" directives to please it (there are a couple of them also in this draft).
For reference, I had a look at https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

app/queries/common.py
g-bar: Potentially overhead yes unless Pydantic optimizes it and skips validation if you already pass instances of the model. 

We could use the cast in this case, instead.

g-bar:
> I'm not against switching to pyright if it provides more detailed checks, unless this forces us to add more "ignore" directives to please it (there are a couple of them also in this draft). For reference, I had a look at https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

I went ahead and fixed the remaining errors and none extra are needed, it's just those two because fastapi-filter doesn't use generics for Select and we cannot change that code.

app/schemas/auth.py
```diff
@@ -30,10 +40,9 @@ class UserContext(BaseModel):
     is_service_admin: bool = False
```
GianlucaFicarelli, line 40: UserContext contains some attributes already defined in the same way in UserContextBase.

app/service/contribution.py
```diff
@@ -32,7 +33,7 @@ def read_many(
     total_items = db.execute(query.with_only_columns(sa.func.count(Contribution.id))).scalar_one()
 
     response = ListResponse[ContributionRead](
-        data=data,
+        data=cast("list[ContributionRead]", data),
```
GianlucaFicarelli, line 36: An alternative to cast could be to call `model_validate`on ListResponse, and pass a dict.
This would probably bypass the pyright check (but we are doing the same by using cast)

app/queries/common.py
```diff
@@ -117,8 +117,8 @@ def router_read_many[T: BaseModel](
         )
     ).scalar_one()
 
-    response = response_schema_class(
-        data=data,
+    response = ListResponse(
+        data=[response_schema_class.model_validate(row) for row in data],
```
GianlucaFicarelli, line 121: If we specify `ListResponse[response_schema_class]` instead of `ListResponse`, we shouldn't need to call model_validate on each item.
```python
    response = ListResponse[response_schema_class](
        data=data,
```
or:
```python
    response = ListResponse[response_schema_class].model_validate(
        {"data": data, ...}
```
Btw, the overhead of calling model_validate for each time should be negligible, because it seems that Pydantic doesn't copy the objects inside the list.

app/filters/base.py
```diff
@@ -73,3 +73,19 @@ def filter(self, query: Query | Select, aliases=None):
...
+            direction = Filter.Direction.asc
+            if field_name.startswith("-"):
+                direction = Filter.Direction.desc
+            field_name = field_name.replace("-", "").replace("+", "")  # noqa: PLW2901
+
+            order_by_field = getattr(self.Constants.model, field_name)
+
+            query = query.order_by(getattr(order_by_field, direction)())
+
+        return query
```
GianlucaFicarelli, line 91: Do we need to redefine `sort` for pyright? The content seems the same as in the base class, except for the annotations.

app/service/contribution.py
g-bar: Would you mind keeping the cast?. 

Between:
```py
response = ListResponse[ContributionRead](
        data=cast("list[ContributionRead]", data),
        pagination=PaginationResponse(
            page=pagination_request.page,
            page_size=pagination_request.page_size,
            total_items=total_items,
        ),
        facets=None,
    )
```
and 
```py
response = ListResponse[ContributionRead].model_validate(
        {
            "data": data,
            "pagination": PaginationResponse(
                page=pagination_request.page,
                page_size=pagination_request.page_size,
                total_items=total_items,
            ),
            "facets": None,
        }
    )
```

probably the second is even less readable due to the extra nested {}, more verbose and we're bypassing pyright either way, plus this woul need to be changed in several places.

app/queries/common.py
g-bar: The first still raises the error so I went with the second one.

app/queries/common.py
```diff
@@ -12,18 +12,18 @@
 from app.filters.base import CustomFilter
 from app.schemas.types import ListResponse, PaginationResponse
 
-ApplyOperations = Callable[[sa.Select], sa.Select]
+type ApplyOperations[T: DeclarativeBase] = Callable[[sa.Select[tuple[T]]], sa.Select[tuple[T]]]
 Aliases = dict[type[Root], type[Root]]
 
 
-def router_read_one[T: BaseModel](
+def router_read_one[T: BaseModel, M: Identifiable](
```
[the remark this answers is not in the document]
g-bar: M stood for Model cause I was thinking DeclarativeBase, changed to I.

app/filters/base.py
g-bar: Yes, better.

app/service/contribution.py
g-bar: Alternatively we can just do a type:ignore there.

GianlucaFicarelli: Yeah I don't like the dict approach either, but even with cast we are lying to pyright.
In the pydantic docs this is mentioned at https://docs.pydantic.dev/latest/integrations/visual_studio_code/#strict-errors that there are some workarounds, none of them very satisfactory:

1. Disable type checks in a line  (Use # type: ignore)
2. Override the type of a variable (use Any)
3. Override the type of a value with cast

GianlucaFicarelli: Ok let's keep the cast

app/schemas/types.py
```diff
@@ -37,6 +37,15 @@ class Facet(BaseModel):
...
+    # of any type that we're wanting to validate, not necessarily list[M].
+    #
+    # At runtime, 'data' is explicitly defined as list[M] so that Pydantic
+    # can enforce validation, ensuring that 'data' is always a list of M instances.
+    if TYPE_CHECKING:
+        data: Any  # Allows flexibility for static type checkers
+    else:
+        data: list[M]  # Ensures Pydantic validation at runtime
     pagination: PaginationResponse
     facets: Facets | None = None
```
g-bar, line 51: Alternatively we could do this, maybe this is better?:

```
    # Pre-validation, data passed to the model can be Any. Post-validation,
    # Pydantic ensures it matches the annotated types (e.g., list[M]),
    # so we use Any as input type annotation to avoid type checker errors before validation.
    def __init__(self, **data: Any):  # noqa: D107
        super().__init__(**data)

    data: list[M]
    pagination: PaginationResponse
    facets: Facets | None = None

```

eleftherioszisis:
pyright seems more complete than mypy. We can try it out and see how it goes.

app/schemas/types.py
GianlucaFicarelli: This has the advantage that it works for any attribute, but it invalidates the autocompletion of parameters in pycharm, besides bypassing the type validation 🤔

g-bar: It doesn't in VSCode it should be possible to configure Pycharm to use Pyright under the hood.
Capture d’écran 2025-04-04 à 15 00 29

As for bypassing type validation, it allows any type to be passed to the model, yes. That makes sense because you. might have any othe data type that you want to validate that conforms to the model.

But we get type validation for the resulting validated model, for example:

```py
a: int = res.data #  Error: The type list[T] is not assignable to int
b: list[T] = res.data. # Doesnt raise errors.
```

I think not allowing any here is just an artifact by the fact that the pydantic BaseModel metaclass turns the model into a dataclass post validation, more than by expclicit intent but even the init of models is typed as Any. See: https://github.com/pydantic/pydantic/blob/main/pydantic/main.py#L243

I believe this to be more reflective of the actual types. Indeed, what use is a model that you already have to pass the valid types to?

We could even extend the concept further and have a custom base model from which all our models inherit, and you could have simple do Model(data) instead of Model.model_validate(data) everywhere while preserve the post-validation type.

GianlucaFicarelli: If I overwrite `__init__` as in the comment, I don't get an error from pyright even if I write the code below, that's assigning an invalid attribute and it's missing the required attributes, so we'll lose some useful validation in this way:

```python
_x = ListResponse[Facet](
    aaa=["asdas"],
)
```

g-bar: My previous comment doesn't in fact the type checks are more strict than the `if TYPE_CHECKING` version.

GianlucaFicarelli: Since I cannot see an easy and generic solution that can work for any type of model, I'm reconsidering the alternative of calling `model_validate` for each element of the list.
I did some basic tests not very accurate, but that should give an idea of the overhead, and it seems negligible for small numbers of rows.

To instantiate a ListResponse with 1,000 morphologies:
- 0.073-0.081 seconds using data=data
- 0.074-0.085 seconds using data=[response_schema_class.model_validate(i) for i in data]

To instantiate a ListResponse with 100,000 morphologies (a lot more than what we should return in a single call):
- 4.60-4.67 seconds using data=data
- 4.69-4.72 seconds using data=[response_schema_class.model_validate(i) for i in data]

The results don't look too bad.

g-bar: Agree, ultimately it's the cleanest solution. If the perfomance impact is almost negligible, let's roll with it.

GianlucaFicarelli (commented):
Now it requires fixing the conflicts but it LGTM

GianlucaFicarelli (approved):

PR #114, opened 2025-04-02
allow morphology get by id to retrieve by legacy id

bilalesi:
1. allow `GET` morphology by id to retrieve by legacy id
2. add type to `ReconstructionMorphologyRead` (this is needed in the UI for every entity)

**NOTE**: the `GET /{id_}` may not conform to your API design, please lemme know if it should be in another endpoint

a use case to use  the `legacy id`:
in the generalization section (this part still use nexus) of the the details page, we list the morpho-similarities of a morphology, the user can click on the morphology card to open the details page, and since we use now the entitycore ID we have to get the mapped id-legacyid from entitycode

Screenshot 2025-04-02 at 11 18 30

### DIFF
--- a/app/routers/morphology.py
+++ b/app/routers/morphology.py
@@ -1,8 +1,10 @@
+import urllib.parse
 import uuid
+from http import HTTPStatus
 from typing import Annotated
 
 import sqlalchemy as sa
-from fastapi import APIRouter, Query
+from fastapi import APIRouter, Depends, Path, Query
 from fastapi_filter import FilterDepends
 from sqlalchemy.orm import (
     aliased,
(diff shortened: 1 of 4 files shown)


app/routers/morphology.py
```diff
@@ -213,3 +166,92 @@ def morphology_query(
...
+        if is_legacy:
+            decoded_id = urllib.parse.unquote(id_)
+            query = constrain_to_accessible_entities(
+                sa.select(ReconstructionMorphology), user_context.project_id
+            ).filter(decoded_id == sa.any_(ReconstructionMorphology.legacy_id))
+        else:
+            uuid_id = uuid.UUID(id_)
+            query = constrain_to_accessible_entities(
+                sa.select(ReconstructionMorphology), user_context.project_id
+            ).filter(ReconstructionMorphology.id == uuid_id)
```
g-bar, line 224: I would prefer another enpoint instead of the conditional logic here, as a matter of fact @GianlucaFicarelli just moved the logic to /service folder in this https://github.com/openbraininstitute/entitycore/pull/111, so maybe we can do the change there. That would also obviate the need for the validate_id dependency.

bilalesi: yeah good to know, 
i did not know that there is such a change, actually I like moving to repos/services
i will wait for the 111-PR to be merged and update this one accordingly

eleftherioszisis: query wouldn't work with the legacy_id perhaps instead of adding another endpoint?

g-bar: Ah that would work too, as a matter of fact, technically we don't need get /{id} either and we could add id to filter, he.

PR #115, opened 2025-04-03
Improve db model

GianlucaFicarelli:
Based on the comments on https://github.com/openbraininstitute/entitycore/pull/111, some changes open for further discussion:

- Remove the `Root` table, and move `type` to `Entity`, `Agent`, `AnnotationBody` (this last one is used only for `datamaturity_annotation_body`, because mtype and etype now have their own tables)
- Add `TimestampMixin` to `Identifiable`.
- Change type from `str` to `StrEnum` in all the polymorphic tables (should have better performance and use less storage, although it requires a db migration if we need to add new types)
- Revert `MorphologyMeasurement` and `MorphologyMeasurementSerieElement` to use int (bigint) instead of UUID. These ids shouldn't be publicly exposed, so it seems safe to handle them as int. If we need to copy the records from prod to staging, the ids can be different, as long as the referential integrity is correct.
- Change `EntityType` from `HyphenatedStrEnum` to `StrEnum` (the hyphens are needed only for the endpoints, so the conversion can be defined in the routers)


app/db/types.py
```diff
@@ -34,25 +34,37 @@ def process_result_value(self, value, dialect):  # noqa: ARG002, PLR6301
...
 
-class EntityType(HyphenStrEnum):
-    """Entities that are directly exposed through the API.
-
-    For each entry:
-
-    - name (underscore separated): for table names and S3 keys
-    - value (hyphen separated): for endpoints
-    """
+class EntityType(StrEnum):
```
eleftherioszisis, line 37: Does the type stored in the entity have to be externally defined here? At this point we have the class name, the route name, the table name, and this type that looks like it is the table name.

Couldn't we for example specify the type of a db model automatically either by the class or the table name? Why do we need to store it here explicitly and therefore having to remember to update this. It would also reduce the number of times we need to repeat the name of the table.

eleftherioszisis: Yeah, or maybe `EntityType.emodel.value`

eleftherioszisis (approved):
LGTM

g-bar:
I think without the root we shouldn't need the Agent aliases on some of the queries.

alembic/versions/20250404_102955_72c5253bf9b1_default_migration_message.py
g-bar, line 1: Could we all the way and merge all migrations into one file?

GianlucaFicarelli: Now the second file just applies the triggers, that need to be generated separately because of the order by which the triggers and the create tables are automatically generated.
We could manually merge them, but I think that's still possible that we'll have to drop the migrations again, so not so useful to do it right now.

g-bar: I don't really understand why, it seems to me that alembic should be able to automatically generate the tables then the triggers. So, let's keep it this way.

GianlucaFicarelli: The reason is that migration of triggers isn't directly handled by alembic, but by alembic-utils.
There is an explanation at `https://github.com/olirice/alembic_utils/issues/79`

app/service/memodel.py
```diff
@@ -31,6 +32,9 @@
 from app.schemas.me_model import MEModelCreate, MEModelRead
 from app.schemas.types import ListResponse
 
+if TYPE_CHECKING:
+    from app.filters.base import Aliases
```
g-bar, line 36: Nitpick: 

I don't think this is all that useful, the Aliases object would still be loaded into memory when app.filters.base is imported elsewhere?

g-bar (approved):

g-bar:
I left a couple nitpicks but it LGTM otherwise.

app/service/memodel.py
GianlucaFicarelli: To be honest I don't care too much if Aliases is always imported or not, because it happens only once at the application start.
But without the type-checking block, ruff complains with TC001 Move application import `app.filters.base.Aliases` into a type-checking block.

g-bar: Ahh, ok. :)

app/filters/base.py
```diff
@@ -2,7 +2,9 @@
 from fastapi_filter.contrib.sqlalchemy.filter import _orm_operator_transformer  # noqa: PLC2701
 from pydantic import field_validator
 from sqlalchemy import Select, or_
-from sqlalchemy.orm import Query
+from sqlalchemy.orm import DeclarativeBase, Query
+
+type Aliases[T: DeclarativeBase] = dict[type[T], type[T]]
```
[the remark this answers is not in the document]
GianlucaFicarelli: I was getting an error when using `dict[type[Identifiable], type[Identifiable]]`, but it doesn't happen anymore.
I'll change that.

PR #120, opened 2025-04-08
Use NameDescriptionVectorMixin in all the entities

GianlucaFicarelli:
This should ensure that all the entities have name, description, and the related indexes.

`name` and `description` are not added directly to the entity table, because it would be beneficial only if we want to search across different types of entities, but I don't think we have this use case.

If this is true also for other fields, we may consider to move them from entity to the subclasses, but need to check how to do it cleanly. The hypothesis to be verified is that smaller indexes on separate tables should perform better than a single index on the entity table (not sure if the join with entity can be avoided, though).


jdcourcol (commented):
good

eleftherioszisis (approved):

g-bar (approved):
LGTM

PR #121, opened 2025-04-09
Import emodel artifacts

g-bar:
Schemas and import script.


app/db/model.py
```diff
@@ -573,6 +573,54 @@ class ExperimentalSynapsesPerConnection(LocationMixin, SpeciesMixin, LicensedMix
...
+
+class IonChannelModel(DescriptionVectorMixin, LocationMixin, SpeciesMixin, Entity):
+    __tablename__ = EntityType.ion_channel_model.value
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+
+    name: Mapped[str] = mapped_column(index=True)
+    description: Mapped[str] = mapped_column(default="")
+    identifier: Mapped[str]
+    modelId: Mapped[str]
```
g-bar, line 601: These are the only cases where name, identifier and modelId are not all equal:
I guess we delete identifier and modelId and keep name?

```sql
      name      |     identifier     |      modelId
----------------+--------------------+--------------------
 CaDynamics_DC0 | CTX_CaDynamics_DC0 | CTX_CaDynamics_DC0
 Ca_HVA2        | CTX_Ca_HVA2        | CTX_Ca_HVA2
 Ca_LVAst       | CTX_Ca_LVAst       | CTX_Ca_LVAst
 SK_E2          | CTX_SK_E2          | CTX_SK_E2
 StochKv3       | CTX_StochKv3       | CTX_StochKv3
 NaTg           | CTX_NaTg           | CTX_NaTg
 SKv3_1         | CTX_SKv3_1         | CTX_SKv3_1
 K_Pst          | CTX_K_Pst          | CTX_K_Pst
 K_Tst          | CTX_K_Tst          | CTX_K_Tst
 KdShu2007      | CTX_KdShu2007      | CTX_KdShu2007
 Ih             | CTX_Ih             | CTX_Ih
 Nap_Et2        | CTX_Nap_Et2        | CTX_Nap_Et2
 TC_ih_Bud97    | Th07-mmb-framework | Th07-mmb-framework
 TC_Nap_Et2     | Th02-mmb-framework | Th02-mmb-framework
 TC_iT_Des98    | Th06-mmb-framework | Th06-mmb-framework
 TC_iA          | Th04-mmb-framework | Th04-mmb-framework
 TC_iL          | Th05-mmb-framework | Th05-mmb-framework
 TC_HH          | Th01-mmb-framework | Th01-mmb-framework
 TC_cad         | Th03-mmb-framework | Th03-mmb-framework
```

app/db/model.py
```diff
@@ -573,6 +573,54 @@ class ExperimentalSynapsesPerConnection(LocationMixin, SpeciesMixin, LicensedMix
...
+
+class IonToIonChannelModel(Base):
+    __tablename__ = "ion__ion_channel_model"
+
+    ion_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("ion.id", ondelete="CASCADE"), primary_key=True
+    )
+    ion_channel_model_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey(f"{EntityType.ion_channel_model}.id", ondelete="CASCADE"), primary_key=True
+    )
```
g-bar, line 590: This is needed because each IonChannelModel may contain a lis of ions.

app/db/model.py
darshanmandge: I think we can delete the `identifier` and `modelId` entries. Thanks.

app/cli/utils.py
```diff
@@ -326,3 +368,142 @@ def find_part_id(data: dict[str, Any], type_: Literal["NeuronMorphology", "EMode
...
+        db_ion = Ion(name=ion.get("label"))
+        db.add(db_ion)
+        db.flush()
+
+    _cache[label] = db_ion.id
+
+    return db_ion.id
+
+
+def import_ion_channel_model(  # noqa: PLR0914
```
g-bar, line 390: This imports the hoc files,  ion channel models, mod files by:

- Finding the EModelWorkflow from the EModel (emodel.generation.activity.followedWorkflow)
- Finding the linked EModelScript to get the hoc file from workflow.generates.
- Store the distribuion of the EModelScript as an asset pointing to the EModel.
- Finding the linked EModelConfiguration (workflow.haspart)
- Finding the SubcellularModelScripts through `config.uses`.
- Save in the db the subcellular model script as IonChannelModel, if not added before (through another EModel).
- For each subcellular model script find the ions and store them in the Ion table if they don't exist.
- Create a record in the IonToIonChannelModel table for each ion in the list, linking it to the ion_channel_model, if there are any.
- For each IonChannelModel linked to a EModel create a record in IonChannelModelToEModel table.

After running the import scripts there are:
- 3 Ions (K, Na, Ca) dropped the ng from the label.
- 118 Emodels.
- Every EModel has 3 associated assets.
   - 1 hoc file.
   - 2 json files (one coming from the distribution of the emode, the other from that of the emodel script), not sure if the'yre the same.
   - for example:
```sql
 EM__MOBopl_mTC_bSTUT                                              | EMS__emodel=MOBopl_mTC_bSTUT__etype=bSTUT__mtype=MOBopl_mTC__species=mouse__brain_region=MOB.json
 EM__MOBopl_mTC_bSTUT                                              | MOBopl_mTC_bSTUT.hoc
 EM__MOBopl_mTC_bSTUT                                              | EM__emodel=MOBopl_mTC_bSTUT__etype=bSTUT__mtype=MOBopl_mTC__species=mouse__brain_region=MOB.json
 
 ```
 - There are 125 IonChannelModels.
 - Every Emodel has IonChannelModels (at least 3). For example:
 
```sql
 name                                |                                                                          ion_channel_model_names
-------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 EM__hipp_rat__CA1_int_cNAC_970717D_20190307182510__3              | ["kdrb", "na3", "kdb", "kmb", "kap", "hd", "can", "cal", "cat", "cagk", "kca", "cacum", "kad"]
 EM__1372346__cADpyr__13                                           | ["CaDynamics_DC0", "Ca_HVA2", "Ca_LVAst", "SKv3_1", "SK_E2", "K_Pst", "K_Tst", "NaTg", "Nap_Et2", "Ih"]
 GIN_etype point-neuron (Alias for EM__2f92aa0__cNAC__16)          | ["Ih", "NaTg", "CaDynamics_DC0", "Ca_HVA2", "Ca_LVAst", "SKv3_1", "SK_E2", "K_Pst", "K_Tst"]
 EM__hipp_rat__CA1_int_cNAC_970627BHP1_20190306162834__3           | ["kdrb", "na3", "kdb", "kmb", "kap", "hd", "can", "cal", "cat", "cagk", "kca", "cacum", "kad"]
 EM__hipp_rat__CA1_int_cNAC_971114B_20190329154607__2              | ["kdrb", "na3", "kdb", "kmb", "kap", "hd", "can", "cal", "cat", "cagk", "kca", "cacum", "kad"]
 EM__hipp_rat__CA1_pyr_cACpyr_mpg141217_A_idB_20190305112803__2    | ["kdr", "nax", "kmb", "kap", "hd", "can", "cal", "cat", "kca", "cagk", "cacum", "kad"]
 EM__hipp_mouse__CA1_pyr_mpg150304_A2_idA_080220242336__7          | ["kdr", "kmb", "kap", "nax", "kdbm", "na3", "hdpas", "can", "cal", "cat", "kca", "cagk", "cacum", "kad"]
 EM__hipp_rat__CA1_pyr_cACpyr_oh140807_A0_idB_20190305105211__3    | ["kdr", "nax", "kmb", "kap", "hd", "can", "cal", "cat", "kca", "cagk", "cacum", "kad"]
```
- All 125 IonChannelModels have an associated mod file.

Does this sound roughly correct @darshanmandge, @ilkilic, @AurelienJaquier ?

AurelienJaquier: Yes, this sounds correct.
The only thing that is (slightly) surprising to me, is that we have 133 SubCellularModelScripts in mmb-point-neuron-framework-model. One of them is the ProbAMPANMDA_EMS synapse. So out of the 132 remaining ion channel models and other mechanisms, you found 'only' 125 to be linked to e-models, meaning that we have 7 of them that are for some reason not used by any e-model? It would be interesting to find out which ones and why they were registered to nexus in the first place, and determine if we will need to move them to entity-core or not.

g-bar: These are the names of the SubCellularModelScripts, that are not linked to any EModel, not sure if some a duplicates?

```py
['ProbAMPANMDA_EMS', 'Kv7', 'kamt', 'kdrmt', 'ks', 'nax', 'kdb', 'kdrb']
```

darshanmandge: > ProbAMPANMDA_EMS is a synapse model. I am not sure to understand why it is there too.

We use this file for validation of cADPyr EModel for EPSP attenuation.

AurelienJaquier: I think we had duplication of nax for thalamus brain region, that might explain why it is in the list.

g-bar: BTW, should the names of the IonChannelModels be unique? 
Although there are 125 imported there are only 112 unique names.

g-bar: 'Kv7', 'kamt', 'kdrmt', 'ks', 'nax', 'kdb', 'kdrb'

All exit in the db excep for 'Kv7'.

darshanmandge: > BTW, should the names of the IonChannelModels be unique?
No. We have same name channel files for different brain region. That was one reason we wanted to add a string to the suffix in front of it. e.g. like TC_nax, TC_kdb. 
I discussed this with @AurelienJaquier today. Aurelien, do you have any suggestions? If we have the brain region somewhere, then each of the file should be stored even though if it is duplicate. Then we don't need to change the SUFFIX

jdcourcol: we don't need the name to be unique in the database, only the ids.
They can be differentied by the brain region where they are applicable. I don't recall if we have brain region btw.

g-bar: Yes there is  `brain_region_id` in the `IonChannelModel`.

g-bar:
> One question, for which I would need suggestions from engg. team:
> 
> For the download button on the EModel page of the platform, I would like to have
> 
>     * all assets of new emodel artifact: hoc, 2 jsons (EMS*.json and EM*.json)
> 
>     * **the emodel morphology included along with all the resources mentioned**
> 
>     * **all ion channel models of the emodel**
> 
> 
> Do you prefer the last 2 entities to be added to the emodel artifact schema here or it can combined later with some code for the download button?

The assets and ion channel models (and their assets [mod files]) will be already in the response.

For the morphology you have the metadata, but not the files, do we want the files as well? Can be added to the response.

Do we need this for the list view too or just for the detail view?

g-bar:
> I think we need to also add a criteria to save icm of each species and brain region.

ICM schema already has species_id and brain_region_id.

> Just another precaution that we should prepare for: In future we might have same name of icm for same species and same brain region but are different then exiting ones and are uploaded by different users

Currently there are no unique constraints other than the id, we can be as strict as we want on this, we could have for instance a unique constraint on (icm.name, brain_region.id, species._id, authorized_project_id).  For the moment I'll leave it as is.

g-bar:
> These too are present for both rat and mouse hippocampus probably:
> kdb Hippocampal formation
> kdrb Hippocampal formation

Indeed these exist for the rat and mouse, however the ones that are not linked to any EModel also have mouse
```

('kdb',
  {'@id': 'mba:1089', 'label': 'Hippocampal formation'},
  {'@id': 'NCBITaxon:10090', 'label': 'Mus musculus'}),
 ('kdrb',
  {'@id': 'mba:1089', 'label': 'Hippocampal formation'},
  {'@id': 'NCBITaxon:10090', 'label': 'Mus musculus'})]
```

I guess repeats and safe to ignore?

darshanmandge:
> For the moment, I'll leave it as is.

Ok.

> 'kdb',
>   {'@id': 'mba:1089', 'label': 'Hippocampal formation'},
>   {'@id': 'NCBITaxon:10090', 'label': 'Mus musculus'}),

>  ('kdrb',
>   {'@id': 'mba:1089', 'label': 'Hippocampal formation'},
>   {'@id': 'NCBITaxon:10090', 'label': 'Mus musculus'})]

If there are the IDs from nexus and they repeat for the same species, then we can keep one copy for that species (e.g. rat).

If the same resource and asset also exist in other species, then we should keep a copy of the other species (mouse) even though we have it for rat. But only one copy of the resource for that species is fine. No need to create multiple resources for each emodel of that species.

> Cerebellum icms

Not checked yet. Working on it

app/cli/utils.py
darshanmandge: >    'ProbAMPANMDA_EMS: 

We should keep this as root brain region as it is used in cADpyr emodel for Basic groups and Cell regions (root region)

The cerebellum icms below can be removed. I can confirm that they are not linked to any emodels:

>     Kv7  Cerebellum
>     kamt Cerebellum
>     kdrmt Cerebellum
>     ks Cerebellum
>     nax Cerebellum

These hippocampus models we discussed. We keep one copy for each species:

> kdb Hippocampal formation
>     kdrb Hippocampal formation

g-bar: Cool added the code to import ProbAMPANMDA_EMS, kdb,kdrb  were already there for both species so ignored.

app/schemas/ion_channel_model.py
```diff
@@ -0,0 +1,46 @@
...
+    model_config = ConfigDict(from_attributes=True)
+    name: str
+    description: str
+    species: SpeciesRead
+    strain: StrainRead | None
+    brain_region: BrainRegionRead
+    is_ljp_corrected: bool
+    is_temperature_dependent: bool
+    temperature_celsius: int
+
```
AurelienJaquier, line 43: I think we are missing the following fields from the [General_schema docx file](https://openbraininstitute.sharepoint.com/:w:/r/sites/OBI-Scientificstaff/_layouts/15/Doc.aspx?sourcedoc=%7BEEE86C08-A09D-469E-9849-CD85599E4026%7D&file=General_schema.docx&action=default&mobileredirect=true&wdOrigin=OUTLOOK-METAOS.FILEBROWSER.FILES-SITES-FOLDER):
- stochastic
- type

g-bar: None of the SubcellularModelScripts seem to have these field. Or do they have different names in the current data from nexus?

AurelienJaquier: Yes, because those fields were not on nexus, we are adding them now.
type can be 'distributed' for every IonChannelModel.
stochastic will be False for most of our models, and True for the few that are stochastic. I believe StochKv3 is one of the stochastic ones. I think @darshanmandge knows if we have other that are stochastic

darshanmandge: Yes, `StochKv3` is currently the only channel which has stochasticity and `stochastic` should be `true` for this channel. 
I don't think I have older versions of the Stochastic channel. You can check If any of the channel names begin with `Stoch`, e.g. StochKv2, then that should also be the property as `true`

jdcourcol: so I think one of you should modify the import script to fill this properties.

g-bar: Update: I updated the import scripts to read `icm_type` and `stochastic` directly from the 'SubcellularModlescript' resource insted of parsing the name. @darshanmandge to update the nexus resources to add those properties.

AurelienJaquier:
Just a quick comment about the ProbAMPANMDA_EMS resource: it is a synapse, and we intend on having another schema specifically for synapses. So even though ProbAMPANMDA_EMS is currently a SubCellularModelScript on nexus, it should not be imported as an IonChannelModel in entitycore.

app/schemas/ion_channel_model.py
AurelienJaquier: Hi @g-bar ! We have some updates about the schema. It turns out that each and every resource that we would ever have would be of icm_type 'distributed', meaning that this type is not very useful to discriminate the resources.
So we think it would be best to simply remove this field altogether. So @darshanmandge will only add the stochastic field to the nexus resources, and not the icm_type.
I apologize for the additional work I have made you do.

app/schemas/ion_channel_model.py
```diff
@@ -0,0 +1,49 @@
...
+    SpeciesRead,
+    StrainRead,
+)
+
+
+class NmodlParameters(BaseModel):
+    range: list[str]
+    read: list[str] | None = None
+    suffix: str | None = None
+    point_process: str | None = None
```
AurelienJaquier, line 21: Here you can also remove point_process because we won't be using it.

app/schemas/ion_channel_model.py
```diff
@@ -0,0 +1,49 @@
...
+    BrainRegionRead,
+    CreationMixin,
+    IdentifiableMixin,
+    SpeciesRead,
+    StrainRead,
+)
+
+
+class NmodlParameters(BaseModel):
+    range: list[str]
```
AurelienJaquier, line 18: Also a quick update here about NmodlParameters: we have added 'global' as a field of NmodlParameters in the schema docx file. We have not added the 'global' data to any nexus resource metadata yet because it would take some time. We intend on doing it in the next iteration.

app/schemas/ion_channel_model.py
darshanmandge: `stochastic` has been [added](https://github.com/openbraininstitute/prod-platform-architecture/issues/47) to both prod and staging nexus. @g-bar you can move forward with the PR merge after the changes request here and import ion channel resources to entirycore.

jdcourcol: The import script above is reading data from dec 2024. If you add directly to nexus, this is not going to be taken into account. I think best is to modify the import script to add this "stochastic" property.

app/schemas/ion_channel_model.py
g-bar: What type should it be, do i add it and set it as optional for the time being?

app/db/model.py
```diff
@@ -698,6 +706,57 @@ class ExperimentalSynapsesPerConnection(
...
+
+class IonChannelModel(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, Entity):
+    __tablename__ = EntityType.ion_channel_model.value
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+
+    is_ljp_corrected: Mapped[bool] = mapped_column(default=False)
+    is_temperature_dependent: Mapped[bool] = mapped_column(default=False)
+    temperature_celsius: Mapped[int]
+    stochastic: Mapped[bool] = mapped_column(default=False)
```
eleftherioszisis, line 734: I would suggest stochastic -> is_stochastic to be in agreement with the format of the booleans above. @AurelienJaquier wdyt?

AurelienJaquier: sounds good to me, let's change it to is_stochastic

g-bar: ok, changed

jdcourcol:
I am lost with the review, what is left for this to be merged ?

ilkilic (approved):

PR #125, opened 2025-04-11
Improve Subject

eleftherioszisis:
Add missing fields to subject that are needed for importing subject information in #7 and #119


eleftherioszisis:
Would it make more sense perhaps to make Age a JSONB instead of a separate table?

tests/utils.py
```diff
@@ -124,3 +124,107 @@ def create_brain_region_id(client, id_: int, name: str):
...
+            "name": "Public Entity",
+            "authorized_public": True,
+        },
+    )
+    public_morph = response.json()
+
+    inaccessible_obj = assert_request(
+        client_user_2.post,
+        url=route,
+        json=json_data | {"name": "inaccessable morphology 1"},
```
GianlucaFicarelli, line 182: Typo
```suggestion
        json=json_data | {"name": "inaccessible morphology 1"},
```

app/db/model.py
```diff
@@ -345,6 +359,27 @@ class Entity(LegacyMixin, Identifiable):
     }
 
 
+class Subject(NameDescriptionVectorMixin, SpeciesMixin, Entity):
```
GianlucaFicarelli, line 362: Just to be sure, now `subject` isn't global anymore, but it's an entity owned by a project_id, so anyone can upload its own subject?

app/db/model.py
```diff
@@ -101,9 +104,20 @@ class Strain(Identifiable):
...
+    min_value: Mapped[int | None]
+    max_value: Mapped[int | None]
+    unit: Mapped[AgeUnit]
+    period: Mapped[AgePeriod]
+
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+    __table_args__ = (
+        UniqueConstraint("value", "min_value", "max_value", "unit", "period", name="uq_age"),
+    )
```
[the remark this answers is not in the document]
eleftherioszisis: Not sure if we have thought well this admin-can-register-some-data-only approach. When registering experimental data does that mean that for each possible age in a dataset one should ask an admin to register them?

While I can get behind the idea of the global resources that only admins can register, it becomes quickly a burden when those resources include ranges, either finite or continuous, such as the age of the subject.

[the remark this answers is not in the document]
GianlucaFicarelli: This case would be solved if we don't store Age in a separate table.
For other global resources, we can create other roles besides admin, if we need to, or make them not global.

app/db/model.py
```diff
@@ -101,9 +104,20 @@ class Strain(Identifiable):
...
-    __tablename__ = "subject"
-    name: Mapped[str] = mapped_column(unique=True, index=True)
+class Age(Identifiable):
+    __tablename__ = "age"
+
+    value: Mapped[int | None]
+    min_value: Mapped[int | None]
+    max_value: Mapped[int | None]
+    unit: Mapped[AgeUnit]
+    period: Mapped[AgePeriod]
```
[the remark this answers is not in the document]
eleftherioszisis: This is data that is coming from the legacy datasets. There are:
* subjects with a min/max range of age
* subjects with a value for age and not min/max
* differentiation between prenatal/postnatal (you cannot encode that into the number unless we consider negatives :) )
* and the unit that although we can have a single one, it would be strange to specify human age in days.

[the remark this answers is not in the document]
GianlucaFicarelli: We could do some conversion when importing legacy datasets, if it makes sense.

> - subjects with a min/max range of age and no specific value

I understand that the min/max range is used for uncertainty. What if we store the mean value, and add a column with the uncertainty? In this way we can still sort and filter by value.

> - subjects with a value for age and not min/max

The uncertainty would be 0 in this case

> - differentiation between prenatal/postnatal (you cannot encode that into the number unless we consider negatives :) )

Using negatives to store prenatal would have some sense... because it would be possible to sort by value regardless of the period, and the value could be presented to the user in a nicer way in the GUI. Not a strong opinion, though.

> - and the unit that although we can have a single one, it would be strange to specify human age in weeks

I agree, but we can store the ages internally in the same unit (interval? days?), and just change how they are presented to the user.

app/db/model.py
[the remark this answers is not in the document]
eleftherioszisis: I removed the Age table and added the age fields in the subject table.

app/db/model.py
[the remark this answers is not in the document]
eleftherioszisis: > I understand that the min/max range is used for uncertainty. What if we store the mean value, and add a column with the uncertainty? In this way we can still sort and filter by value.

I wouldn't want to make any assumptions as to how these ranges were calculated, therefore I left the age_value, age_min, age_max as they were, converting them into intervals.

I used `timedelta` intervals that map to postgres INTERVAL type. Note that on the Pydantic side I opted for a float in seconds serialization, because it will be easier for the frontend to work for it instead of the iso8601 duration serialization.

app/db/model.py
```diff
@@ -345,6 +359,27 @@ class Entity(LegacyMixin, Identifiable):
...
 
 
+class Subject(NameDescriptionVectorMixin, SpeciesMixin, Entity):
+    __tablename__ = EntityType.subject.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    age_id: Mapped[Age | None] = mapped_column(ForeignKey("age.id"))
+    age = relationship(Age, uselist=False, foreign_keys=[age_id])
+
+    sex: Mapped[Sex | None]
+    weight: Mapped[float | None]
```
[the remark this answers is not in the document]
eleftherioszisis: Added an annotation that explains it is in grams.

tests/utils.py
eleftherioszisis: Fixed.

app/db/model.py
eleftherioszisis: Yes, it is an entity indeed. Let's leave it like this for now and we can revisit it in the future, when we have most of the data imported.

app/schemas/base.py
```diff
@@ -128,3 +123,23 @@ class MeasurementCreate(BaseModel):
...
+class SubjectBase(BaseModel):
+    model_config = ConfigDict(from_attributes=True, ser_json_timedelta="float")
+    name: str
+    description: str
+    sex: Sex | None = None
+    weight: Annotated[float | None, Field(title="Weight", description="Weight in grams")] = None
+    age_value: timedelta | None = None
+    age_min: timedelta | None = None
+    age_max: timedelta | None = None
+    age_period: AgePeriod | None = None
```
GianlucaFicarelli, line 137: For confirmation, is it acceptable to not specify the optional fields sex, weight, age during the creation?

I think we should do more validation during the creation of the subject, for example:
- if age_min is specified, also age_max should be (and vice-versa)
- If they are specified, age_min < age_max
- if any age is specified, also age_period must be specified
- if age_value is specified, neither age_min nor age_max should be specified (unless we consider age_value as a possible intermediate value, but it doesn't seem the case)
- all the ages should be > 0
- weight should be > 0

eleftherioszisis: Added field and model validators to subject schema to account for all constraints mentioned.

app/schemas/base.py
```diff
@@ -129,13 +129,59 @@ class SubjectBase(BaseModel):
...
+            msg = "age_value and age_min/age_max cannot both be provided"
+            raise ValueError(msg)
+        return self
+
+    @model_validator(mode="after")
+    def min_max_age_range_consistency(self):
+        """Age min and max must be provided together or not at all."""
+        if self.age_min and self.age_max:
+            if self.age_min >= self.age_max:
+                msg = "age_min cannot be greater than age_max"
```
GianlucaFicarelli, line 175: Minor, to be more precise "age_min cannot be greater or equal to than age_max", or "age_max must be greater than age_min"

eleftherioszisis: Sure, done.

GianlucaFicarelli (approved):

PR #128, opened 2025-04-17
Filter by id morphology

bilalesi:


app/service/morphology.py
```diff
@@ -117,16 +117,22 @@ def read_many(
...
-        q.join(Species, ReconstructionMorphology.species_id == Species.id)
-        .outerjoin(Strain, ReconstructionMorphology.strain_id == Strain.id)
-        .outerjoin(Contribution, ReconstructionMorphology.id == Contribution.entity_id)
-        .outerjoin(Agent, Contribution.agent_id == Agent.id)
-        .outerjoin(
-            MTypeClassification, ReconstructionMorphology.id == MTypeClassification.entity_id
+    def apply_filter_query_operations(q):
+        """Apply filtering by IDs if provided, then apply necessary joins."""
+        if morphology_filter.id__in:
+            q = q.filter(ReconstructionMorphology.id.in_(morphology_filter.id__in))
```
eleftherioszisis, line 123: Could you check if this behavior for `id__in` is already covered by:
https://github.com/openbraininstitute/entitycore/blob/c27be9fbf73b8c7160779300066493b1e61933df/app/filters/base.py#L56-L59
?

bilalesi: yep that's right, i return it back the the lambda fn filter
thanks @eleftherioszisis

eleftherioszisis (approved):

PR #129, opened 2025-04-24
Curation of nwb files

eleftherioszisis:


app/cli/curation/nwb.py
```diff
@@ -0,0 +1,251 @@
...
+            full_new_path = f"{base_path}{new_path}"
+
+            # update the paths
+            if full_old_path not in f:
+                continue
+
+            f.move(full_old_path, full_new_path)
+
+
+def _simulation_update(f, base_path, ecode_mapping):
```
AurelienJaquier, line 118: You can delete this function. It was renamed into sscx_simulation_update, so we don't need the old one anymore.

app/cli/curation/nwb.py
```diff
@@ -0,0 +1,251 @@
...
+            f"/data_organization/{cellid}/{new_ecode}",
+        )
+    else:
+        for rep in f[f"/data_organization/{cellid}/{old_ecode}"]:
+            old_rep = f"/data_organization/{cellid}/{old_ecode}/{rep}"
+            new_rep = f"/data_organization/{cellid}/{new_ecode}/{rep}"
+            if new_rep not in f:
+                f.move(old_rep, new_rep)
+            else:
+                for sweep in f[f"/data_organization/{cellid}/{old_ecode}/{rep}"]:
```
AurelienJaquier, line 187: I guess you could use the old_rep variable here

app/cli/curation/nwb.py
```diff
@@ -0,0 +1,251 @@
...
+        hold_hyp -> Spontaneous
+    Also, we add ic / ics as a prefix and add a 'unique' index as a suffix
+    """
+    stim_num = 0
+    for old_path in f[base_path]:
+        if old_path in ecode_mapping:
+            new_ecode = ecode_mapping[old_path]
+            new_path = f"{new_ecode}__{stim_num}"
+            units = _get_units(f, base_path, old_path)
+            if "acquisition" in base_path and units == "volts":
```
AurelienJaquier, line 73: I think you can use the _get_prefix function here

AurelienJaquier (commented):
About the stimuli that have been added when checking the thalamus simulation nwb file: I am 90% sure that all of the thalamus simulation nwb files use the same stimuli, but I would like to check in order to become 100% sure. If I am wrong and we discover other stimuli, then we would have to add them to the list.

@eleftherioszisis could you tell me how I could get all of those files so that I could check their stimulus names?

app/cli/curation/nwb.py
eleftherioszisis: Done.

app/cli/curation/nwb.py
eleftherioszisis: Indeed.

eleftherioszisis:
> About the stimuli that have been added when checking the thalamus simulation nwb file: I am 90% sure that all of the thalamus simulation nwb files use the same stimuli, but I would like to check in order to become 100% sure. If I am wrong and we discover other stimuli, then we would have to add them to the list.
> 
> @eleftherioszisis could you tell me how I could get all of those files so that I could check their stimulus names?

I'll share them with you.

app/cli/curation/nwb.py
eleftherioszisis: Done.

AurelienJaquier (approved):
I checked the files @eleftherioszisis shared with me, and added the new protocol names that I found. This PR is good to be merged from my point of view.

eleftherioszisis:
I revisited the curation because it was only applied per file previously. However, there are cases that multiple assets within a single entity need to be parsed/curated/merged.

app/cli/curation/electrical_cell_recording.py
```diff
@@ -13,21 +16,98 @@
...
+        assets: Dictionary with content_type as keys and assets as values.
+        out_dir: Output directory for writing curated files.
+        is_dry_run: If True curation will not modify database assets.
+
+    Returns:
+        Dictionary with digests as keys and curated file paths as values.
+    """
+    # Multiple assets: keep the nwb one if any.
+    if len(assets) > 1:
+        if nwb_asset := assets.get(ContentType.nwb, None):
```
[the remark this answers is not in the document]
eleftherioszisis: Because I forgot to clean it up. :D

eleftherioszisis:
@AurelienJaquier are you good with the new version of the curation? Any objections?

AurelienJaquier:
@eleftherioszisis looks good to me! I have no objections

PR #131, opened 2025-04-24
Update ion channel model

ilkilic:


app/cli/utils.py
```diff
@@ -419,41 +419,52 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
 
-    assert nmodl_parameters_validated  # noqa: S101
+    assert neuron_block_validated  # noqa: S101
 
     db_ion_channel_model = IonChannelModel(
         legacy_id=[legacy_id],
         legacy_self=[legacy_self],
         name=script["name"],
-        description=script.get("descripton", ""),
+        description=script.get("description", ""),
```
g-bar, line 463: 👍

app/service/ion_channel_model.py
```diff
@@ -0,0 +1,73 @@
...
+from app.db.model import IonChannelModel
+from app.dependencies.auth import UserContextDep, UserContextWithProjectIdDep
+from app.dependencies.common import PaginationQuery
+from app.dependencies.db import SessionDep
+from app.errors import ensure_result
+from app.schemas.ion_channel_model import IonChannelModelCreate, IonChannelModelRead
+from app.schemas.types import ListResponse, PaginationResponse
+
+
+def read_many(
```
g-bar, line 15: I think we can use these functions to avoid rewriting the logic:
https://github.com/openbraininstitute/entitycore/blob/main/app/queries/common.py#L68

We would need to define filters and add tests too.

I can create a new PR (or update this one) which such changes if you wish.

app/cli/utils.py
```diff
@@ -419,41 +419,52 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
+        ion_entries = script.get("ion")
+        if isinstance(ion_entries, dict):
+            ion_entries = [ion_entries]
+
+        useion_structured = [
+            {
+                "ion": {
+                    "ontology_id": ion_data.get("@id"),
+                    "name": ion_data.get("label"),
+                },
```
g-bar, line 435: Now we store the ions twice one here and one as  references to the Ion table through this table: 
https://github.com/openbraininstitute/entitycore/blob/7ae9373abce828d9c0abca0e7c3e23c214f1e887/app/db/model.py#L715

So we should keep just one of them. 

Jdc had suggested to have a Ion table https://github.com/openbraininstitute/prod-platform-architecture/issues/39#issuecomment-2776604101

but this is incompatible with that.

One way to resolve this could be to instead of a json field create a table NeuronBlock, which contains the relationships with Ion (as does IonChannelModel now) and adds the other properties and have a one-to-one relationship betweoin NeuronBlock and IonChannelModel.

g-bar (changes requested):

app/cli/utils.py
```diff
@@ -419,41 +419,52 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
         legacy_id=[legacy_id],
         legacy_self=[legacy_self],
         name=script["name"],
-        description=script.get("descripton", ""),
+        description=script.get("description", ""),
         is_ljp_corrected=script.get("isLjpCorrected", False),
         is_temperature_dependent=script.get("isTemperatureDependent", False),
         temperature_celsius=int(temperature_value),
-        nmodl_parameters=nmodl_parameters_validated.model_dump(),
+        neuron_block=neuron_block_validated,
```
g-bar, line 467: I'm not sure if Sqlalchemy can receive a PydanticModel, this might fail.

app/cli/utils.py
ilkilic: I agree we should avoid storing ions twice. My current thinking is to keep Ion as a proper table (as Jdc suggested) and reference it through useion inside NeuronBlock, without duplicating it at the root level of IonChannelModel.

But I am not sure if creating a NeuronBlock table might introduce unnecessary complexity since it is only used once per model and is not shared. 

If there is no other clean alternative, maybe we could just use the ion name (as a string) in useion and not a full reference and keep the ions field at the root level as it was before, what do you think? pinging @AurelienJaquier @darshanmandge as well

class UseIon(BaseModel):
    **ion_name**: ~~Ion~~ str
    read: list[str] | None = None
    write: list[str] | None = None
    valence: int | None = None
    main_ion: bool | None = None

app/service/ion_channel_model.py
ilkilic: I was trying to add some tests but wasn’t exactly sure how but feel free to go ahead and update this PR

app/cli/utils.py
g-bar: There are three options:

- The names are unique so having `ion_name` would work but we would lose strict referencial integrity, i.e. we can ensure the name exists in the ion table at the application layer but not at database layer.

- Another way is to keep them in the Ionchannel model as they are now (the main_ion property can be added to the relationship table), move them off the NeuronBlock, the consumer can just read it from there, not sure is semantically correct though.

- Extra table, more complexity, yes, but still manageable with sqlalchemy and is the cleanest solution while maintaining strict constraints.

ilkilic: Thanks for the breakdown!

What about a mix of Option 1 and 2: use `ion_name` in useion and keep `ions` at the root level. Yes, there is some duplication, both useion and ions reference the same ions but useion specifically captures NEURON-related syntax. I believe it is reasonable to keep it in both places, as it serves distinct purposes. And since NeuronBlock is one-to-one with the model and usually only includes one useion, I think this small duplication is acceptable, it keeps registration simple and still preserves integrity. What do you think?

app/schemas/ion_channel_model.py
```diff
@@ -13,20 +11,26 @@
 )
 
 
-class NmodlParameters(BaseModel):
-    range: list[str]
+class Ion(CreationMixin, IdentifiableMixin, AuthorizationMixin, BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    ontology_id: str | None = None
+    name: str
```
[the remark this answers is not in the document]
ilkilic: sorry, I wasn’t sure whether it should be a separate class, I have reverted it

app/cli/utils.py
g-bar: I think it's preferable to go with Option 1, at least for now.

The mixed approach doesn't solve the referential integrity issue, just moves the goalpost, it doesn't ensure the names in the `use_ion`, `ion_name` fields are valid Ion names and or that they're in sync with the `ions` field.
Furthermore this might be a source of confusion for users having ions in two places, better to keep it simple.

Strict integrity at db level is nice but ultimately is useful in the case of a db admin with write permissions manually changing the fields, for most cases on the application layer should be ok. If needed we can add a trigger that enforces this at db level but for now it should be ok to enforce it just at the application level.

I made the changes to do that.

Furthermore I made a slight change for `list` fields to be just `list` and not `list | None`, the empty case being represented with an empty list, otherwise we could have two null cases `None` and `[]`.

app/cli/utils.py
```diff
@@ -416,44 +415,49 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
+
+    if neuron_block_raw:
+        ion_entries = ensurelist(script.get("ion", []))
+        useion_structured = [
+            {
+                "ion_name": get_or_create_ion(ion, db).name,
+                "read": ensurelist(neuron_block_raw.get("read", [])),
+                "write": ensurelist(neuron_block_raw.get("write", [])),
+                "valence": neuron_block_raw.get("valence"),
+                "main_ion": None,
```
g-bar, line 431: This is a bit odd to me, do we need two null cases `None` and `False` ? can `main_ion` be just `bool` and default to false?

app/cli/utils.py
ilkilic: Thanks for the clarification, that makes sense. I agree it is cleaner to go with Option 1 for now, keeping everything simpler and handling validation at the application level sounds like a good balance. Also, good point about using empty lists instead of None to avoid double cases

app/cli/utils.py
ilkilic: Allowing None would give more flexibility when registering, since users don’t have to specify it. Defaulting to False would suggest the ion is not the main ion, even though it could be

g-bar: So None has the meaning of "unset" or "unknown"?

I guess that works.

ilkilic: Yes exactly. Great!

g-bar:
> I see that we don't have the ontology_id anymore, shouldn't we add it to useion class? (It would be optional, it is a url pointing to an external ontology) `ontology_id: str | None`

I think it was decided to drop the ontology_id, or is that none the same thing?
https://github.com/openbraininstitute/prod-platform-architecture/issues/39#issuecomment-2786023490

> Also, I realized I forgot to add the acronym field to IonChannelModel: `acronym: str | None = None`

> Are you working on the PR or would it be ok if I make these changes?
I am still to work on the routers, tests, but feel free to make changes.

app/service/ion_channel_model.py
g-bar: Added routes and tests.

app/errors.py
```diff
@@ -106,6 +107,24 @@ def ensure_uniqueness(
         raise
 
 
+@contextmanager
+def ensure_valid_fks(
```
eleftherioszisis, line 111: Minor: I would prefer a full name instead of fks.

app/service/ion_channel_model.py
```diff
@@ -0,0 +1,106 @@
...
+
+    stmt = (
+        sa.select(sa.func.count() == len(ion_names)).select_from(Ion).where(Ion.name.in_(ion_names))
+    )
+
+    all_names_exist = db.execute(stmt).scalar_one()
+
+    if not all_names_exist:
+        msg = "Ion name does not exist"
+        raise ApiError(message=msg, error_code=ApiErrorCode.ION_NAME_NOT_FOUND)
```
eleftherioszisis, line 97: It seems that this can be added into the IonChannelCreate schema as a [model validator](https://docs.pydantic.dev/latest/concepts/validators/#model-validators).

app/db/model.py
```diff
@@ -732,16 +722,8 @@ class IonChannelModel(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
-    nmodl_parameters: Mapped[JSON_DICT]
-
-    ions: Mapped[list[Ion]] = relationship(
-        primaryjoin="IonChannelModel.id == IonToIonChannelModel.ion_channel_model_id",
-        secondary="ion__ion_channel_model",
-        uselist=True,
-        viewonly=True,
-        order_by="Ion.name",
-    )
+    acronym: Mapped[str] = mapped_column(default="")
```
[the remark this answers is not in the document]
AurelienJaquier: There has been a misunderstanding in the General_schema.docx. This should not be there (I believe @ilkilic will remove it).
This will be (in a future iteration) in the brain region, and will indicate the acronym used by Allen Institute to design this brain region. It is usually shorter than the actual brain region name we are using in our atlas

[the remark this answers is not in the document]
eleftherioszisis: Yeah, there brain region acronym is already there:
https://github.com/openbraininstitute/entitycore/blob/9a37dc8b82395601620f98cc4b3a7e16c2746692/app/db/model.py#L89

app/service/ion_channel_model.py
g-bar: Tricky to do it in a model_validator coz I need to pass the db, so I defined a dependency in the same module.

app/db/model.py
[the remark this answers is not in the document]
ilkilic: yes, sorry about the confusion, I have removed it in my last commit

app/schemas/ion_channel_model.py
```diff
@@ -34,11 +34,11 @@ class IonChannelModelBase(BaseModel):
     model_config = ConfigDict(from_attributes=True)
     description: str
     name: str
+    nmodl_suffix: str
     is_ljp_corrected: bool = False
     is_temperature_dependent: bool = False
     temperature_celsius: int
     is_stochastic: bool = False
-    acronym: str = ""
```
g-bar: Need to remove acronym from the db model (not just the schema) and add nmodel_suffix to the db model too. And run the migrations.

app/cli/utils.py
```diff
@@ -418,44 +419,50 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
 
-    assert nmodl_parameters_validated
+    assert neuron_block_validated
 
     db_ion_channel_model = IonChannelModel(
         legacy_id=[legacy_id],
         legacy_self=[legacy_self],
         name=script["name"],
-        description=script.get("descripton", ""),
+        nmodel_suffix=script.get("suffix"),
```
g-bar, line 458: Is it `nmodel_suffix` or `nmodl_suffix`?

ilkilic: good catch! it was a typo, it is `nmodl_suffix`. I corrected it in my last commit

app/cli/utils.py
```diff
@@ -418,44 +419,50 @@ def import_ion_channel_model(script: dict[str, Any], project_context: ProjectCon
...
 
-    assert nmodl_parameters_validated
+    assert neuron_block_validated
 
     db_ion_channel_model = IonChannelModel(
         legacy_id=[legacy_id],
         legacy_self=[legacy_self],
         name=script["name"],
-        description=script.get("descripton", ""),
+        nmodl_suffix=script.get("suffix"),
```
g-bar, line 458: Now the import is failing because the db model field is not nullable, but apparently this doesn't exist in all the ion channel models.

Should it be nullable or should there be a default value?

ilkilic: I think it wasn’t always consistent in Nexus, sometimes suffix was inside `nmodlParameters`. I just checked all the resources on Nexus and it should be present in at least one of those places.  I updated the script to fall back to that if needed (and use an empty string if both are missing, though that shouldn’t happen).  Let me know if it works now

g-bar: Thanks I check the db after import and there are no ICMs with an empty string as nmodl_suffix, so it works :).

app/schemas/ion_channel_model.py
```diff
@@ -1,46 +1,64 @@
...
+    nmodl_suffix: str
+    is_ljp_corrected: bool = False
+    is_temperature_dependent: bool = False
+    temperature_celsius: int
+    is_stochastic: bool = False
+    neuron_block: NeuronBlock
+
+
+class IonChannelModelCreate(IonChannelModelBase, AuthorizationOptionalPublicMixin):
+    species_id: UUID
```
jdcourcol, line 46: this is not in a subject ? I thought we moved these properties into a subject ?

ilkilic: Yes, it is already in the Subject, so I think I can make that change already. But Strain should have been in Subject as well, should we add it now before merging or wait? maybe we can update it later since Scientific artifacts scheme will bring more changes anyway

ilkilic: The ReconstructionMorphology schema also doesn’t use Subject at the moment. Would it be ok with you if we update all of them together once the Scientific Artifacts schema is introduced?

app/schemas/ion_channel_model.py
```diff
@@ -24,8 +24,8 @@ class UseIon(BaseModel):
 
 
 class NeuronBlock(BaseModel):
-    global_: list[str] = Field(default=[], alias="global")
-    range: list[str] = []
+    global_: list[dict[str, str]] = Field(default=[], alias="global")
+    range: list[dict[str, str | None]] = []
```
g-bar, line 28: This update broke the import script again because range is expected to be a list[dict] and it's being passed a list[str],
here is the validation error:

```sh
pydantic_core._pydantic_core.ValidationError: 6 validation errors for NeuronBlock
range.0
  Input should be a valid dictionary [type=dict_type, input_value='gksbar', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
range.1
  Input should be a valid dictionary [type=dict_type, input_value='gks', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
range.2
  Input should be a valid dictionary [type=dict_type, input_value='pinf', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
range.3
  Input should be a valid dictionary [type=dict_type, input_value='qinf', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
range.4
  Input should be a valid dictionary [type=dict_type, input_value='taup', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
range.5
  Input should be a valid dictionary [type=dict_type, input_value='tauq', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/dict_type
make: *** [import] Error 1
```

I recommend you run the import scripts to make sure it succeds since that isn't in the CI.
`make import`

you will  need the data in your machine, instructions here: 
https://github.com/openbraininstitute/entitycore?tab=readme-ov-file#optional-populate-the-database-with-sample-data

ilkilic: I have updated the import function and confirmed that `make import` runs successfully. I also updated the Ion table to allow ontology_id to be null since custom ions may not have an ontology id

g-bar (approved):
LGTM

ilkilic:
Thanks @g-bar , all set on my side, shall we go ahead and merge it?

ilkilic:
Once this is merged, could someone also review the entitysdk [PR](https://github.com/openbraininstitute/entitysdk/pull/42)? Thank you

PR #132, opened 2025-04-25
Fix missing schema fields in experimental densities

eleftherioszisis:


bilalesi (approved):
Thanks @eleftherioszisis, it worked

PR #143, opened 2025-04-30
the ScientificArtifact.py file was added, but not mixed in

dkeller9:
This branch is just intended for evaluation by the scientific team.

### DIFF
--- a/app/schemas/scientific_artifact.py
+++ b/app/schemas/scientific_artifact.py
@@ -1,0 +1,38 @@
+import uuid
+from datetime import datetime, timedelta
+from typing import Annotated, TYPE_CHECKING
+from pydantic import UUID4, BaseModel, ConfigDict
+
+from datetime import date
+
+#if TYPE_CHECKING:
+from app.schemas.agent import PersonRead
+from app.schemas.contribution import ContributionReadWithoutEntity
+ 
+# where was the artifact published?
+class IsDefinedByType(BaseModel):
+    """
+    Represents ways to define or identify an entity, such as by DOI, PMID, UUID, or other means.
+    All fields are optional.
+    """
+    DOI: str | None = None  # Optional DOI (Digital Object Identifier) as a string
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    Other: str | None = None # Optional alternative identifier as a string
+
+class ScientificArtifactMixin(BaseModel):
+    name :str
+    description:str  
+    subject_id : uuid.UUID
+    
+    brain_region_id: int
+    contributions: list["ContributionReadWithoutEntity"] | None
+    license_id: uuid.UUID 
+    experiment_date: date 
+
+    IsDefinedBy : IsDefinedByType
+    
+    validation_tags: dict[str, bool] #This is a dict{“properties_check”: T/F} (determined by a script not a user input. Should this be here or an annotation?) 
+
+    contact_id : uuid.UUID
+    contact_email: str


james-isbister:
Several comments and questions:

IsDefinedByType, IsDefinedBy: 
- These terms aren't clear to me. Would something like "original_source_location" be more explicit?
- What would the value be for an scientific artifact generated by the platform? It is defined by the entity itself, no? An optional original_source_location parameter might be suitable?

subject_id:
- What would the subject be for an analysis of data across multiple species?
- What would the subject be for a new brain region model combining data from multiple "subjects"? 
- Maybe this would be better as an optional parameter?

contributions:
- What is the idea here?

license_id: 
- So the user has to select a license every time they create an artifact? I think this would slow users down, and is only needed once the entity is public

experiment_date
- Is there not already some storage if the creation date for entities made on the platform?

validation_tags:
- Indeed this might be a some type of entity itself. We might see how best to format validations when we implement the morphology validation in obi-one?

contact_email:
- Would this be part of the contact that is linked by the contact_id?

chr-pok:
Here are a few more comments/questions from my side:
- `brain_region_id`: What is that ID based on, some predefined list of regions? Couldn't it be a UUID as well referencing a unique region entity?
- Defined in [General_schema.docx](https://openbraininstitute.sharepoint.com/:w:/s/OBI-Scientificstaff/EQhs6O6doJ5GmEnNhVmeQCYBPLrBgs0njgUDSt0xVGam4w?e=4Bhngm) but seems to be missing here: `[optional] additional_brain_regions` in case more than one region is involved.
- I agree that`license_id` is only required when made public. Perhaps it could be optional.
- `experiment_date` should be optional as well. Not all scientific artifacts contain experimental data.
- I agree that `contact_email` should be better part of the entity that is referenced by `contact_id`. Also, it should be optional.
- `validation_tags`: Is it even possible to put a dict into the database? From earlier discussions I thought this would be handled by annotations which would contain more information than just a boolean value, like validation date, version of the asset being validated, version of the code used to run the validation, etc.

dkeller9:
`IsDefinedByType, IsDefinedBy:

These terms aren't clear to me. Would something like "original_source_location" be more explicit?
What would the value be for an scientific artifact generated by the platform? It is defined by the entity itself, no? An optional original_source_location parameter might be suitable?`

I changed it to PublishedInType and added original_source_location parameter.

`subject_id:

What would the subject be for an analysis of data across multiple species?
What would the subject be for a new brain region model combining data from multiple "subjects"?
Maybe this would be better as an optional parameter?`

Now made optional.


`contributions:
What is the idea here?`

This is already in entity so removed.


`license_id:

So the user has to select a license every time they create an artifact? I think this would slow users down, and is only needed once the entity is public`

made optional

`experiment_date

Is there not already some storage if the creation date for entities made on the platform?`

Only creation and update date, not the actual experiment date.

`validation_tags:

Indeed this might be a some type of entity itself. We might see how best to format validations when we implement the morphology validation in obi-one?`

Sure, could be an entity. We had discussed adding tags in the science meeting.

`contact_email:

Would this be part of the contact that is linked by the contact_id?`

It is actually not in the contact class but maybe I will put it in there and remove it here.

`As there are some differences between entities created on the platform vs entities created externally, it might be worth considering having two types of mixin which inherit several shared properties?`

What other features would be needed that would justify the extra complexity?

dkeller9:
> brain_region_id: What is that ID based on, some predefined list of regions? Couldn't it be a UUID as well referencing a unique region entity?

It's the id of the Brain Region entity.

> Defined in [General_schema.docx](https://openbraininstitute.sharepoint.com/:w:/s/OBI-Scientificstaff/EQhs6O6doJ5GmEnNhVmeQCYBPLrBgs0njgUDSt0xVGam4w?e=4Bhngm) but seems to be missing here: [optional] additional_brain_regions in case more than one region is involved.


Thanks, now added.

> I agree thatlicense_id is only required when made public. Perhaps it could be optional.
> experiment_date should be optional as well. Not all scientific artifacts contain experimental data.
> I agree that contact_email should be better part of the entity that is referenced by contact_id. Also, it should be optional.

These fields are now optional.

> validation_tags: Is it even possible to put a dict into the database? From earlier discussions I thought this would be handled by annotations which would contain more information than just a boolean value, like validation date, version of the asset being validated, version of the code used to run the validation, etc.

Further discussion is still needed.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    DOI: str | None = None  # Optional DOI (Digital Object Identifier) as a string
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    original_source_location : str
+    other: str | None = None # Optional alternative identifier as a string
+
+class ScientificArtifactMixin(BaseModel):
+    name :str
+    description:str  
+    subject_id : uuid.UUID = None
```
chr-pok, line 27: If I am not mistaken, the correct notation should probably be:
```suggestion
    subject_id : uuid.UUID | None = None
```

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    
+    brain_region_id: int
```
chr-pok, line 29: Just wondering why this ID is defined as `int`, whereas other IDs as `uuid.UUID`? Same applied to `additional_brain_regions`.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    additional_brain_regions: list[int] | None
```
chr-pok, line 30: I would set it to None by default:
```suggestion
    additional_brain_regions: list[int] | None = None
```

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    license_id: uuid.UUID = None #only needed when public
```
chr-pok, line 31: Same as above:
```suggestion
    license_id: uuid.UUID | None = None #only needed when public 
```

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+
+class ScientificArtifactMixin(BaseModel):
+    name :str
+    description:str  
+    subject_id : uuid.UUID = None
+    
+    brain_region_id: int
+    additional_brain_regions: list[int] | None
+    license_id: uuid.UUID = None #only needed when public 
+    experiment_date: date = None
```
chr-pok, line 32: Same as above:
```suggestion
    experiment_date: date | None = None 
```

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    brain_region_id: int
+    additional_brain_regions: list[int] | None
+    license_id: uuid.UUID = None #only needed when public 
+    experiment_date: date = None 
+
+    PublishedIn : PublishedInType
+    
+    validation_tags: dict[str, bool] #This is a dict{“properties_check”: T/F} (determined by a script not a user input. Should this be here or an annotation?) 
+
+    contact_id : uuid.UUID
```
chr-pok, line 38: I would say this should be optional as well:
```suggestion
    contact_id : uuid.UUID | None = None
```

chr-pok (commented):
Thanks for the changes so far! I have just added a few more comments/suggestions directly into the code.

jdcourcol (commented):
There is already a location mixin with brain_region_id.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    Represents ways to define or identify an entity, such as by DOI, PMID, UUID, or other means.
+    All fields are optional.
+    """
+    DOI: str | None = None  # Optional DOI (Digital Object Identifier) as a string
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    original_source_location : str
+    other: str | None = None # Optional alternative identifier as a string
+
+class ScientificArtifactMixin(BaseModel):
```
jdcourcol, line 24: why is this not derived from Entity ?

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,38 @@
...
+    brain_region_id: int
+    additional_brain_regions: list[int] | None = None
+    license_id: uuid.UUID | None = None #only needed when public 
+    experiment_date: date | None = None 
+
+    PublishedIn : PublishedInType
+    
+    validation_tags: dict[str, bool] #This is a dict{“properties_check”: T/F} (determined by a script not a user input. Should this be here or an annotation?) 
+
+    contact_id : uuid.UUID | None = None
```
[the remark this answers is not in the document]
dkeller9: This is the contact person if there are questions about the artifact.

app/schemas/scientific_artifact.py
dkeller9: Not all entities will need things particular to scientific artifact, for example experiment date,

dkeller9:
> There is already a location mixin with brain_region_id.

Okay, brain region is removed and will come in through the mixin, although additional brain regions are retained. We might need to rethink how this is done. Also needed is a way to indicate which hierarchy classification schema the brain regions come from.

chr-pok:
> > There is already a location mixin with brain_region_id.
> 
> Okay, brain region is removed and will come in through the mixin, although additional brain regions are retained. We might need to rethink how this is done. Also needed is a way to indicate which hierarchy classification schema the brain regions come from.

Actually, I think in this case it would be nice to have additional brain regions directly added to the `BrainRegionFilterMixin` so it could be consistently used in other places as well. However, we should clearly specify what the purpose of `additional_brain_regions` actually is. Should it contain additional brain regions on the same level of granularity (e.g., brain region = S1FL, additional brain regions = [S1HL]), or a finer resolution of the (main) brain region (e.g., brain region = S1, additional brain regions = [S1FL, S1HL])?

app/schemas/scientific_artifact.py
dkeller9: I speculate that it does not refer to the entity ID but rather the Id in the hierarchical atlas.

dkeller9:
> > > There is already a location mixin with brain_region_id.
> > 
> > 
> > Okay, brain region is removed and will come in through the mixin, although additional brain regions are retained. We might need to rethink how this is done. Also needed is a way to indicate which hierarchy classification schema the brain regions come from.
> 
> Actually, I think in this case it would be nice to have additional brain regions directly added to the `BrainRegionFilterMixin` so it could be consistently used in other places as well. However, we should clearly specify what the purpose of `additional_brain_regions` actually is. Should it contain additional brain regions on the same level of granularity (e.g., brain region = S1FL, additional brain regions = [S1HL]), or a finer resolution of the (main) brain region (e.g., brain region = S1, additional brain regions = [S1FL, S1HL])?

Okay, I have added  the additional brain regions directly to the `BrainRegionFilterMixin`. I see them as completely different brain regions rather than children of the main brain region. It is true though that the  BrainRegion already has a 'children' field already though. Perhaps this can be repurposed to be additonal_brain_regions.

app/schemas/scientific_artifact.py
jdcourcol: I agree. The question still stands though , why a ScientificArtifact does not derive from Entity ?

app/schemas/scientific_artifact.py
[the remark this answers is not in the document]
jdcourcol: I am confused, I don't see changes in the database model ?

app/db/model.py
```diff
@@ -143,47 +160,183 @@ class SpeciesMixin(Base):
...
+    )
+    __mapper_args__ = {
+        "polymorphic_identity": __tablename__,
+        "polymorphic_on": "type",
+    }
+
+
+
+class ScientificArtifact(
+    Entity,
```
jdcourcol, line 274: nice

app/db/model.py
```diff
@@ -143,47 +160,183 @@ class SpeciesMixin(Base):
...
+    SubjectMixin,
+    NameDescriptionVectorMixin,
+    SpeciesMixin,
+    LocationMixin,
+    LicensedMixin
+):
+    __tablename__ = "scientific_artifact"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    experiment_date: Mapped[datetime] = mapped_column(DateTime, nullable=True)
```
jdcourcol, line 283: "experiment_date" : does that work well for a model ?

app/db/model.py
```diff
@@ -143,47 +160,183 @@ class SpeciesMixin(Base):
...
+    published_in: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
+    validation_tags: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
```
jdcourcol, line 285: I don't feel comfortable with a unstructured dict for this. This looks a recipe for mess.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,84 @@
...
+#    experiment_date: date | None = None 
+#    PublishedIn : PublishedInType
+#    validation_tags: dict[str, bool] #This is a dict{“properties_check”: T/F} (determined by a script not a user input. Should this be here or an annotation?) 
+#    contact_id : uuid.UUID | None = None
+
+#    class Config:
+#        from_attributes = True  # Allow mapping from SQLAlchemy objects
+
+
+class ScientificArtifactCreate(BaseModel,BrainRegionFilterMixin):
```
jdcourcol, line 44: should this derive from entityread?

app/db/model.py
dkeller9: Yes, it should.

app/schemas/scientific_artifact.py
dkeller9: Good point, I will investigate.

app/db/model.py
dkeller9: For published_in or validation_tags or both?

jdcourcol: for validation tags especially. I think this needs more specification. 
First, I don't think it should be a property of the scientific artefact but an assertion on the artefact. Which means that this is very close to the annotation pattern.
So more like another entity pointing to the artefact , with properties.
Second, who did run the validation, when , which code etc.. This should show up somewhere.

jdcourcol (changes requested):
I would say that we can merge the scientific artefact (assuming there are tests and end point added ..) as long as we remove the published_in & validation_tags for now.

I am against unstructured data for these use cases.
Having scientific artefact merged would help @chr-pok  to move forward for the circuit entity though.

dkeller9:
Okay, I will push a version without the unstructured data.

app/db/model.py
```diff
@@ -76,24 +89,75 @@ class LegacyMixin:
...
+        return relationship(
+            "BrainRegion",
+            primaryjoin=lambda: sa.or_(
+                cls.specific_brain_region_ids.any(BrainRegion.id),
+                cls.brain_region_id == BrainRegion.id
+            ),
+            uselist=True,
+            viewonly=True,
+            order_by="BrainRegion.brain_region_name"
+        )
```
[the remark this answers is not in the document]
dkeller9: Yes, specific_brain_regions is needed because there are artifacts that cannot be assigned to a specific brain region. We decided that they would always be children of the brain_region. IGiven the nature of data available in the literature, I think specific_brain_regions would be needed by most classes inheriting from ScientificArtifact. We can express it as an association_table I think. Is there an example in entitycore where this is done already for other features?

app/db/model.py
```diff
@@ -102,88 +166,198 @@ class Strain(Identifiable):
...
+        "polymorphic_identity": EntityType.scientific_artifact,
+        "inherit_condition": id == Entity.id,
+    }
+
+    @staticmethod
+    def is_scientific_artifact(entity_type: EntityType) -> bool:
+        return entity_type in [
+            EntityType.cell_morphology,
+            EntityType.electrical_cell_recording,
+        ]
```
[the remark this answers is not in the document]
dkeller9: It is just a sanity check because there were potential issues in deployment with cell_morphology.

app/db/model.py
```diff
@@ -76,24 +89,75 @@ class LegacyMixin:
...
+            "MEModel",
+            "Mesh",
+            "SingleNeuronSynaptome",
+            "SingleNeuronSimulation",
+            "SingleNeuronSynaptomeSimulation",
+            "ExperimentalNeuronDensity",
+            "ExperimentalBoutonDensity",
+            "ExperimentalSynapsesPerConnection",
+            "AnalysisSoftwareSourceCode"
+        ):
```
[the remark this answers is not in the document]
dkeller9: I had some odd issues during deployment and so added this check to be sure.

app/db/model.py
```diff
@@ -38,16 +46,21 @@
     Sex,
     SingleNeuronSimulationStatus,
     ValidationStatus,
+    MorphologyType,
```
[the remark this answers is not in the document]
dkeller9: Yes, it is in a separate PR to be considered after all issues with ScientificArtifact are resolved. I already ran ScientificArtifact with the updated Morphology Schema, but the Morphology Schema is something that we cannot easily swap out yet because there are so many morphologies already in use.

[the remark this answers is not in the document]
dkeller9: We can remove MorphologyType before merging.

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: I am not sure I understand. What is the difference between brain_region and specific_brain_regions ? and why "specific" btw ?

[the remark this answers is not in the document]
GianlucaFicarelli: > Given the nature of data available in the literature, I think specific_brain_regions would be needed by most classes inheriting from ScientificArtifact.

In that case, cannot we define only `specific_brain_regions` (and call it `brain_regions`), and remove `brain_region`?

An example of many-to-many association can be the one defined MTypesMixin, that uses the tables `mtype_classification` (the association table) and `mtype_class` (the table of unique mtypes).
Side note, mtype_class doesn't allow to specify different schemes of mtypes yet, but it can be changed in the future.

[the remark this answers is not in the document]
jdcourcol: brain_regions is used in explore to filter. We are adding ScientificArtefact for circuit CNS use case. I would postpone that modification after CNS.

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: so let's remove it

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: I would remove it then.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: Many external databases and prior literature use atlas annotation schemes that do not map directly to ours. It is sometimes possible to assign a single coarse brain region but this does not capture the exact location of data. For that, we need specific_brain_regions which is a set of regions in our data that best reflects the original region assignment. I would be more than happy to remove "brain_region" entirely and just use a list of brain regions, but it is my understanding that some filtering operations are based on the assumption of one brain region. So keeping a list of specific_brain_regions and a single brain_region could allow compatability to be maintained.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    DOI: str | None = None  # Optional DOI (Digital Object Identifier) as a string
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    original_source_location : str
+    other: str | None = None # Optional alternative identifier as a string
+
+from app.filters.common import LocationFilterMixin
+
+class ScientificArtifactBase(LocationFilterMixin, BaseModel):
```
eleftherioszisis, line 30: Schemas and filters are completely different things and cannot be mixed. Filters define how the user can query the read_many endpoint, whereas schemas determine the responses returned by the read/create endpoints.

dkeller9: Okay, I removed the filter dependency.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: MorphologyType is removed since this PR is just ScientificArtifact.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: Okay, it is removed

app/db/model.py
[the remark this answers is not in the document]
dkeller9: removed

app/db/model.py
dkeller9: validation_tags is removed

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,55 @@
...
+    creation_date: datetime| None = None
+    update_date: Optional[datetime] = None
+    name: str
+    description: Optional[str] = None
+    subject_id : uuid.UUID | None = None
+    license_id: Optional[UUID] = None
+    authorized_project_id: UUID
+    authorized_public: bool = False
+    createdBy_id: Optional[UUID] = None
+    updatedBy_id: Optional[UUID] = None
```
danielaegassan, line 40: Why is this not a child of Entity instead where most of these fields are? 

Isn't CreatedBy mandatory since it's the person that registers the entity in the database?

Does PublishedIn allow more than one entry?  I think we need that.

dkeller9: Fixed in latest commit.

jdcourcol: >Great, however, I'm still confused as to why we specify attributes in class ScientificArtifactBase that are already in entity?

yes this should be fixed.

jdcourcol: >Also the issue with some of them being optional that I believe should be mandatory e.g., CreatedBy

I think @eleftherioszisis made a PR for that and that should not be optional

danielaegassan: The thing is that if this is just inherited from Entity, then they will  be done consistently across all entities no? I don't get why we have to add them again.

PR #144, opened 2025-04-30
Morphology measurements v4

GianlucaFicarelli:
Similar to https://github.com/openbraininstitute/entitycore/pull/142, see the description copied below, but in this PR the active annotation is stored in `Entity.measurement_annotation_id` instead of `MeasurableEntity.measurement_annotation_id` (that's an abstract class).

This approach slightly simplifies the retrieval of the `is_active` attribute of the annotation, when multiple annotations are retrieved at the same time, at the cost of having a column `measurement_annotation_id` that can be null for entities not considered measurable.

---

Description copied from https://github.com/openbraininstitute/entitycore/pull/142:

- Make generic tables (not specific to morphologies):
	- morphology_feature_annotation -> measurement_annotation
	- measurement -> measurement_kind
	- measurement_serie_element -> measurement_item
- Allow to specify a single active annotation for each measurable entity, using MeasurableEntity.measurement_annotation_id (1-to-1 relationship)
- Allow to keep all the annotations until they are explicitly deleted, using measurement_annotation.entity_id (many-to-1 relationship)
- GET /measurement-annotation: search and retrieve a list of measurement annotations, based on metrics (see example 1)
- GET /measurement-annotation/{id}: retrieve a single measurement annotation
- POST /measurement-annotation: create a new measurement annotation
- GET /reconstruction-morphology: now it's possible to filter by any measurement annotation and metrics (see example 2)
- GET /reconstruction-morphology/{id}?expand=measurement_annotation: returns the specified morphology including the active measurement annotation

Example 1:
```
curl -v "http://127.0.0.1:8000/measurement-annotation?
measurement_kind__pref_label=terminal_path_lengths&
measurement_item__name=mean&
measurement_item__value__gte=152&
measurement_item__value__lte=153&
measurement_item__unit=%CE%BCm"
 -H "content-type: application/json" -H "virtual-lab-id: $VLAB_ID" -H "project-id: $VLAB_ID"
```

Example 2:
```
curl -v http://127.0.0.1:8000/reconstruction-morphology?
measurement_kind__pref_label=terminal_path_lengths&
measurement_item__name=mean&
measurement_item__value__gte=152&
measurement_item__value__lte=153&
measurement_item__unit=%CE%BCm"
 -H "content-type: application/json" -H "virtual-lab-id: $VLAB_ID" -H "project-id: $VLAB_ID"
```


app/service/measurement_annotation.py
```diff
@@ -0,0 +1,172 @@
...
+def _load_from_db_with_constraints(q: sa.Select, project_id: uuid.UUID | None) -> sa.Select:
+    q = constrain_to_accessible_entities(
+        q.join(Entity, Entity.id == MeasurementAnnotation.entity_id),
+        project_id=project_id,
+    )
+    return _load_from_db(q=q)
+
+
+def _get_join_onclause(filter_model: MeasurementAnnotationFilter):
+    """Allow to filter by `is_active` and `entity_type`."""
```
GianlucaFicarelli, line 53: I didn't find a cleaner way to integrate the filter by `is_active` and `entity_type` with the CustomFilter, because accessing the hybrid_property causes an error, so suggestions are welcome. On the other side, this is more explicit.

app/db/model.py
```diff
@@ -471,50 +489,99 @@ class MEModel(
...
-    measurements = relationship("MorphologyMeasurement", uselist=True)
 
+    @hybrid_property
+    def entity_type(self) -> str:
+        """Return the type of the associated entity."""
+        return str(self.entity.type)
+
+    @hybrid_property
+    def is_active(self) -> bool:
+        """Return True if this annotation is actively associated with an entity."""
```
jdcourcol, line 534: what is this ?

app/db/model.py
```diff
@@ -471,50 +489,99 @@ class MEModel(
...
-class MorphologyMeasurement(Base):
-    __tablename__ = "measurement"
+
+class MeasurementKind(Base):
+    __tablename__ = "measurement_kind"
     id: Mapped[int] = mapped_column(BigInteger, Identity(), primary_key=True)
-    measurement_of: Mapped[str] = mapped_column(index=True)
-    morphology_feature_annotation_id: Mapped[uuid.UUID] = mapped_column(
-        ForeignKey("morphology_feature_annotation.id"), index=True
+    pref_label: Mapped[str] = mapped_column(index=True)
```
jdcourcol, line 541: this is going to be repeated for each morphology no ? so the same definition will be repeated multiple times

app/cli/utils.py
```diff
@@ -654,3 +661,53 @@ def curate_age(data):
...
+    if not measurement_items:
+        # L.debug("measurement has no items")
+        return None
+    measurement_meta = measurement.get("isMeasurementOf", {})
+    definition = measurement_meta.get("label")
+    pref_label = measurement_meta.get("prefLabel") or to_pref_label(definition)
+    return MeasurementKind(
+        pref_label=pref_label,
+        definition=definition,
+        structural_domain=STRUCTURAL_DOMAIN_MAP[measurement.get("compartment")],
```
eleftherioszisis, line 697: structural_domain only makes sense for morphology annotations, no? What about other kind of annotations?

app/db/types.py
```diff
@@ -156,9 +156,25 @@ class MeasurementStatistic(StrEnum):
     sample_size = auto()
     standard_error = auto()
     standard_deviation = auto()
+    raw = auto()
+    minimum = auto()
+    maximum = auto()
+    sum = auto()
+    n = "N"  # TODO: rename to something different? sample_size? count?
```
eleftherioszisis, line 163: Should be sample_size from above, yeah.

app/db/types.py
```diff
@@ -156,9 +156,25 @@ class MeasurementStatistic(StrEnum):
...
     dimensionless = auto()
-    linear_density__1_um = auto()
-    volume_density__1_mm3 = auto()
+    linear_density__1_um = "1/μm"
+    volume_density__1_mm3 = "1/mm³"
+    linear__um = "μm"
+    area__um2 = "μm²"
+    volume__mm3 = "μm³"
+    angle__radian = "radian"
+    dhv = "dendritic volume / hull volume"  # TODO: what unit?
```
eleftherioszisis, line 174: It should be dimensionless, no?

app/db/model.py
GianlucaFicarelli: With the approach in this PR:
1. the /reconstruction-morphology endpoint returns a single measurement annotation attached to the morphology (the latest one)
2. the /measurement-annotation allows to retrieve all the annotations, also the old ones, so it's exposing the history of annotations (potentially filtered). The `is_active` attribute is returned for each annotation, and it would tell which ones are active, and the same attribute can also be specified in the request for filtering the returned data. However, if the `is_active` flag is not considered useful, it could be removed and some logic can be simplified.

eleftherioszisis: Why only one annotation? Is this a requirement for annotations in general? Wouldn't it be saner to just join all the annotations related to an entity?

Also, I am not sure I follow the old/new annotation classification. If it exists in the database and is related to an entity it should be joined, otherwise it should be deleted/replaced.

eleftherioszisis (approved):

app/db/model.py
GianlucaFicarelli: Yes, each annotation allows to specify different "measurement kinds" identified by pref_label, so the same kind is repeated, and there is no validation on the pref_label.
We can validate and restrict the possible kinds by adding another table with global uuids (not depending on the project_id?) that have to be retrieved and used to build a valid annotation payload.

In this way in the this table measurement_kind, `pref_label` and `definition` are going to be replaced by this global id (while I think we should keep structural_domain, if the same kind pref_label can be used with different domains).

app/db/model.py
GianlucaFicarelli: As discussed, we could change it and keep only the latest annotation, but we need to implement auditing in case data are overwritten by mistake.

app/cli/utils.py
GianlucaFicarelli: It's a good point, we should clarify what other entities require measurements so we can decide how to store these additional metadata. For example we could store them in a JSONB column if they aren't homogeneous across entity types (with some additional complexity if we need to be able to search in the content).

app/db/model.py
jdcourcol: how would that work @GianlucaFicarelli when you write an annotation you have to update the previous one and mark it as inactive ?
This is assuming you can do that (i.e you have the right to write the previous one).
isn't it fragile wrt concurrent writing ?

GianlucaFicarelli: With the approach in this PR, in the create_one endpoint, the entity is read for permission (because only the project owner can add the measurement annotation) and it's locked for update, and the lock is released at the end of the transaction after the entity is updated.

Instead, if we keep only one annotation, we can set a unique key on the column `entity_id` in the measurement_annotation table to ensure integrity, and not add the column `entity.measurement_annotation_id`, so it doesn't need to be updated.
However, when the annotation needs to be replaced we have a few options:
1. the client needs to call an update endpoint using the same annotation uuid, passing the new measurement_kinds and items (that completely replace the old kinds and items, but the old measurement_annotation uuid remains the same).
2. the client needs to delete the old measurement annotation using the old id, and issue a new post request to create the new measurement annotation

Both the options require that the client knows that the entity has a measurement annotation, and the uuid of the annotation to be updated or deleted, that may not be optimal but it's difficult to avoid if we want clean CRUD endpoints (i.e. we don't want to automatically delete the old annotation whenever a new annotation is created through POST).
The old measurement annotation uuid can be obtained for example:
- by calling the morphology endpoint before to get the morphology, including the annotation
- or by calling the measurement_annotation get_many endpoint searching by entity_id.

app/db/types.py
GianlucaFicarelli: not sure, is this a ratio between dendritic volume and hull volume?
In the existing data, the kind is always "mean_volume_density" and the statistic is "mean".

app/db/types.py
GianlucaFicarelli: I checked and the only measurement kind using N as unit is `soma_number_of_points`, it seems reasonable to change it to sample_size

app/db/model.py
```diff
@@ -471,50 +480,113 @@ class MEModel(
...
+        back_populates="measurement_kinds", viewonly=True
+    )
+    measurement_items: Mapped[list["MeasurementItem"]] = relationship(
+        back_populates="measurement_kind", passive_deletes=True
+    )
+    __table_args__ = (
+        UniqueConstraint(
+            "measurement_annotation_id",
+            "pref_label",
+            "structural_domain",
```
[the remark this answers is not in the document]
GianlucaFicarelli: In relation to the previous comment, all the annotations with `structural_domain` not null come from annotation with legacy_id of the form `https://bbp.epfl.ch/neurosciencegraph/data/annotations/{uuid}`
and with maximum creation_date `2022-08-15 17:03:51.113 +0200`
so they aren't imported because there are more recent annotations.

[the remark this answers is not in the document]
eleftherioszisis: Let\s leave the structural domain for now. When calculating morphometrics is often more useful to calculate them on specific tree types, rather than on the entire morphology. Usually, these calculations are used in classification and model parameter estimation for synthesis and all of those are usually performed on a per structural domain basis.

app/utils/entity.py
```diff
@@ -0,0 +1,10 @@
+from enum import StrEnum
+
+from app.db.model import Base, MeasurableEntity
+
+MEASURABLE_ENTITIES: dict[str, type[MeasurableEntity]] = {
+    mapper.class_.__tablename__: mapper.class_
+    for mapper in Base.registry.mappers
+    if issubclass(mapper.class_, MeasurableEntity) and mapper.class_.__tablename__
+}
+MeasurableEntityType = StrEnum("MeasurableEntity", list(MEASURABLE_ENTITIES))
```
eleftherioszisis, line 10: Wouldn't this model be better at app/db/types.py ?

GianlucaFicarelli: It cannot be moved to app/db/types.py because it would cause a circular dependency:
app/db/model.py -> app/db/types.py -> app/db/model.py

But I can move it to app/db/utils.py

eleftherioszisis: Ah, I see.

eleftherioszisis (approved):

PR #146, opened 2025-05-02
expand me-model to include all morphology/emodel read schema

bilalesi:


eleftherioszisis (approved):
I moved AssetsMixin into EModelRead to minimize the amount of joins needed in the service, given that the assets are what's needed by @bilalesi 

fyi @g-bar

PR #147, opened 2025-05-02
add types to entities read schema

bilalesi:
I was thinking of another way to do it using another type-mixin
ex:
```python
class EntityTypeMixin(BaseModel):
    type: str | None = None
```
and then update the common query methods; 
```python
def router_read_one[T: BaseModel, I: Identifiable](
    *,
    id_: uuid.UUID,
    db: Session,
    db_model_class: type[I],
    authorized_project_id: uuid.UUID | None,
    response_schema_class: type[T],
    apply_operations: ApplyOperations[I] | None,
) -> T:
    # ... old code
    validated_data = response_schema_class.model_validate(row)
    if hasattr(validated_data, "type") and hasattr(db_model_class, "__tablename__"):
        validated_data = response_schema_class.model_validate(
            {
                **validated_data.model_dump(),
                "type": db_model_class.__tablename__,
            }
        )

    return validated_data**
```

but find out to have it in the response read schema is more readable and no need for another validation change

### DIFF
--- a/app/schemas/base.py
+++ b/app/schemas/base.py
@@ -1,10 +1,13 @@
 import uuid
-from datetime import datetime, timedelta
-from typing import Annotated
+from datetime import datetime
 
-from pydantic import UUID4, BaseModel, ConfigDict, Field, model_validator
+from pydantic import UUID4, BaseModel, ConfigDict
 
-from app.db.types import AgePeriod, Sex
+from app.db.types import EntityType
+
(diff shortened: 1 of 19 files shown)


eleftherioszisis (approved):
I used the `type` that is stored in the entity table and can be used to populate the read schema. As far as entities are concerned it should not be required to modify the router logic.

PR #148, opened 2025-05-02
initial version of within_brain_region

mgeplf:


app/db/model.py
```diff
@@ -80,14 +81,25 @@ class Identifiable(TimestampMixin, Base):
...
-    # See https://github.com/openbraininstitute/core-web-app/blob/cd89893db3fe08a1d2e5ba90235ef6d8c7be6484/src/types/ontologies.ts#L7
-    id: Mapped[int] = mapped_column(BigInteger, primary_key=True, autoincrement=False)
     name: Mapped[str] = mapped_column(unique=True, index=True)
-    acronym: Mapped[str] = mapped_column(unique=True, index=True)
-    children: Mapped[list[int] | None] = mapped_column(ARRAY(BigInteger))
+
+
+class BrainRegion(Identifiable, Base):
+    __tablename__ = "brain_region"
+    ROOT_PARENT_UUID = uuid.UUID("00000000-0000-0000-0000-000000000000")
```
mgeplf, line 92: I'm not sure I like this; the alternative is to make `parent_structure_id` be an "FK" into this table, which I think I would prefer - it would at least ensure consistency, but then I have to make it nullable. Thoughts?

app/routers/brain_region.py
```diff
@@ -1,15 +1,14 @@
...
     prefix="/brain-region",
     tags=["brain-region"],
 )
 
-read_hierarchy = router.get("")(app.service.brain_region.read_hierarchy)
+read_many = router.get("")(app.service.brain_region.read_many)
 read_one = router.get("/{id_}")(app.service.brain_region.read_one)
-create_one = router.post("", dependencies=[Depends(user_with_service_admin_role)])(
-    app.service.brain_region.create_one
+read_hierarchy_name_hierarchy = router.get("/{hierarchy_name}/")(
```
mgeplf, line 12: This is according to the spec I wrote, but I wonder if it should be more clearly namespaced: currently `read_one` looks nearly the same as getting the hierarchy.  Maybe it should be `get("hierarchy/{hierarchy_name}")`?

app/dependencies/common.py
```diff
@@ -122,6 +123,27 @@ def __call__(self, q: sa.Select[tuple[T]], vector_col: InstrumentedAttribute):
...
 
 
+class InBrainRegionQuery(BaseModel):
+    within_brain_region: str | None = None
+
+    def __call__(self, query: sa.Select, db_model_class):
+        if not self.within_brain_region:
+            return query
+
+        [hierarchy_id, hierarchy_name, *with_ascendants] = self.within_brain_region.split(",")
```
mgeplf, line 133: this also follows the "spec", but I'm not sure I like having a comma separated query param. Thoughts?

mgeplf: I should add that I also don't like that I'm using the "hierarchy_id" (ie: 997 for root), and not the "id" from the database (ie: UUID)

g-bar: FastAPI supports list query params.
It’s also possible for caller to simply define multiple values instead of string with comma separated values 

?param=1&param=2

https://fastapi.tiangolo.com/tutorial/query-params-str-validations/

app/cli/import_data.py
```diff
@@ -1092,7 +1099,7 @@ def _do_import(db, input_dir, project_context):
     importers = [
         ImportAgent,
         ImportAnalysisSoftwareSourceCode,
-        ImportBrainRegionMeshes,
+        # ImportBrainRegionMeshes,
```
mgeplf, line 1102: there are meshes that correspond to regions that don't exist; my plan is to only import the latest atlas release, instead of polluting the database with unnecessary meshes.
That ok? (ping @jdcourcol )

eleftherioszisis: This is how I did go about for the CellComposition too. Otherwise, there is a lot of garbage. We can carefully pick what we need and only import those datasets.

app/routers/brain_region.py
eleftherioszisis: How about `/ontology/hierarchy/{hierarchy-name}` so that we can have this router to do the usual operations read_one/read_many/create_one for BrainRegion entities and we don't mix the two?

app/cli/import_data.py
mgeplf: Cool, good to know - I will prob. do the same.

jdcourcol: yes, only import what is referenced by the latest atlas release.

app/cli/import_data.py
```diff
@@ -274,7 +275,7 @@ def is_correct_type(data) -> bool:
 
     @staticmethod
     @abstractmethod
-    def ingest(db, project_context, data_list, all_data_by_id: dict[str, Any]):
+    def ingest(db, project_context, data_list, all_data_by_id: dict[str, Any], hierarchy_name: str):
```
jdcourcol, line 278: we need multiple hierarchy name to import legacy data ?

app/routers/brain_region.py
eleftherioszisis: In an atlas release where the hierarchy is linked by id, how could I get the metadata of the atlas hierarchy corresponding to it?

app/cli/import_data.py
mgeplf: No, but I also didn't want to hard-code it, so the name is plumbed through the whole import

app/db/model.py
```diff
@@ -80,14 +81,25 @@ class Identifiable(TimestampMixin, Base):
...
+
+    hierarchy_id: Mapped[int] = mapped_column(BigInteger, index=True)
```
eleftherioszisis, line 94: Would `parcellation_id/voxel_id` make more sense in hinting what it actual is this id for?

mgeplf: dunno, I meant it as "the id that you would find in the original hierarchy"

app/cli/import_data.py
jdcourcol: yeah, I figure out . that is a "lot" of modification for a unique name.

app/routers/brain_region.py
mgeplf: > In an atlas release where the hierarchy is linked by id, how could I get the metadata of the atlas hierarchy corresponding to it?

I don't understand the question.

This is disjoint of an atlas, but an atlas uses a hierarchy, so I would assume it has a FK to the `BrainRegionHierarchyName`, maybe that answers it?

app/db/model.py
eleftherioszisis: Sure, but at the end of the day it's the id value in the annotation volume, which is what's it will be used for. Also, when I see `hierarchy_id` my brain is hardwired in expecting an atlas-hierarchy entity id, so I need a second to sync with reality. :D

app/routers/brain_region.py
mgeplf: I'm not sure I see the benefit of having another endpoint (`/ontology/*`) to separate things; it's not intuitive, to me, what would belong there versus having, for instance, the all the `brain-region` functionality hanging off the same endpoint.

jdcourcol: I would prefer the id than the name.

mgeplf: Ok, done.

mgeplf: With respect to where to download the hierarchy, I could also put it under:
`/brain-region-hierarchy-name/{uuid}/hierarchy?view=vertical`; that probably makes the most sense.

mgeplf: Implemented the changes and pushed.

eleftherioszisis: > With respect to where to download the hierarchy, I could also put it under: `/brain-region-hierarchy-name/{uuid}/hierarchy?view=vertical`; that probably makes the most sense.

How about `/brain-region-hierarchy/{uuid}/download` ?

app/service/brain_region_hierarchy_name.py
```diff
@@ -0,0 +1,94 @@
...
+        data=[BrainRegionHierarchyNameRead.model_validate(d) for d in data],
+        pagination=PaginationResponse(
+            page=pagination_request.page,
+            page_size=pagination_request.page_size,
+            total_items=total_items,
+        ),
+        facets=None,
+    )
+
+    return response
```
GianlucaFicarelli, line 39: Couldn't this function call `common.router_read_many`? or are there differences for this case?

app/dependencies/common.py
g-bar: Ah right, didn't realize about the types.

> One option would be to make them separate params, but with a common prefix?

That would seem like the best option to me.

app/db/model.py
g-bar: Does it have to be nullable? We could have a pre-defined UUID for the root, like we do for the public project, vlab.

mgeplf: Ok, let's go nullable then.
I agree about preferring an FK to keep integrity.
Adding a sentinel makes the queries slightly more complicated (so the sentinel never gets returned)

app/dependencies/common.py
mgeplf: Ok, will do.

app/db/model.py
```diff
@@ -80,14 +81,25 @@ class Identifiable(TimestampMixin, Base):
...
+    __tablename__ = "brain_region_hierarchy_name"
 
-    # See https://github.com/openbraininstitute/core-web-app/blob/cd89893db3fe08a1d2e5ba90235ef6d8c7be6484/src/types/ontologies.ts#L7
-    id: Mapped[int] = mapped_column(BigInteger, primary_key=True, autoincrement=False)
     name: Mapped[str] = mapped_column(unique=True, index=True)
-    acronym: Mapped[str] = mapped_column(unique=True, index=True)
-    children: Mapped[list[int] | None] = mapped_column(ARRAY(BigInteger))
+
+
+class BrainRegion(Identifiable, Base):
```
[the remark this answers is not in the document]
mgeplf: done

app/service/brain_region_hierarchy_name.py
mgeplf: done: it's about the same length (but yes, I understand DRY)

app/schemas/base.py
```diff
@@ -61,10 +61,13 @@ class LicenseRead(LicenseCreate, CreationMixin, IdentifiableMixin):
 
 class BrainRegionCreate(BaseModel):
     model_config = ConfigDict(from_attributes=True)
-    id: int
+    id: uuid.UUID
```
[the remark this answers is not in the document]
mgeplf: I just kept BrainRegionRead

app/service/brain_region.py
```diff
@@ -1,59 +1,45 @@
...
+
+    response = ListResponse[BrainRegionRead](
+        data=[BrainRegionRead.model_validate(d) for d in data],
+        pagination=PaginationResponse(
+            page=pagination_request.page,
+            page_size=pagination_request.page_size,
+            total_items=total_items,
+        ),
+        facets=None,
+    )
```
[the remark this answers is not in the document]
mgeplf: done

eleftherioszisis:
Would it be possible to rename brain-region-hierarchy-name into brain-region-hierarchy? Not sure why we need the name postfix, given that we fetch it by uuid.

app/db/model.py
mgeplf: ok; I've chosen `annotation_value`; anything w/ `_id` in it suggests it refers to a table in the DB, so I wanted to avoid that. I also wanted to avoid anything with `hierarchy`, since that's also confusing.

eleftherioszisis (approved):
Neat!

PR #149, opened 2025-05-02
rename contribution facet to contributions

bilalesi:
the frontend rely heavily on the naming fields to be consistent on all type of entities and facets
so fields in response should match facets fields

### DIFF
--- a/app/queries/facets.py
+++ b/app/queries/facets.py
@@ -4,8 +4,8 @@ from app.dependencies.common import FacetQueryParams
 brain_region: dict[str, FacetQueryParams] = {
     "brain_region": {"id": BrainRegion.id, "label": BrainRegion.name},
 }
-contribution: dict[str, FacetQueryParams] = {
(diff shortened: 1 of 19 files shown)


eleftherioszisis (approved):
Makes sense.

GianlucaFicarelli:
Only the contribution need to be changed? There was already [this PR](https://github.com/openbraininstitute/entitycore/pull/70) with some discussion for singular/plural fields, and the facets were made to match the singular form used for the the filter.

bilalesi:
> Only the contribution need to be changed? There was already [this PR](https://github.com/openbraininstitute/entitycore/pull/70) with some discussion for singular/plural fields, and the facets were made to match the singular form used for the the filter.

okey, since you have already an agreement on naming (and actually make sense for plural and singular), 
I fix it in the frontend so we keep the same logic, I will close this PR,
thank you all

PR #153, opened 2025-05-07
Add a label column to Asset to allow classification by name.

eleftherioszisis:
Until now there was no good way to classify assets of the same content type under the same entity.

The changes in this PR allow to assign a label to an asset when uploading.

I would like to use it is CellComposition which has the volumes and summary, both json assets under the same entity. By assigning the labels:
```
cell_composition_summary
cell_composition_volumes
```
Allows fetching the correct asset by label.


app/db/model.py
```diff
@@ -769,6 +769,7 @@ class Asset(Identifiable):
     size: Mapped[BIGINT]
     sha256_digest: Mapped[bytes | None] = mapped_column(LargeBinary(32))
     meta: Mapped[JSON_DICT]  # not used yet. can be useful?
+    label: Mapped[str | None]
```
GianlucaFicarelli, line 772: Wouldn't it be safer if we define the label as a StrEnum?
We could also consider to validate the value depending on the entity type of the entity associated to the asset

eleftherioszisis: Done.

app/errors.py
```diff
@@ -119,3 +121,19 @@ def ensure_authorized_references(
...
+    error_message: str, error_code: ApiErrorCode = ApiErrorCode.GENERIC_ERROR
+) -> Iterator[None]:
+    """Context manager that raises ApiError when a schema validation error is raised."""
+    try:
+        yield
+    except ValidationError as err:
+        raise ApiError(
+            message=error_message,
+            error_code=error_code,
+            http_status_code=HTTPStatus.FORBIDDEN,
```
GianlucaFicarelli, line 137: 403 FORBIDDEN is usually used for auth errors, while for validation error we should use 422 UNPROCESSABLE_ENTITY (alternatively 409 HTTPStatus.CONFLICT but it seems less appropriate)

app/errors.py
```diff
@@ -119,3 +121,19 @@ def ensure_authorized_references(
                 message=error_message, error_code=error_code, http_status_code=HTTPStatus.FORBIDDEN
             ) from err
         raise
+
+
+@contextmanager
+def ensure_valid_schema(
+    error_message: str, error_code: ApiErrorCode = ApiErrorCode.GENERIC_ERROR
```
GianlucaFicarelli, line 128: It can be ApiErrorCode.INVALID_REQUEST instead of ApiErrorCode.GENERIC_ERROR, although it can be overridden in the call

eleftherioszisis: Sure.

GianlucaFicarelli (approved):

PR #155, opened 2025-05-11
Add brain region mesh routers/schemas/service

eleftherioszisis:
Allow querying brain region meshes using brain region ids.


mgeplf (commented):
I thought this was going to be part of the BrainAtlas work, so I've been putting somewhat equivalent functionality (along w/ the volumes) into the implementation for the spec: https://github.com/openbraininstitute/entitycore/pull/168

mgeplf:
IMO, a mesh is only useful within the context of a BrainAtlas (ie: you don't need to get all the meshes); as such, I think it should be modeled as an asset "attached" to a `BrainAtlasRegion` (ie: a region within the `BrainAtlas`)

eleftherioszisis:
Having a BrainAtlasRegion with a mesh asset and other information created due to the atlas context makes perfect sense. I still think though we need to be able to also query them independently, but that can come later if needed.


I will close this PR.

PR #163, opened 2025-05-13
Add Cell Composition Volumes transformation NEXUS -> NTTCore

eleftherioszisis:


app/cli/curation/cell_composition.py
```diff
@@ -69,14 +72,26 @@ def _curate_summary_asset(
     return {new_digest: str(target_file)}
 
 
-def _curate_volumes_asset(*, asset: Asset, is_dry_run: bool) -> dict[str, str]:
-    # TODO: Add curation
+def _curate_volumes_asset(
```
GianlucaFicarelli, line 75: Looking at the content of the file, it seems to me that it's duplicating what is present in the db.
I can get a list of me_type_density.id, mtype, etype with a query like:

```sql
select e.id, mc2.pref_label, ec2.pref_label
from me_type_density mtd
join entity e on mtd.id = e.id
join mtype_classification mc1 on mtd.id = mc1.entity_id
join mtype_class mc2 on mc2.id = mc1.mtype_class_id 
join etype_classification ec1 on mtd.id = ec1.entity_id
join etype_class ec2 on ec2.id = ec1.etype_class_id
```

However, the result is 324 records, while the file has 248. Am I missing something in the query?
How are the cell composition records related to the me_type_density records?

eleftherioszisis: There may be multiple densities for me combinations. There is no context that determines which densities belong to that CellComposition. The composition manipulation task for example would generate exactly the same me composition with new densities.

GianlucaFicarelli: I'm a bit concerned that we are storing relations between records into a json file, instead of using tables that ensure referential integrity.
Do I understand correctly that to store the relations in the db we should have a many-to-many relationship between `cell_composition` and `me_type_density`?

eleftherioszisis: Yeah, there are two datasets associated with a CellComposition. The first is the summary that lists a set of metrics per region/mtype/etype and the second one is the volumes distribution which lists per mtype/etype the respective nrrd volume density.

eleftherioszisis: I forgot to mention that a cell composition model is created using a specific brain hierarchy.

eleftherioszisis:
I am closing this for now. To be revisited and define proper database relationships.

PR #165, opened 2025-05-14
fix `brain_region` not filtering on morphology endpoint, add facets and more filters to ion-channel-model

mgeplf:
* add tests for `brain_region` filters while auditing that the rest worked.


mgeplf:
hat tip to @jankrepl for catching that the morphology endpoint wasn't filtering on `brain_region`

eleftherioszisis (approved):
Nice.

PR #166, opened 2025-05-14
Validation proposal

dkeller9:
I'm opening this to PR to get feedback on the validation schema.

### DIFF
--- a/app/db/validation.py
+++ b/app/db/validation.py
@@ -1,0 +1,40 @@
+from sqlalchemy import Column, Integer, String, Table, ForeignKey
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship
+from sqlalchemy.dialects.postgresql import ARRAY  # Using PostgreSQL's array type for sets
+
+Base = declarative_base()
+
+class Validation(Base):
+    __tablename__ = "validations"
+
+    id = Column(Integer, primary_key=True, index=True)
+    entity_type = Column(String, nullable=False, index=True)
+    
+    # Using PostgreSQL ARRAY type to store sets of strings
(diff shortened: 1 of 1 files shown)


app/db/validation.py
```diff
@@ -0,0 +1,40 @@
...
+from sqlalchemy.orm import relationship
+from sqlalchemy.dialects.postgresql import ARRAY  # Using PostgreSQL's array type for sets
+
+Base = declarative_base()
+
+class Validation(Base):
+    __tablename__ = "validations"
+
+    id = Column(Integer, primary_key=True, index=True)
+    entity_type = Column(String, nullable=False, index=True)
```
jdcourcol, line 12: I am not sure I understand :( . Can you give an example of 1 instance of this ? and how it would be consumed ?

dkeller9: It might be easier to look at the schema rather than the data model. So let's say the user contributes a new morphology. The validation logic looks up the morphology entity type in the table then applies the "must_pass_to_upload" validations. If these are successful it runs the validations in "must_run_upon_upload" which annotates the entity, and "must_pass_to_simulate" which generates additional validations needed for the artifact to be simulated.

jdcourcol: ok, then it is not a validation, it is rather something like a "validation plan".

also what does the list of string contains ? I know it is a list of validations. but how do we go from a string to the actual execution of a code ?

dkeller9: This is the validation plan schema stored in the database. The string contains a relative path in a github repo, for example "/morphologies/validate_morphology.py" I wrote a simple "validate_morphology.py" script that sees if it can be read in by NeuroM and returns an error if not. There will need to be some support logic that processes the lists, which I can add to entitycore once we are satisfied with this plan.

PR #167, opened 2025-05-14
Add currents to MEModel, add assets to simulation response

g-bar:
Required for:
https://github.com/openbraininstitute/prod-platform-architecture/issues/16


app/schemas/me_model.py
```diff
@@ -24,6 +24,8 @@ class MEModelBase(BaseModel):
     name: str
     description: str
     validation_status: ValidationStatus = ValidationStatus.created
+    holding_current: float = 0.0
+    threshold_current: float = 1.0
```
g-bar, line 28: What are good default values @AurelienJaquier ?

app/cli/import_data.py
```diff
@@ -848,7 +848,10 @@ def ingest(db, project_context, data_list, all_data_by_id, hierarchy_name: str):
                 strain_id=morphology.strain_id,
                 creation_date=createdAt,
                 update_date=updatedAt,
+                holding_current=data.get("holding_current", 0),
+                threshold_current=data.get("threshold_current", 1),
```
eleftherioszisis, line 852: If they are not available I would prefer if they could be set to None. Having holding and threshold currents as 0/1 or an arbitrary value will result into unusable models.

app/db/model.py
```diff
@@ -488,6 +488,9 @@ class MEModel(
         "ReconstructionMorphology", foreign_keys=[morphology_id], uselist=False
     )
 
+    holding_current: Mapped[float] = mapped_column(default=0)
+    threshold_current: Mapped[float] = mapped_column(default=1)
```
eleftherioszisis, line 492: Same here. There should be no defaults.

app/schemas/me_model.py
g-bar: Nvm, see Eleftherios' comments.

app/cli/import_data.py
g-bar: I think they are available in all existing MEModels in nexus, so I will make non nullable.

g-bar: There is an older MEModel in the december dump  MEM__1372346-C060114A5__cADpyr_L5_TPCA, but it's not there in the current data set (prod). This one does not have any currents defined, so I will just ignore that one.

eleftherioszisis:
Looks good to me. Maybe @AurelienJaquier has some comments?

AurelienJaquier:
This also looks good to me. I just have one question: will we be able to register new ME-Models on entity-core with holding_current and threshold_current set to None?
I am asking because I believe when we create ME-Models on the platform we don't have any holding or threshold currents. Those will be computed afterwards when we validate the ME-Model (the validation script in still in progress), and we will update the resource with them at this point.

eleftherioszisis:
IMO the simulation should not be allowed if the memodel is not conformant to its requirements. bluenass should not do detective work in order to dig up what it needs...

Provenance data should be optional, which means that the MEModel should also work without it.

g-bar:
Yes, I would agree with that, but IMHO this requires further discussion because that departs significantly with what we have now, going beyond migrating the db and changing product behaviour. (For now they would now need to spend more credits to first validate the model, then run the simulation).

Any thoughts @jdcourcol ?

AurelienJaquier:
The thing is that when we create an ME-Model by combining a hoc file (i.e. EModelScript) with a morphology, then the model behavior is different form the one that was used to optimise the EModel (if the morphology is different), meaning that the holding and threshold currents become different also. We cannot just get them from EModelScript.
In my opinion, there is no escaping the re-computation of the threshold current, whether in validation, or elsewhere.

g-bar:
If there is no fallback we cannot run the sim at all, because there are no currents. So we do that, disable sims for unvalidated models?

jdcourcol:
We can put default values.  As @AurelienJaquier said, there are no good values.

AurelienJaquier:
Do we need the holding and threshold on the platform, though?
On the platform, holding current is given by the user, and we do not use threshold-based stimulation, only absolute current stimulation.
I think we can run the models on the platform even if they do not have holding and threshold current. If you still need default values for the code to work properly, you can use 0 for holding current and 0.1 for threshold current

g-bar:
> Do we need the holding and threshold on the platform, though? On the platform, holding current is given by the user, and we do not use threshold-based stimulation, only absolute current stimulation. I think we can run the models on the platform even if they do not have holding and threshold current. If you still need default values for the code to work properly, you can use 0 for holding current and 0.1 for threshold current

A thanks, yeah they're used on initialization, but didn't know the treshold current wasn't used. So I'll put those defaults on bluenaas and make the fields nullable here.

AurelienJaquier (approved):

PR #168, opened 2025-05-14
initial atlas doc

mgeplf:


docs/API.md
```diff
@@ -212,6 +212,48 @@ GET /reconstruction-morphology?within_brain_region_hierachy_id=3f41b5b5-4b62-40d
...
+
+
+`GET brain-atlas/$UUID/regions`
+
+[{
+    "brain_region_id: $UUID,
+    "volume": 33
+}]
+The volume is in um^3.
+Note: It returns only the leaf-nodes, and is meant to be used with the associated hierarchy.
```
mgeplf, line 254: there are pluses and minuses with storing all the volumes all the way down, but my suspicion is that in addition to allow for multiple views, it's probably faster to download less data and compute it locally :)

docs/API.md
```diff
@@ -212,6 +212,48 @@ GET /reconstruction-morphology?within_brain_region_hierachy_id=3f41b5b5-4b62-40d
...
+    "id": $UUID,
+    "name": "human_v0.1.1",
+    "species_id": $UUID,
+    "hierarchy_id: $UUID,
+    "update_date": "2025-05-09T13:32:18.034672Z"
+}
+```
+
+
+`GET brain-atlas/$UUID/regions`
```
mgeplf, line 247: I'm not sure how happy we are having "depth" (ie: not just top level routes); I'm open to other ides.
We could also potentially make these `annotations` or `measurements`, but I think that makes everything more complicated than it needs to be.

docs/API.md
```diff
@@ -212,6 +212,48 @@ GET /reconstruction-morphology?within_brain_region_hierachy_id=3f41b5b5-4b62-40d
...
+
+[{
+    "brain_region_id: $UUID,
+    "volume": 33
+}]
+The volume is in um^3.
+Note: It returns only the leaf-nodes, and is meant to be used with the associated hierarchy.
+By only storing the leaf nodes, it composes with the different views by clibing the tree, and summing all the children along the way.
```
mgeplf, line 255: In addition; this makes it clear which nodes are leaf nodes; anything that isn't in this list is either not in the atlas, or a composite node.

docs/API.md
jdcourcol: you don't mention the mesh ? There is mesh & volumes mentioned in the introduction but in the body I see only volume.

jdcourcol, line 254: should we then call the end point `leaves` not `regions` ?

mgeplf: > you don't mention the mesh ?

Yeah, I'm not sure how we'll handle those yet; perhaps another endpoint `brain-atlas/$UUID/meshes`?

> should we then call the end point leaves not regions ?

Sure, but let's decide if we're happy with this first - ie: the `depth` thing.

mgeplf: Yeah, I'm not sure how we'll handle those yet; perhaps another endpoint brain-atlas/$UUID/meshes?

eleftherioszisis: What if one needed a list of all region data, e.g. :
```
[{
    "brain_region_id": $UUID,
    "brain_region_mesh_id": $UUID,
    "volume": 33,
}]
```
In addition, I wonder would it be useful instead of getting a list of all regions, to get instead a hierarchy 
```
brain-atlas/$UUID/hierarchy?with=region_id&with=brain_region_mesh&with=region_volume
```
Resulting in a hierarchy with all that data added to it. Not sure if that would be more useful than the list above though.

docs/API.md
```diff
@@ -212,6 +212,48 @@ GET /reconstruction-morphology?within_brain_region_hierachy_id=3f41b5b5-4b62-40d
 In other words, the name of the hierarchy, and the id which will be recursively included.
 This can happen in either by the `descendants` (the default) or by `ascendants`.
 
+# Brain Atlas:
+
+A BrainAtlas is an volumetric concept describing the locatations of brain regions in space.
+It is composed of an `annotation` (also known as `parecellation`) of a voxels, storred in an NRRD file.
```
jdcourcol, line 218: typo

docs/API.md
jdcourcol: I don't understand.

docs/API.md
mgeplf: > perhaps another endpoint brain-atlas/$UUID/meshes?

I should note that the meshes won't just be at the leaves; they need to be at the levels that are required for the interface.

> Not sure if that would be more useful than the list above though.

I'm my mind, the list can be converted to a dict based on ID; and then queried w/ as required as one traverses the hierarchy.

docs/API.md
mgeplf: so far, other than the `/hierarchy`, everything (?) is just at the base route, ie: `/reconstruction-morphologies`, `/experimental-bouton-density`, etc.

Here I'm considering adding more depth: ie: `regions` in `brain-atlas/$UUID/regions`; this breaks the symmetry that we have with all the other endpoints (except `brain-region-hierarchy`, which I consider special)

jdcourcol: but that would still be 
/regions/$UUID for a particular region no ?

docs/API.md
jdcourcol: what is the `depth` thing ?

eleftherioszisis: > should we then call the end point `leaves` not `regions` ?

If we are going to have multiple end points, e.g. regions/meshes/etc. I would prefer if all of them exhibited the same behavior regarding returning leaves or intermediate nodes.

That said, the frontend probably needs the meshes of all regions, intermediate included. Mike suggested having a "leaves" flag to switch between leaves and all regions.

docs/API.md
eleftherioszisis: I don't think we can avoid this additional "depth" for this functionality. brain-atlas will need to be treated in a special way from the sdks, and the same is true for the hierarchy.

May I suggest some consistency with the entities endpoints for regions and meshes? For example, the endpoints could be:
```
brain-atlas/$UUID/brain-region
brain-atlas/$UUID/brain-region-mesh
```
To match the /brain-region and /brain-region-mesh (open PR) routes.

docs/API.md
eleftherioszisis: Ah, my bad for the misunderstanding. That's what I wanted to hear I guess. :)

docs/API.md
mgeplf: (I had those in my local branch already)

jdcourcol: yeah just saw that. I was about to comment about the negative number for non leaves. but it is replaced by null now.

mgeplf: no; either is_leaf_region = True && volume is a number, or is_leaf_region=False && volume is none.

jdcourcol:
@mgeplf can we merge this ?

jdcourcol (approved):

PR #170, opened 2025-05-14
Raise for SAWarning and fix filter queries

GianlucaFicarelli:


g-bar (approved):
Ah good one!

eleftherioszisis (approved):
Much needed indeed!

app/logger.py
```diff
@@ -99,3 +101,13 @@ def configure_logging() -> int:
...
+
+
+def configure_warnings():
+    """Configure warnings.
+
+    Raise an error in case of potentially wrong queries, for example:
+    - SELECT statement has a cartesian product
+    - An alias is being generated automatically
+    """
+    warnings.simplefilter("error", sqlalchemy.exc.SAWarning)
```
GianlucaFicarelli, line 113: Side note, it's not possible to set the env variable PYTHONWARNINGS to a submodule because it will cause:
```
PYTHONWARNINGS=error::sqlalchemy.exc.SAWarning python
Invalid -W option ignored: invalid module name: 'sqlalchemy.exc'
Python 3.12.8 (main, Jan 14 2025, 23:36:58) [Clang 19.1.6 ] on darwin
```
although it should be imported [here](https://github.com/python/cpython/blob/2dc476bcb9142cd25d7e1d52392b73a3dcdf1756/Lib/warnings.py#L263)

PR #171, opened 2025-05-15
validation result model

jdcourcol:
The immediate need is to store things like: 
grafik

We want to store the result of the execution of the validation and the generated assets for display in the GUI. 

The PR does not address:
- the definition of the validation itself (code, what is the validation for, description...), which should be a separate entity.
- the provenance (when was the validation executed, who executed it ...), which should follow standard model with activity, contributors, agents...
- other complications like "score"
- change over time: newer execution results may show up, invalidating the previous ones.

To be added later:
- the schema migration, REST api, import and tests


AurelienJaquier:
I don't see any entry for the generated figure in your ValidationResult model. Would the figure be stored in the ValidationResult, or would it be stored elsewhere?

jdcourcol:
The figure is stored as an asset. Entities can have assets (usually a file).

AurelienJaquier:
Ok, I see. I have another question. I see that the ValidationResult contains a link to the validated entity. Would the validation result show in the validated entity, like the annotations? Or do we have to manually add an annotation into the validated entity linking to the ValidationResult?

jdcourcol:
no, there is no annotation to add. The UI can fetch all the ValidationResult pointing to the displayed entity.

chr-pok:
Overall looks good to me!

Just one additional item would be required that is outlined in [General_schema.docx](https://openbraininstitute.sharepoint.com/:w:/s/OBI-Scientificstaff/EQhs6O6doJ5GmEnNhVmeQCYBPLrBgs0njgUDSt0xVGam4w?e=fN8zG0), namely some kind of logging information for the user. The idea is to store additional details on the validation run, e.g., if the validation was skipped, if there were any warnings, why the validation failed etc. Otherwise, it will be difficult for a user to know how to fix the artifact in case the validation failed.

I imagine this could be either another asset containing a log file, or better an optional string with (human-readable) details on the validation run. We called it `notes` in the docx but perhaps `validation_details` would be better.

jdcourcol:
yes, that would be another asset indeed.

AurelienJaquier:
I have modified the General_schema.docx to rename notes into validation_details, and make it an asset

dkeller9:
@jdcourcol some validation results will be in the format of key: value, where the value is of either boolean or a number. Should we make these results as an asset or is it better to have a dedicated structure in entitycore?

dkeller9:
I am worried that if we add this in now without handling version and provenance it will be hard to migrate in the future. I would suggest to have a meeting on Tuesday to discuss this.

jdcourcol:
yes, but we can add that when we face it. A good reference would be https://sciunit.readthedocs.io/en/latest/basics.html#score

jdcourcol:
> I am worried that if we add this in now without handling version and provenance it will be hard to migrate in the future.

Can you elaborate why ?

dkeller9:
If version and provenance information is not present then what would be migrated? Do we just assume that what's in there is the base version and that all validations have been run on it? If so that's fine but we need to be sure to rerun the validations any time things change.

jdcourcol:
For the time being, we are going to import the existing validation results stored in the knowledge graph.

eleftherioszisis (approved):

PR #174, opened 2025-05-17
Add simulations to me model

g-bar:


g-bar:
Actually not needed can only call the simulations enpoint and filter by emodel.

PR #178, opened 2025-05-19
Improve the queries for filtering and retrieving facets by using only the required joins

GianlucaFicarelli:
Improve the queries for filtering and retrieving facets by using only the required joins.
See the issue description at https://github.com/openbraininstitute/prod-platform-architecture/issues/68

- With a populated db and using the default page=1, page_size=100
time curl -vs 'http://127.0.0.1:8000/reconstruction-morphology?with_facets=true' > /dev/null

```
~ 4.6 seconds with the main branch 
~ 0.16-0.22 seconds with this branch
```

- With a populated db and using a custom page=1, page_size=10
time curl -vs 'http://127.0.0.1:8000/reconstruction-morphology?with_facets=true&page_size=10' > /dev/null

```
~ 4.4 seconds with the main branch 
~ 0.14 seconds with this branch
```

- With a populated db and using a custom page=1, page_size=1000
time curl -vs 'http://127.0.0.1:8000/reconstruction-morphology?with_facets=true&page_size=1000' > /dev/null

```
~ 7.2-7.5 seconds with the main branch 
~ 0.5-0.6 seconds with this branch
```

Other:
- use join instead of "in" with hierarchy CTE
- fix typo in within_brain_region_hierachy_id (breaking change)
- use IdFilterMixin in BrainRegionFilter
- use SubjectFilterMixin instead of SpeciesFilterMixin in DensityFilterBase
- add species facets to IonChannelModel (it already allowed to filter by species)
- add BrainRegionFilterMixin to MEModelFilter (it already joined with BrainRegion but it didn't allow filtering)
- rename the facet `single_neuron_synaptome` to `synaptome` in SingleNeuronSynaptomeSimulation for consistency (synaptome is used in both the model and in the filter)
- Fix `species_id_in` -> `species_id__in` in `SpeciesFilterMixin`
- Fix type of `score__lte` and `score__gte` from int to float


app/service/single_neuron_synaptome.py
```diff
@@ -82,13 +82,20 @@ def read_many(
...
     )
-    apply_filter_query = lambda query: (
-        query.join(BrainRegion, SingleNeuronSynaptome.brain_region_id == BrainRegion.id)
-        .outerjoin(Contribution, SingleNeuronSynaptome.id == Contribution.entity_id)
-        .outerjoin(Agent, Contribution.agent_id == Agent.id)
-        .outerjoin(created_by_alias, SingleNeuronSynaptome.createdBy_id == created_by_alias.id)
-        .outerjoin(me_model_alias, SingleNeuronSynaptome.me_model_id == me_model_alias.id)
-    )
+    filter_joins = {
+        "brain_region": lambda q: q.join(
```
eleftherioszisis, line 86: I wonder, would it be possible instead of strings to be able to pass here `CustomFilter.brain_region` ? This is something I wanted to also check if possible with the multiple aliases per model dict (or change the aliases to a mapping like here, from filter name to (alias, model) or sth along these lines) so that we ensure that whatever is passed as key is always correct.

GianlucaFicarelli: I checked but `CustomFilter.brain_region` is an instance of CustomFilter/Filter that is a subclass of a Pydantic model, so it's not hashable by default.
Even setting `frozen=True` isn't enough, because some internal attributes are mutable (for example, `order_by` is a list), so it seems not possible unless dealing with internals of fastapi-filter.

app/service/emodel.py
```diff
@@ -110,7 +108,6 @@ def read_many(
             "createdBy": created_by_alias,
             "updatedBy": updated_by_alias,
         },
-        IonChannelModel: ion_channel_model_alias,
```
GianlucaFicarelli, line 113: I removed IonChannelModel because it's not defined in EModelFilter.
If we need it, we can keep IonChannelModel here, and add a nested IonChannelModelFilter to EModelFilter.

eleftherioszisis (approved):
Great work!

app/service/emodel.py
mgeplf: makes sense

mgeplf (approved):
LGTM:

One general question I have/though is that perhaps the `filter_joins` should be moved out of the functions, so the dictionary doesn't need to be recreated each time, including all the lambdas.

The other thought I had was maybe we could decrease the repatition by having a factory function that takes `factory(model, ("brain_regions", "contribution", etc))` and creates the dict?

GianlucaFicarelli:
> One general question I have/though is that perhaps the `filter_joins` should be moved out of the functions, so the dictionary doesn't need to be recreated each time, including all the lambdas.

Similarly to `name_to_facet_query_params`, also `filter_joins` quite often requires the aliases, so in those cases it cannot be moved outside. Alternatively, all the definitions of aliases, name_to_facet_query_params and filter_joins, can be moved into a separate function that returns name_to_facet_query_params and filter_joins, but it wouldn't be very beneficial because it cannot be reused.

> 
> The other thought I had was maybe we could decrease the repetition by having a factory function that takes `factory(model, ("brain_regions", "contribution", etc))` and creates the dict?

If I understand correctly, we would need to pass the aliases also in this case. I can give it a try, maybe in a separate PR?

mgeplf:
Ah, good point about the aliases; annoying.  I think you're onto something w/ the definitions all having to be created simulataneously.

Anyways, that can all be for a different PR.  I think this is good to go.

PR #179, opened 2025-05-20
Single cell simulation label

g-bar:


app/db/types.py
```diff
@@ -189,6 +189,7 @@ class AssetLabel(StrEnum):
     cell_composition_volumes = auto()
     single_neuron_synaptome_config = auto()
     single_neuron_synaptome_simulation_io_result = auto()
+    single_cell_simulation = auto()
```
eleftherioszisis, line 192: The asset label should define what kind of asset it is. This has the same name as the entity and isn't very clear as to what asset it corresponds to.

app/filters/single_neuron_simulation.py
```diff
@@ -19,6 +20,7 @@ class SingleNeuronSimulationFilter(
 ):
     status: SingleNeuronSimulationStatus | None = None
 
+    me_model_id: UUID | None = None
```
eleftherioszisis, line 23: Hmm, is this needed? Wouldn't me_model__id work?

app/schemas/asset.py
```diff
@@ -61,4 +61,4 @@ def ensure_entity_type_label_consistency(self):
 
 
 class AssetsMixin(BaseModel):
-    assets: list[AssetRead] | None
+    assets: list[AssetRead]
```
eleftherioszisis, line 64: Wouldn't there be a validation error if the entity is fetched before registering the assets?

g-bar: No, the uselist here https://github.com/openbraininstitute/entitycore/blob/0d79b1301d433665b0d7c9ad8f1ca13897e47357/app/db/model.py#L365

Ensures it's always defined (empty list if not assets).

app/filters/single_neuron_simulation.py
g-bar: There was no me_model__id, I can change it to that if preferable.

app/db/types.py
g-bar: Good point, the asset that is stores contains config, results and stimuli in same so wasn't sure what to name it.

app/schemas/asset.py
eleftherioszisis: Ah I see. ok then

app/db/types.py
eleftherioszisis: Yeah, it has everything in one sadly. Maybe single_cell_simulation_data for now and we separate in the future into config/result assets?

app/filters/single_neuron_simulation.py
g-bar: Changed it to me_model__id

app/db/types.py
g-bar: Sounds good for now.

I'm unsure about splitting actually the MEModel detail view has to download the assets for every simulation in display that uses that MEModel, so splitting would require 3x more requests.
But maybe there is a usecase that requires separate files, so let's cross that bridge if/when needed.

eleftherioszisis: Most workflow tasks in general differentiate between config and generated result. Mixing them together makes it hard to reason about the config being changed and the result being generated. However, in this particular case it doesn't really matter as the tool is handling that resource.

eleftherioszisis (approved):

g-bar:
Thanks @eleftherioszisis

PR #182, opened 2025-05-20
Circuit schema implementation

chr-pok:
Added a first draft of a database model for representing (tiny) circuits. Assumes that `ScientificArtifact` already exists as a base class for `Circuit`. Added a simple "Read One" endpoint as well.

Related issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/10

Still work in progress!

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -42,6 +42,7 @@ from app.db.types import (
     AssetStatus,
     CircuitBuildCategory,
     CircuitScale,
+    DerivationType,
     ElectricalRecordingOrigin,
     ElectricalRecordingStimulusShape,
     ElectricalRecordingStimulusType,
@@ -55,7 +56,6 @@ from app.db.types import (
     SingleNeuronSimulationStatus,
     StructuralDomain,
     ValidationStatus,
-    DerivationType,
 )
 from app.schemas.publication import Author, PublicationType
 from app.utils.uuid import create_uuid


chr-pok:
This is just a first draft with many elements still missing. However, some of them are not required for tiny circuits which is our primary goal so they can come later. Even though it is still work in progress, it would be great to get some feedback already!

chr-pok:
Some specific questions I had (@jdcourcol and others):
- How to deal with "connectivity_matrices" which may contain multiple .h5 files. Add them individually as different assets (.h5 file each), or as one asset (folder with all .h5 files), or as separate entities in the DB each of which has an asset (.h5 file), or something else?
- Similarly, how to deal with "overview_figures" which may contain multiple .png files which should be used later on on the detailed explore page.
- "is_simulatable": We have it defined in "General_schema.docx" but now I think this sounds more like a validation tag perhaps?
- More scientific: So far, I have defined 3 categories for "CircuitBuildCategory", to be further discussed/refined:
  - data_driven_model (like SSCx)
  - statistical_model (like simplified circuits)
  - em_reconstruction (reconstructed from EM)

app/db/types.py
```diff
@@ -207,4 +209,22 @@ class AssetLabel(StrEnum):
...
+
+class CircuitBuildCategory(StrEnum):
+    data_driven_model = auto()
+    statistical_model = auto()
+    em_reconstruction = auto()
+    # TODO: Add & refine categories
+
+
+class CircuitScale(StrEnum):
+    single = auto()
```
jdcourcol, line 224: do you have any definition how we distinguish these scale ?

app/db/model.py
```diff
@@ -865,3 +868,61 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+
+    has_morphologies: Mapped[bool] = mapped_column()
+    has_point_neurons: Mapped[bool] = mapped_column()
+    has_electrical_cell_models: Mapped[bool] = mapped_column()
+    has_spines: Mapped[bool] = mapped_column()
+    # is_simulatable: Mapped[bool] = mapped_column()  # FIXME: Could be more like a validation (?)
+
+    version: Mapped[str] = mapped_column(default="")
+
+    number_neuron: Mapped[int] = mapped_column()
```
jdcourcol, line 912: number_neuron -> number_neurons

app/db/model.py
```diff
@@ -865,3 +868,61 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    build_category: Mapped[CircuitBuildCategory]
+    scale: Mapped[CircuitScale]
+
+    has_morphologies: Mapped[bool] = mapped_column()
+    has_point_neurons: Mapped[bool] = mapped_column()
+    has_electrical_cell_models: Mapped[bool] = mapped_column()
+    has_spines: Mapped[bool] = mapped_column()
+    # is_simulatable: Mapped[bool] = mapped_column()  # FIXME: Could be more like a validation (?)
+
+    version: Mapped[str] = mapped_column(default="")
```
jdcourcol, line 910: this one is tough. I think we have to define how to different versions are linked together.

jdcourcol:
> * "is_simulatable": We have it defined in "General_schema.docx" but now I think this sounds more like a validation tag perhaps?

It sounds like an annotation.

app/db/types.py
chr-pok: We had definitions as follows in mind:
- `single`: Single neuron + extrinsic connectivity
- `pair`: Two neurons + intrinsic connectivity + extrinsic connectivity
- `tiny`: Small number of neurons + intrinsic connectivity + extrinsic connectivity. "Small" meaning small enough to be simulated with BlueCelluLab. Perhaps in the order of 3-100 neurons. Still thinking about a better term, like "motif circuit" or similar (since we are aiming for specific connectivity motifs, rather than isolated neurons).
- `microcircuit`: Single cortical column
- `region`: Atlas-based continuous volume of a brain region or sub-region(s)
- `system`: Non-continuous circuit consisting of at least two microcircuits/regions that are connected by inter-region connectivity
- `whole_brain`: Circuit representing an entire brain

In general, we don't want to put neuron numbers for distinguishing these scales since they may be very different depending on the species for example.

app/db/model.py
chr-pok: Thanks, fixed!

app/db/model.py
chr-pok: What we had in mind is not so much a strict relations between different versions of a circuit, since I imagine this would be really complex indeed.

The idea is more to have a field where the user could just put an (optional) string to describe the build or release version of a circuit. Like for example "v5", "v6", or "v7" for BBP circuits, but could be a different notation depending on the lab as well. Not sure if it is a good idea to allow an arbitrary string though?

app/db/model.py
```diff
@@ -865,3 +868,61 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    # - description: ...
+    # - brain_region: ...
+    # - specific_brain_regions: ...
+    # - ...
+
+    # Asset(s):
+    # - sonata_circuit: Folder containing SONATA circuit
+    # - ...
+
+    parent_circuit_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("circuit.id"), index=True, nullable=True, default=None)
```
eleftherioszisis, line 895: Parent circuit? Is this a derivation where this circuit is derived from another one?

app/db/types.py
```diff
@@ -207,4 +209,22 @@ class AssetLabel(StrEnum):
...
     },
     EntityType.single_neuron_simulation: {AssetLabel.single_cell_simulation_data},
+    EntityType.circuit: {AssetLabel.sonata_circuit},
 }
+
+
+class CircuitBuildCategory(StrEnum):
+    data_driven_model = auto()
+    statistical_model = auto()
+    em_reconstruction = auto()
```
eleftherioszisis, line 219: It's not clear what these mean as circuit build categories.

app/db/model.py
```diff
@@ -865,3 +868,61 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    has_point_neurons: Mapped[bool] = mapped_column()
+    has_electrical_cell_models: Mapped[bool] = mapped_column()
+    has_spines: Mapped[bool] = mapped_column()
+    # is_simulatable: Mapped[bool] = mapped_column()  # FIXME: Could be more like a validation (?)
+
+    version: Mapped[str] = mapped_column(default="")
+
+    number_neurons: Mapped[int] = mapped_column()
+    number_connections: Mapped[int] = mapped_column()
+    number_synapses: Mapped[int] = mapped_column()
```
eleftherioszisis, line 914: These look like measurements that should be optional. Are they summed over all node and edge populations in the circuit? What if the circuit has other types of nodes apart from neurons and other types of edges apart from chemical ones?

app/db/types.py
chr-pok: > It seems that in a single enum that implies only size we mix size, embedding, and connectivity type. Isn't it confusing?

It is not really about size, more about the general structure of a circuit on a conceptual level which represents the use cases and types of experiments that can be conducted with such circuits. And this is also reflected in the way the platform displays the different build/simulate options at the moment (except tiny circuit, but we made a proposal for adding it). So the idea was to make it less confusing but easily distinguishable actually. But the exact terms can still be refined of course...

app/db/model.py
chr-pok: When extracting sub-circuits from a given circuit, we would like to keep track of the parent circuit a sub-circuit has been extracted from. Like the ones we have on the platform right now: nbS1-O1 <- nbS1-HEX0 <- nbS1-HEX0-L1
So each circuit may have no or exactly one parent circuit. I think this is not the same as "derived from", is it?

app/db/types.py
chr-pok: It should provide a means to distinguish circuits based on the method/principle it has been built upon. This is still under discussion...

app/db/model.py
dkeller9: It would be nice if the script and parameters used to extract a circuit artifact could be tracked somehow in provenance, but maybe that is outside the actual artifact.

eleftherioszisis: It does sound exactly like a derivation process. Concerning the provenance this could fit in an activity/entity pattern:
https://www.w3.org/TR/prov-o/#starting-points-figure

where the `Activity` is the extraction process with a `used` config storing all the parameters and a `generated` which is the sub-circuit artifact. The sub-circuit entity in turn would have a `wasDerivedBy` pointing to the parent circuit entity and a `wasGeneratedBy` pointing to the extraction process activity.

app/db/model.py
eleftherioszisis: `number_of_connections` is generic as it doesn't describe the type of connection. Number of neurons/synapses on the other hand is specific to neuronal circuits only. Maybe keeping number_of_connections/number_of_cells would be more meaningful?

Have you tried any benchmarks how costly it is to calculate these numbers as a function of the circuit size? Is it a matter of just checking the populations sizes, or does it also require filtering which would be very costly for circuits with billions+ connections.

app/db/types.py
eleftherioszisis: I can say with certainly that all those methods are combined for a circuit to be built :D

app/db/model.py
chr-pok: > Why do you need to do that to count the connections?

Not saying we need to do it like this. But if we have the connectivity matrices, we can get this count very easily. And we need these matrices anyway for doing lots of other statistics/plots.

> Can't it be as simple as getting the size of the edge populations?

For synapse counts yes, but not for connection counts which would require a `unique` operation on the source/target node pairs (unless I am missing something).

app/db/model.py
jdcourcol: To switch circuits to entity core, we don't need the provenance (i.e used + activity + generated), we can tackle that later. 
We can indeed model the link between parent and sub circuit by a derivation link.

app/db/types.py
jdcourcol: > We had definitions as follows in mind:
> 
>     * `single`: Single neuron + extrinsic connectivity
> 
>     * `pair`: Two neurons + intrinsic connectivity + extrinsic connectivity
> 
>     * `tiny`: Small number of neurons + intrinsic connectivity + extrinsic connectivity. "Small" meaning small enough to be simulated with BlueCelluLab. Perhaps in the order of 3-100 neurons. Still thinking about a better term, like "motif circuit" or similar (since we are aiming for specific connectivity motifs, rather than isolated neurons).
> 
>     * `microcircuit`: Single cortical column
> 
>     * `region`: Atlas-based continuous volume of a brain region or sub-region(s)
> 
>     * `system`: Non-continuous circuit consisting of at least two microcircuits/regions that are connected by inter-region connectivity
> 
>     * `whole_brain`: Circuit representing an entire brain
> 
> 
> In general, we don't want to put neuron numbers for distinguishing these scales since they may be very different depending on the species for example.

Thank you, I would put that next to this class.

app/db/model.py
james-isbister: Other properties that could be useful to make immediately accessible (so that the nodes/edges table don't need to be loaded to find them) could be the:
- The list of populations
- The column names of the nodes and edge tables, and the unique values for any categorical column. 
- The list of node_sets
- The outer x,y,z bounds of the circuit

All of these are important for the GUI. For example:
- Stimulate neurons with based on a property filter
- Stimulate a particular pre-existing node_set (these could be read from the node_sets file, so not essential)

eleftherioszisis: Then these numbers should be a measurement annotation that is not by default attached to the circuit.

When a circuit is built it will not generate connectivity matrices in the same task as the building task.

It should be decoupled code that runs on a dedicated task and annotates the circuit with the matrices/numbers.

Not all circuits that are going to be generated are going to be used. It will be wasteful to attach that overhead to each building task.

In addition, each task may have a different distributed mechanism/configuration, which may not be compatible with the connectivity generation task.

In general, computation intensive tasks should be independent from each other.

app/db/types.py
james-isbister: Perhaps it is good to categorize the anatomical and physiological categories of the model separately. 

For example:
AnatomicalBuildCategory: 
- fully_synthesized_original
- full_em_reconstruction_original
- placed_morphologies_original

For each of these there could be option with an additional suffix _connectivity_modification, such as placed_morphologies_connectivity_modification for circuits which modify the connectivity based on an previous initial (i.e. original) wiring.

It may even make sense to have a separate AnatomicalConnectivityBuildCategory as well

app/db/model.py
james-isbister: Given our discussions about how we will represent the node remapping between root circuits and subcircuits, I think it make sense to explicitly represent the root circuit i.e. nbS1-O1, rather than having to recursively figure this out.

Although we definitely want to display extracted circuits as children of a parent circuit, and grandchildren of a root circuit, the display of this information could get a little tricky if people make extractions directly from a grandparent, but really its conceptually a child of particular parent. 

I'm not sure what the solution would be here, and maybe we just have to do it based on extractions

app/routers/circuit.py
dkeller9, line 1: Do you think people would want to post a circuit from external sources, or would they go through our build chain?

app/db/model.py
jdcourcol: If I understand correctly, what you say is that we should have multiple relationships , root_circuit + direct_parent, to avoid doing recursion ?

app/db/model.py
danielaegassan: Also extracting connectivity matrices is a costly operation for large circuits.  However, many connectivity analyses are more efficient or impossible without a connectivity matrix.  This is why, in my opinion, it is a good idea to generically extract the connectivity matrix of a built circuit.  But it could a a separate step of course.

danielaegassan: Alternatively, @chr-pok I don't see where in the schema we are saying what is the "scale" of the circuit.  It would be good to know if it is: 

- tiny (up to 10/15 neurons)
- micro (larger with a single local node population)
- system (more than one local node population)

No?

app/db/types.py
danielaegassan: @chr-pok Wasn't tiny going to be up to 10-15 neurons?  At least I am assuming this for the explore design ...

danielaegassan: Also I find region vs microcircuit in this list potentially confusing, we can discuss more on Thursday.

chr-pok: > @chr-pok Wasn't tiny going to be up to 10-15 neurons? At least I am assuming this for the explore design ...

For CNS yes, but not in general. But we can talk more tomorrow...

app/db/model.py
eleftherioszisis: Yeah, I am not against calculating it, just as a separate step. In addition, for just the number of connections, if the edge list is sorted by source/target it should be possible in principle to calculate the number of connections in O(n) and without having to store the entire adjacency in memory. Not sure though if we always sort the edges in the edge populations.

danielaegassan: If we add matrices for all circuits, as we have done for the one currently in explore, these could obtained from that.  The matrices can be computed as a separate (mandatory?) step.

app/routers/circuit.py
danielaegassan: I imagine we also want to allow people adding external circuits.

app/db/types.py
danielaegassan: My suggestion for this is to keep the name CircuitBuildCategory to be consistent with morphologies.  The logic is that this describes the input to build this entity.  I suggest the following categories:

- data_driven_detailed_model (like SSCx) 
- data_driven_point_neuron_model  
- em_reconstruction (reconstructed from EM)

danielaegassan: works for me!

app/db/types.py
mgeplf: > (otherwise it will break the visualization pipeline for such type of circuits).

I would caution against building schemas that depend on external factors like this; I would assume that the visualization pipeline will improve over time, and allow for bigger circuits.  By encoding these sorts of assumptions into the schema, they will live forever.

danielaegassan: @mgeplf 

The visualization pipeline is about what we show about this circuit.  It is super different to interactively visualize 10 neurons and choose one of them by clicking that do this on 100 neurons. A different design of UI will be needed.

Moreover, it is also about what do you show for circuits of different sizes, in particular in Explore.

If circuits have hundreds of thousands of neurons, you can do network statistics on them and use them as defining features of your circuit (see for example what we have in explore now).  However, these metrics, would not make sense for a circuit of 10 neurons, one has to show something different for those. 

So making a size cutoff makes sense.  It is an arbitrary number though.  Is there a principled way to choose it?

james-isbister:
> > @GianlucaFicarelli @mgeplf , do you have any opinion on this ?
> 
> I missed this before; I would like to see the code that generates a connectivity matrix, and see how hard it is to create on the fly from a SONATA file - having duplicates nearly always leads to problems, in my experience.

@mgeplf 
https://github.com/openbraininstitute/obi-one/blob/main/obi_one/scientific/connectivity_matrix_extraction/connectivity_matrix_extraction.py

danielaegassan:
@mgeplf you can see the extraction code here 

https://github.com/openbraininstitute/obi-one/tree/main/obi_one/scientific/connectivity_matrix_extraction

I don't know if we would do it on the fly each time, because it can be expensive, but rather do it once every time upon the generation of a circuit.

mgeplf:
@danielaegassan > @mgeplf you can see the extraction code here

Thanks, I will have a look when I have time.

app/db/types.py
mgeplf: > So making a size cutoff makes sense. It is an arbitrary number though. Is there a principled way to choose it?

I would say that a choice would be to not have types for the circuit size that rely on cutoffs:
I see there is a `number_neurons`; can't that be used instead?  That way whichever tool is accessing/presenting the information, it can decide what falls within its purview, and there doesn't need to be a concept that is a moving target over time.

jdcourcol: I asked chatgpt:
grafik

on a different note: `Single cortical column` is very cortex centric + column is not widely accepted concept iirc (I had to rename  nmc portal to neocortical microcircuitry because of that )

chr-pok: Since we treat everything with connectivity as being a circuit, I think it is important to label them as it will be very different how to build them, visualize them, what simulator to use for them (single, pair, motif = BlueCelluLab; larger = Neurodamus), and what kinds of experiments a user will be able to do with them.

Using a cutoff based on number_of_neurons would lead to more confusion I believe, since it may highly depend on the species, e.g., a full brain of C.elegans should be always labeled as whole_brain even though it has only 302 neurons. There may be edge cases as always, so having not a strict cutoff will allow some flexibility.

app/db/model.py
chr-pok: So in the end we would need "parent_circuit", "root_circuit", and "was_derived_from", because they all mean different things. I added them to the schema, but I am wondering if there is a reference implementation or example I can follow for implementing especially the `wasDerivedFrom` relation mentioned by @eleftherioszisis since this seems to be well-defined?

PR #183, opened 2025-05-21
Fix missing ion channel models in memodel's emodel

eleftherioszisis:
Use expanded EModel schema everywhere

Fixes https://github.com/openbraininstitute/entitysdk/issues/49

### DIFF
--- a/app/schemas/emodel.py
+++ b/app/schemas/emodel.py
@@ -56,7 +56,4 @@ class EModelRead(
     mtypes: list[MTypeClassRead] | None
     etypes: list[ETypeClassRead] | None
     exemplar_morphology: ExemplarMorphology
-
-
-class EModelReadExpanded(EModelRead, AssetsMixin):
     ion_channel_models: list[IonChannelModelWAssets]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -444,27 +444,27 @@ def emodel_id(create_emodel_ids: CreateIds) -> str:
 
 @pytest.fixture
 def create_memodel_ids(
-    db, morphology_id, brain_region_id, species_id, strain_id, emodel_id, agents
+    db, client, morphology_id, brain_region_id, species_id, strain_id, emodel_id, agents
 ) -> CreateIds:
     def _create_memodel_ids(count: int):
         memodel_ids: list[str] = []
         for i in range(count):
-            memodel_id = add_db(
-                db,
-                MEModel(
-                    name=f"{i}",
-                    description=f"{i}_description",
-                    brain_region_id=brain_region_id,
-                    species_id=species_id,
-                    strain_id=strain_id,
-                    morphology_id=morphology_id,
-                    emodel_id=emodel_id,
-                    authorized_public=False,
-                    authorized_project_id=PROJECT_ID,
-                    holding_current=0,
-                    threshold_current=0,
-                ),
-            ).id
+            memodel_id = assert_request(
+                client.post,
+                url="/memodel",
+                json={
+                    "name": f"{i}",
+                    "description": f"{i}_description",
+                    "brain_region_id": str(brain_region_id),
+                    "species_id": str(species_id),
+                    "strain_id": str(strain_id),
+                    "morphology_id": str(morphology_id),
+                    "emodel_id": str(emodel_id),
+                    "authorized_public": False,
+                    "holding_current": 0,
+                    "threshold_current": 0,
+                },
+            ).json()["id"]
 
             add_contributions(db, agents, memodel_id)
 
--- a/app/service/emodel.py
+++ b/app/service/emodel.py
@@ -28,7 +28,7 @@ from app.dependencies.common import (
 from app.dependencies.db import SessionDep
 from app.filters.emodel import EModelFilterDep
 from app.queries.common import router_create_one, router_read_many, router_read_one
-from app.schemas.emodel import EModelCreate, EModelRead, EModelReadExpanded
+from app.schemas.emodel import EModelCreate, EModelRead
 from app.schemas.types import ListResponse
 
 if TYPE_CHECKING:
@@ -61,13 +61,13 @@ def read_one(
     user_context: UserContextDep,
     db: SessionDep,
     id_: uuid.UUID,
-) -> EModelReadExpanded:
+) -> EModelRead:
     return router_read_one(
         id_=id_,
         db=db,
         db_model_class=EModel,
         authorized_project_id=user_context.project_id,
-        response_schema_class=EModelReadExpanded,
+        response_schema_class=EModelRead,
         apply_operations=_load,
     )
 
@@ -96,7 +96,7 @@ def read_many(
     with_search: SearchDep,
     facets: FacetsDep,
     in_brain_region: InBrainRegionDep,
-) -> ListResponse[EModelReadExpanded]:
+) -> ListResponse[EModelRead]:
     morphology_alias = aliased(ReconstructionMorphology, flat=True)
     agent_alias = aliased(Agent, flat=True)
     created_by_alias = aliased(Agent, flat=True)
@@ -170,7 +170,7 @@ def read_many(
         apply_filter_query_operations=None,
         apply_data_query_operations=_load,
         pagination_request=pagination_request,
-        response_schema_class=EModelReadExpanded,
+        response_schema_class=EModelRead,
         name_to_facet_query_params=name_to_facet_query_params,
         filter_model=emodel_filter,
         filter_joins=filter_joins,
--- a/app/service/memodel.py
+++ b/app/service/memodel.py
@@ -57,6 +57,7 @@ def _load(select: Select):
             joinedload(EModel.createdBy),
             joinedload(EModel.updatedBy),
             selectinload(EModel.assets),
+            selectinload(EModel.ion_channel_models),
         ),
         joinedload(MEModel.morphology).options(
             joinedload(ReconstructionMorphology.brain_region),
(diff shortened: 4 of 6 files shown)


tests/conftest.py
```diff
@@ -444,27 +444,27 @@ def emodel_id(create_emodel_ids: CreateIds) -> str:
...
+                    "brain_region_id": str(brain_region_id),
+                    "species_id": str(species_id),
+                    "strain_id": str(strain_id),
+                    "morphology_id": str(morphology_id),
+                    "emodel_id": str(emodel_id),
+                    "authorized_public": False,
+                    "holding_current": 0,
+                    "threshold_current": 0,
+                },
+            ).json()["id"]
```
GianlucaFicarelli, line 467: What is the reason to execute a call instead of adding the record to the db?
My main concern is that this pattern could be more expensive, and slow down the tests if used in several tests and loops, but it should be measured to see if it's negligible or not.

app/schemas/emodel.py
```diff
@@ -56,7 +56,4 @@ class EModelRead(
     mtypes: list[MTypeClassRead] | None
     etypes: list[ETypeClassRead] | None
     exemplar_morphology: ExemplarMorphology
-
-
-class EModelReadExpanded(EModelRead, AssetsMixin):
     ion_channel_models: list[IonChannelModelWAssets]
```
GianlucaFicarelli, line 59: With this change, in the get_many endpoint we return all the ion_channel_models, with all the assets (that are not paginated because they are nested entities).
Do we have an idea of the numbers of nested entities, and how it affects the response time and size?
Should we consider reducing the default number of results, that's now set to 100 ?

eleftherioszisis: @AurelienJaquier could you please provide some insight here?

tests/conftest.py
eleftherioszisis: The main reason is registering the createdBy/updatedBy that I then check on the tests. We can assume it works but I am more reassured with the explicit checks.

app/schemas/emodel.py
AurelienJaquier: Each ion channel model should have only one asset: the mod file.
The e-models we currently have on the platform each have a small number of ion channel models (around 10). In the future, we will probably have e-models with genetic ion channals. Those e-models will have a larger number of ion channel models (>30 if I recall correctly).

g-bar: @AurelienJaquier  I don't understand why you need the ion channel models in the list view of MEModel, do you do some sort of bulk processing of memodels that need the ion channel models for all.

Otherwise you can just do one extra call to get the emodel from the me.model.emodel.id?

g-bar: Maybe for validation?

eleftherioszisis: So should memodel's emodel not include assets and channels? Not that this will be consistent among all endpoints of MEModel

g-bar: Assets are used for the frontend in list view (to display the emodel thumbnail).

If @AurelienJaquier is ok with doing an extra call then ion_channel_models not needed.

(Ideally we have the ion_channel_models in the read_one endpont only but I understand we want to keep the response consistent between read_one, read_many. So this is the best compromise betweem consistency and performance IMHO).

eleftherioszisis: I am not fond of drawing lines on what to be present and whatnot based on favoring specific use cases...

Each schema represents an entity and should always be the same or in the case that we don't want to be included it full we include a "view" of it that is always the same view, ideally an id and a type.

However, that ship of having linked entities as views has sailed long time ago. :D

eleftherioszisis: How would entitysdk know that a schema is not full though? This is why I mentioned views as a specific type that will be associated with a type and then be lazily loaded.

eleftherioszisis: We could have the complete schemas defined and select what needs to be returned with graphql. Everyone's happy and no hidden omissions. :)

GianlucaFicarelli: I'm ok with the change but I'd like to see the resulting response time and size, so we can decide if want to keep or change the default page_size=100.
Btw is the frontend passing a custom page_size? Can it be set differently by the user?

g-bar: The frontend uses page_size=30 not customizable by user.

GianlucaFicarelli: Note that this approach probably cannot be applied to morphologies and measurements, because adding measurements to all the morphologies returned by the get_many endpoint could increase the payload size considerably.

To use the same model in get_one and get_many endpoints, an alternative could be to remove EModelReadExpanded/ReconstructionMorphologyAnnotationExpandedRead, and expose the related resources in a different endpoint:

- `GET /emodel/<uuid>/ion-channel-models`
- `GET /reconstruction-morphology/<uuid>/measurement-annotation`

That's quite usual in REST endpoints, but there are some downsides:
- it requires an additional call to get the list of resources
- the client needs to handle the different logic
- the model might not map well to graphql if we want to support it.

g-bar: That would be fine by me.

But what  I find unsual is the necessity of making read_one and read_many have the same schema, that's what's unusual in rest apis.

Normaly list view (read_many) returns a summary of the resource (incomplete schema), while the detail view (read_one) full schema. Nested fields are to be avoided in the list view, except where that leads to N+1 requests.

AurelienJaquier: I am also ok with that

bilalesi: I am fine with it too

bilalesi: need a change but it's small thing

eleftherioszisis: Sadly, this was too good to be true as there is not direct emodel_id in IonChannelModel. We would need to make a mess with using the associative table to make that work.

I will consider @GianlucaFicarelli proposal:
```
GET /emodel/<uuid>/ion-channel-models
GET /reconstruction-morphology/<uuid>/measurement-annotation
```

This is something that we will also need to address in the SDKs for the brain-atlas anyway.

GianlucaFicarelli: I'm working on a similar case (adding a derivation that returns the traces used to generate emodel), and I tried to allow searching `ElectricalCellRecording` by `generated_emodel__id__in` with the approach suggested by Gil.

To make it work I didn't need to add `generated_emodel` to the `ElectricalCellRecording` model, but I needed to add it to the `ElectricalCellRecordingFilter` as in
```python
    generated_emodel: Annotated[
        EModelFilter | None, FilterDepends(with_prefix("generated_emodel", EModelFilter))
    ] = None
```
and add to filter_joins something like:
```python
        "generated_emodel": lambda q: q.join(
            Derivation, Derivation.used_id == db_model_class.id
        ).join(emodel_alias, Derivation.generated_id == emodel_alias.id),
```
In this way I don't need to define a relationship, although maybe it can be still useful in some cases.

However, I'm not fully convinced of the approach, because the relationship between EModel and ElectricalCellRecording (or between EModel and IonChannelModel) is not explicit, since it's not mentioned in the schema, nor in the openapi endpoints (it's just a possible parameter for filtering the get_many endpoint of ElectricalCellRecording/IonChannelModel).

g-bar: Couldn’t this work?

The backend continues to differentiate between Expanded and not expanded schemas. 

When sdk users fetch an MEModel the EModel field is incomplete. 

On the entitysdk side MEModel.emodel is a property that when accessed by the user triggers a call to /emodels/id and caches it / returns it to the user.

Like this we don’t compromise performance, and users get the full schemas on entitysdk.

eleftherioszisis: The sdk does not handle the entities one by one the same way the backend is exposing different routers. Therefore,  we need a way of annotating an entity as incomplete if we want to go that way.

On top of that, I would prefer to keep service and data model decoupled, therefore I am not very fond of lazy loading especially when the backend does not support views/links consistently. With nexus all linked entities had to be fetched, therefore that was a good candidate for lazy loading. Here, we only want to do it for arbitrary properties that we decide we don't want to expand for each endpoint.

Concerning the expanded/non-expanded, unless hardcoded on the sdk side (which bring us back to the sdk not treating any route in a special way at the moment), the sdk does not know what should it expand and with which parameters. Maybe we could have a single expand parameter that covers everything, instead of having multiple expanding options that are different for each endpoint? I could by default pass that and receive the full data.

g-bar: Back to pragmatism, we could handle the MEModel.emodel lazily which seems particularly problematic for now. And decide later on a more general solution if / when the need arises?

eleftherioszisis: Or just ignore it for now. I am not eager rewriting the sdk for a single thing that needs to be fixed.

GianlucaFicarelli: I would avoid lazy implicit loading in entitysdk because:
- it has side effects that are hidden to the user
- it doesn't fit well with the async version (that's not available but it would be needed for integration in async services), since you cannot await on an attribute

Having the single expanded parameter could work, to avoid adding the complexity of multiple expanded models, although at the moment the parameters is available only in the get_one endpoint.

g-bar: > it has side effects that are hidden to the user

The entirety of entitysdk is predicated on hiding the complexity of a rest api from end users, otherwise we could do away with it and everyone just does calls directly to the api, furthermore it can be documented that it does an extra call.

> it doesn't fit well with the async version

@properties can be made awaitable if they return a coroutine.

> single expanded parameter could work

Yeah I agree.

Ultimately for now I think we should ignore it we're splitting hairs over some nested data that's not even needed by anyone.

Maybe for now emit a warning when users access MEModel.emodel.ion_channel_models to made them aware that they need to make an extra call? to get the full emodel? @eleftherioszisis

PR #186, opened 2025-05-21
Add CalibrationResult model

AurelienJaquier:
I made changes in a similar way to PR #171

CalibrationResult schema is still in discussion with scientific team, so this is subject to change.


jdcourcol:
@AurelienJaquier I did not notice you created one.

I made one [here](https://github.com/openbraininstitute/entitycore/pull/188) 

The difference is that I want to enforce that specific properties are present. Otherwise it is a bit "free form"

PR #191, opened 2025-05-22
Add validation results in MEModelRead

AurelienJaquier:
I made it so the ValidationResult entities that point to an MEModel show up in the MEModelRead.
I am not sure if the code is good quality, but I tested it locally and it was working

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -303,6 +303,73 @@ class ETypesMixin:
         )
 
 
+class Entity(LegacyMixin, Identifiable):
+    __tablename__ = "entity"
+
+    type: Mapped[EntityType]
+    annotations = relationship("Annotation", back_populates="entity")
+
+    # TODO: keep the _ ? put on agent ?
+    created_by = relationship("Agent", uselist=False, foreign_keys="Entity.created_by_id")
+    # TODO: move to mandatory
+    created_by_id: Mapped[uuid.UUID | None] = mapped_column(ForeignKey("agent.id"), index=True)
+    updated_by = relationship("Agent", uselist=False, foreign_keys="Entity.updated_by_id")
+    # TODO: move to mandatory
+    updated_by_id: Mapped[uuid.UUID | None] = mapped_column(ForeignKey("agent.id"), index=True)
+
+    authorized_project_id: Mapped[uuid.UUID]
+    authorized_public: Mapped[bool] = mapped_column(default=False)
+
+    contributions: Mapped[list["Contribution"]] = relationship(uselist=True, viewonly=True)
+    assets: Mapped[list["Asset"]] = relationship(
+        "Asset",
+        foreign_keys="Asset.entity_id",
+        uselist=True,
+        viewonly=True,
+    )
+
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+        "polymorphic_on": "type",
+    }
+
+
+class ValidationResult(Entity):
+    __tablename__ = EntityType.validation_result.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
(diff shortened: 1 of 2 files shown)


app/db/model.py
```diff
@@ -303,6 +303,74 @@ def etypes(cls) -> Mapped[list["ETypeClass"]]:
         )
 
 
+class Entity(LegacyMixin, Identifiable):
```
mgeplf, line 306: it would be helpful if the `class Entity` wasn't moved around, so the diff is smaller, and it makes it easier to see the changes, could you move it back?

AurelienJaquier: done, I moved it back

app/schemas/me_model.py
```diff
@@ -59,3 +60,4 @@ class MEModelRead(
     etypes: list[ETypeClassRead] | None
     morphology: ReconstructionMorphologyRead
     emodel: EModelRead
+    validation_results: list[ValidationResultRead] | None = None
```
g-bar, line 63: There is a discussion ongoing on the other PR on how to handle nested resources:

https://github.com/openbraininstitute/entitycore/pull/183

Depending on what's decided there we would potentially remove this field and instead have a 
/validation_results endpoint with a memodel_id__in filter.

g-bar: I think it should be fine to add it to the list view, there are only a few validations per me_model and they only contain 1 asset is that right @AurelienJaquier ?

What do you think @eleftherioszisis ?

g-bar, line 63: validation_results: list[ValidationResultRead] | None = None

`None` is not a possible value, if there are no validation results sqlalchemy will return an empty list.

eleftherioszisis: Sure, fine by me!

g-bar: Actually @AurelienJaquier 

The response is already quite hefty, look at all those joins.
And we already have a /validation_results enpoint which is filterable by `validated_entity_id` that's all the frontend needs.

Unless you need the validation results in batch, I suggest we use the endpoint above instead of nesting the data here.

```
 return select.options(
        joinedload(MEModel.species),
        joinedload(MEModel.strain),
        joinedload(MEModel.emodel).options(
            joinedload(EModel.species),
            joinedload(EModel.strain),
            joinedload(EModel.exemplar_morphology),
            joinedload(EModel.brain_region),
            selectinload(EModel.contributions).joinedload(Contribution.agent),
            selectinload(EModel.contributions).joinedload(Contribution.role),
            joinedload(EModel.mtypes),
            joinedload(EModel.etypes),
            joinedload(EModel.created_by),
            joinedload(EModel.updated_by),
            selectinload(EModel.assets),
        ),
        joinedload(MEModel.morphology).options(
            joinedload(ReconstructionMorphology.brain_region),
            selectinload(ReconstructionMorphology.contributions).selectinload(Contribution.agent),
            selectinload(ReconstructionMorphology.contributions).selectinload(Contribution.role),
            joinedload(ReconstructionMorphology.mtypes),
            joinedload(ReconstructionMorphology.license),
            joinedload(ReconstructionMorphology.species),
            joinedload(ReconstructionMorphology.strain),
            joinedload(ReconstructionMorphology.created_by),
            joinedload(ReconstructionMorphology.updated_by),
            selectinload(ReconstructionMorphology.assets),
        ),
        joinedload(MEModel.brain_region),
        selectinload(MEModel.contributions).joinedload(Contribution.agent),
        selectinload(MEModel.contributions).joinedload(Contribution.role),
        joinedload(MEModel.mtypes),
        joinedload(MEModel.etypes),
        joinedload(MEModel.created_by),
        joinedload(MEModel.updated_by),
        selectinload(MEModel.validation_results),
```

AurelienJaquier: Ok, then I will close this PR, and just use entitysdk to search directly `ValidationResult` with filtering by `validated_entity_id`.

PR #195, opened 2025-05-23
Initial implementation of `brain-atlas` endpoints

mgeplf:
See https://github.com/openbraininstitute/entitycore/pull/168 for the specification.

Briefly; a `BrainAtlas` is an named entity that has an asset `annotation.nrrd` attached to it. It is associated with a particular `BrainRegionHierarchy`

In addition, `BrainAtlasRegion` gives metadata for all the regions within a `BrainAtlas`;
* their volume in the `annotation.nrrd` if they are a leaf, -1 otherwise.
* an attached asset with the .obj mesh


app/db/model.py
```diff
@@ -875,9 +869,30 @@ class METypeDensity(
...
+
+class BrainAtlasRegion(Entity, LocationMixin):
+    __tablename__ = EntityType.brain_atlas_region
+
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+
+    # only the volume for leaf nodes is saved; the consumer must calculate
+    # volumes depending on which view of the hierarchy they are using
+    volume: Mapped[float] = mapped_column(nullable=True)
+    leaf_region: Mapped[bool] = mapped_column(default=False)
```
eleftherioszisis, line 892: imho `is_leaf_region` would make it.a little more clear it's a flag.

app/db/model.py
```diff
@@ -875,9 +869,30 @@ class METypeDensity(
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
 
 
-class BrainAtlas(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, Entity):
+class BrainAtlas(NameDescriptionVectorMixin, SpeciesMixin, Entity):
```
eleftherioszisis, line 872: Do we need a brain region for the atlas too? It could be the case that we have atlases for specific regions, otherwise it could be Root for the whole brain.

tests/test_brain_atlas.py
```diff
@@ -0,0 +1,162 @@
...
+    assert response.status_code == 200
+    assert response.json()["data"] == [
+        {
+            "brain_atlas_id": str(brain_atlas.id),
+            "brain_region_id": str(regions["root"].id),
+            "creation_date": ANY,
+            "id": str(ids["root"].id),
+            "leaf_region": False,
+            "update_date": ANY,
+            "volume": -1.0,
```
eleftherioszisis, line 121: Why not 0.0? It would allow to easily sum over all regions without a check and in principle no real region can have 0.0 volume.

app/service/brain_atlas.py
```diff
@@ -0,0 +1,72 @@
...
+        name_to_facet_query_params=None,
+        filter_model=brain_atlas_filter,
+    )
+
+
+def read_one(id_: uuid.UUID, db: SessionDep) -> BrainAtlasRead:
+    with ensure_result(error_message="Brain Atlas not found"):
+        stmt = sa.select(BrainAtlas).filter(BrainAtlas.id == id_)
+        row = db.execute(stmt).scalar_one()
+    return BrainAtlasRead.model_validate(row)
```
eleftherioszisis, line 41: Could be replaced with `app.queries.common.router_read_one`.

app/service/brain_atlas.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+
+def read_one_region(id_: uuid.UUID, region_id: uuid.UUID, db: SessionDep) -> BrainAtlasRegionRead:
+    with ensure_result(error_message="Brain Atlas Region not found"):
+        stmt = sa.select(BrainAtlasRegion).filter(
+            sa.and_(BrainAtlasRegion.brain_atlas_id == id_, BrainAtlasRegion.id == region_id)
+        )
+        row = db.execute(stmt).scalar_one()
+    return BrainAtlasRegionRead.model_validate(row)
```
eleftherioszisis, line 72: This one too probably with an additional filter query.

app/service/brain_atlas.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def read_one(id_: uuid.UUID, db: SessionDep) -> BrainAtlasRead:
+    with ensure_result(error_message="Brain Atlas not found"):
+        stmt = sa.select(BrainAtlas).filter(BrainAtlas.id == id_)
+        row = db.execute(stmt).scalar_one()
+    return BrainAtlasRead.model_validate(row)
+
+
+def read_many_region(
```
eleftherioszisis, line 44: Is this providing the atlas context to the query? It seems it is missing.

app/db/model.py
mgeplf: its type is bool; but I can change it

app/db/model.py
mgeplf: I hope not, I think we should only support full brains.

app/service/brain_atlas.py
mgeplf: yup

tests/test_brain_atlas.py
mgeplf: I had `NaN`, but that's doesn't serialize according to.
I don't think 0.0. is a good value, because it doesn't show up wrong enough; I should pick a more negative number, so it's easier to see mistakes, though.  Good point.

app/service/brain_atlas.py
mgeplf: yup, it wasn't really being used yet (all atlases are public), but it's best to add it now. tnx.

app/service/brain_atlas.py
mgeplf: done.

tests/test_brain_atlas.py
```diff
@@ -0,0 +1,178 @@
...
+                    "children": [],
+                }
+            ],
+        },
+    ],
+}
+
+
+@pytest.fixture
+def asdf():
```
GianlucaFicarelli, line 50: Probably leftover :)

app/db/model.py
```diff
@@ -875,9 +869,30 @@ class METypeDensity(
...
+
+
+class BrainAtlasRegion(Entity, LocationMixin):
+    __tablename__ = EntityType.brain_atlas_region
+
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+
+    # only the volume for leaf nodes is saved; the consumer must calculate
+    # volumes depending on which view of the hierarchy they are using
+    volume: Mapped[float] = mapped_column(nullable=True)
```
GianlucaFicarelli, line 891: If it's nullable, it should be defined as:
```suggestion
    volume: Mapped[float | None]
```

app/db/model.py
GianlucaFicarelli: This isn't changed, but I find `is_leaf_region` clearer as well

app/service/brain_atlas.py
```diff
@@ -0,0 +1,83 @@
...
+def read_one_region(
+    user_context: UserContextDep, atlas_id: uuid.UUID, atlas_region_id: uuid.UUID, db: SessionDep
+) -> BrainAtlasRegionRead:
+    return app.queries.common.router_read_one(
+        id_=atlas_region_id,
+        db=db,
+        db_model_class=BrainAtlasRegion,
+        authorized_project_id=user_context.project_id,
+        response_schema_class=BrainAtlasRegionRead,
+        apply_operations=lambda q: q.filter(BrainAtlasRegion.brain_atlas_id == atlas_id),
```
GianlucaFicarelli, line 82: Minor, can we use `q.where` instead of `q.filter` in new code here and at line 62?
They are aliases, but `where` is more similar to how the conditions is represented in the final query.

mgeplf: let me try again, I think I did that the first time, but I'm not sure.

app/cli/import_data.py
```diff
@@ -555,47 +560,97 @@ def ingest(
...
-            createdAt, updatedAt = utils.get_created_and_updated(data)
+            atlas_region = db.execute(
+                sa.select(BrainAtlasRegion).filter(
+                    BrainAtlasRegion.brain_region_id == str(brain_region_id)
+                )
+            ).scalar_one_or_none()
+
+            if atlas_region is None:
+                annotation_id = curate.curate_brain_region(brain_region_data)["@id"]
+                volume = -1_000_000_000
```
GianlucaFicarelli, line 633: This -1_000_000_000 looks like a magic number, should it be moved to a constant? Or can we use None?

tests/test_brain_atlas.py
GianlucaFicarelli: In tests this is still -1, should it be changed to -1_000_000_000?
What about returning None?

app/db/model.py
mgeplf: actually, I will make it nullable!

eleftherioszisis (approved):

PR #196, opened 2025-05-26
clean commit of Scientific Artifact based off of 26.5.2025 entitycore

dkeller9:
This branch has only scientific artifact schema and model files added.

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1024,5 +1024,4 @@ class ScientificArtifactPublicationLink(Identifiable):
 
     __table_args__ = (
         UniqueConstraint("publication_id", "scientific_artifact_id", name="uq_publishedin_ids"),
-        {"extend_existing": True},
     )


app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,57 @@
...
+from app.schemas.contribution import ContributionRead
+from app.schemas.asset import AssetRead
+from app.schemas.base import LicenseRead,BrainRegionRead, EntityTypeMixin
+from app.db.model import Entity
+from enum import Enum, auto
+
+#the nature of the publication
+class PubType(str, Enum):
+    EntitySource = "EntitySource" #where the model is published
+    ComponentSource  = "ComponentSource" #source of any one item used in the model)
```
danielaegassan, line 19: Nice, this will work.  Extra parenthesis at the end of comment (typo)

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,57 @@
...
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    original_source_location : str
+    other: str | None = None # Optional alternative identifier as a string
+    publication_type: PubType | None = None # Optional publication type as a string
+
+class ScientificArtifactBase(EntityTypeMixin):
+    model_config = ConfigDict(from_attributes=True)
+    creation_date: datetime| None = None
+    update_date: Optional[datetime] = None
```
danielaegassan, line 38: I thought update_date is also mandatory in entity core in general

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,57 @@
...
+    name: str
+    description: Optional[str] = None
```
danielaegassan, line 40: I thought description is mandatory (at least it is in the word doc.)

danielaegassan (commented):
This mostly works for me, I just left a few comments on some fields I thought would be mandatory.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Represents ways to define or identify an entity, such as by DOI, PMID, UUID, or other means.
+    All fields are optional.
+    """
+    DOI: str | None = None  # Optional DOI (Digital Object Identifier) as a string
+    PMID: int | None = None # Optional PMID (PubMed Identifier) as an integer
+    UUID: uuid.UUID | None = None # Using uuid.UUID type for clarity, assuming it refers to a standard UUID
+    original_source_location : str
+    other: str | None = None # Optional alternative identifier as a string
+    publication_type: PubType | None = None # Optional publication type as a string
```
chr-pok, line 33: Would require a change here as well:
```suggestion
    publication_type: PublicationType | None = None # Optional publication type as a string
```

app/db/model.py
```diff
@@ -372,6 +372,26 @@ class Entity(LegacyMixin, Identifiable):
...
+    SubjectMixin,
+    NameDescriptionVectorMixin,
+    SpeciesMixin,
+    LocationMixin,
+    LicensedMixin
+):
+    __tablename__ = "scientific_artifact"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    experiment_date: Mapped[datetime] = mapped_column(DateTime, nullable=True)
+    published_in: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
```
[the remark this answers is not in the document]
GianlucaFicarelli: If it's specified as JSONB as above, it's a json blob that can contain any json data.
In this case the content can be validated with a pydantic schema at the application level (instead of the db level) when the blob is written and read.

In RDBMS the first choice is usually to use a table, but postgresql allows to use JSONB, that can be a good choice in some use cases (when no referential integrity is needed, and when the schema needs to be more complex, or needs flexibility in case of changes).
Note also that searching in a JSONB is less performant than searching in a column, so this could be taken in consideration as well.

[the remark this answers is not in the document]
danielaegassan: @GianlucaFicarelli I am a bit confused as to how we will actually use this in the UI.  At the moment, for the circuits in Explore, we are using a version of PublishedIn that has more data.  More precisely, we have the following:

"PublicationType": str from controlled vocabulary (as in Dan's suggestion)
"title": str,
"authors": list of {"name", "lastname"} entries
"doi": str
"url": str
"journal": str
"publicationDate": Date
"abstract": str

I thought we would put all but the first on a table once, and then add the PublicationType as some kind of relationship that explains how this is related to the entity.

If we just have a json for each entity we will have a lot of redundant data.  For instance, at the moment, we have 77 circuits in Explore but only 3 types of publication jsons, because the sources of many are the same.  Does this inform the decision on what is the best way to go?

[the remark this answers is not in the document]
danielaegassan: @GianlucaFicarelli  It's a single relation, meaning for a given entity, a PublishedIn can be only of one type.  Either: 

- EntitySource: It's a publication where the entity itself was published (could be more than one) 
- ComponentSource: It's a publication where components of this entity where published (i.e., the morphology of a circuit)
- Application: It's a publication where this model was used 

A publication can only be one of these for a given entity, but it could have a different role for another entity.

We use this now in Explore > Circuits > Provenance/Related Publications.  And the idea is to be able to do that for all model data.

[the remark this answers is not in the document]
chr-pok: When thinking about it, I believe `PublishedIn` and `Publication` could/should be even two separate tables:
- `PublishedIn` only contains a `PublicationType` and a `Publication` ID, and a `ScientificArtifact` ID it belongs to.
- `Publication` contains the actual information about a publication, like DOI, title, authors, abstract, etc. as mentioned by @danielaegassan before.

In this way, every scientific artifact could be associated to any number of publications of any type without duplicating publications. Would this make sense @GianlucaFicarelli?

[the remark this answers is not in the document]
GianlucaFicarelli: If PublishedIn contains the ids as in the comment by @chr-pok , then it should be a table to ensure referential integrity between the other tables.
This would be an association table that allows many-to-many relationships (this is why I asked if multiple relations were needed):
- a ScientificArtifact can be linked to multiple publications
- a publication can be linked to multiple ScientificArtifacts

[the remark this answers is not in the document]
chr-pok: Yes exactly, it is a many-to-many relationship!

And the association table (`PublishedIn`) would also contain information about the `PublicationType`, i.e., a given publication could be linked to different or even the same scientific artifact with different types (but only once for a given type and publication). So the triple (scientific_artifact_id, publication_id, publication_type) should be unique (=primary key, I suppose).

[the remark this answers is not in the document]
chr-pok: > However, for a given entity it can only have one PublicationType.

Why? I think it could happen that a publication is the source publication of an artifact but also shows some applications, in which case it could be both "EntitySource" and "Application", no?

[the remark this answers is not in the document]
danielaegassan: This is not how we defined it though. In particular in the UI, the source articles already show up twice: in provenance and in related literature.  If we add this option they will show up three times. I think, generally all publications of a model some sort of application.  We can discuss this further in the office.

app/db/model.py
```diff
@@ -372,6 +372,26 @@ class Entity(LegacyMixin, Identifiable):
         "polymorphic_on": "type",
     }
 
+class ScientificArtifact(
+    Entity,
+    SubjectMixin,
+    NameDescriptionVectorMixin,
+    SpeciesMixin,
```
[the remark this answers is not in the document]
eleftherioszisis: I believe we will eventually move to using Subject everywhere, which includes the species, therefore no need to also include a SpeciesMixin here.

app/db/model.py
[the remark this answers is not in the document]
chr-pok: Right, we had this discussion already, so let's stick to what we agreed on already. In this case, if I understand correctly, this would mean that the tuple (scientific_artifact_id, publication_id) should be unique (=primary key; not including publication_type), so that there can only be one association (which could have any type).

app/schemas/scientific_artifact.py
GianlucaFicarelli: Usually the convention is to set the update_date the same as the creation_date, when the record is created, so both creation_date and update_date should be not nullable.

danielaegassan: Right, I didn't think about that.

GianlucaFicarelli: I need to add a clarification: my considerations were for the db model, while for the pydantic schema used for the web API, these fields are required in the Read schema (used when fetching a resource), but they shouldn't be present in the Create schema (used to create a resource)

app/db/model.py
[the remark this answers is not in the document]
dkeller9: Okay, will remove SpeciesMixin.

app/db/model.py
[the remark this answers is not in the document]
danielaegassan: Unfortunately Christop is off until Monday.  What I understand the suggestion is, that we use: 

1. A table "Publications" that is a table with all publications used and it has unique entries.  It has 

```
"id": UUID 
"title": str,
"authors": list of {"name", "lastname"} entries
"identifier": "DOI"/"PMID"/"Other"
"url": str (optional)
"journal": str (optional)
"publicationDate": Date
"abstract": str (optional?)
```

2. Another table (or relation?) "PublishedIn" that has: 

```
"scientific_artifact_id": UUID
"publication_id": UUID
"PublicationType": str from controlled vocabulary  
```

Alternatively, this could be a json inside ScientificArtifact of the form 
[ {"publication_id": UUID,  "PublicationType": str from controlled vocabulary"}]


Maybe @GianlucaFicarelli can guide us here?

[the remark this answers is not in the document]
GianlucaFicarelli: Correct, there should be 2 additional tables, NOT a json because it cannot easily ensure the referential integrity, as I mentioned in a previous comment.

To prevent duplications in the publication table, it would be better if it's possible to define a unique key besides the id. Is the identifier a good candidate? But what if a publication has multiple ids? ("DOI"/"PMID"/"Other")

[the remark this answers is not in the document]
dkeller9: Because of the quirks of the publication process, the same publications cannot be assured of having the same identifier or even an identifier at all.

app/schemas/scientific_artifact.py
dkeller9: corrected thank you

app/db/model.py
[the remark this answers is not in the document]
danielaegassan: If there is a DOI or PMID we can make them unique but if there is no identifier, we can try to check unicity on the name, but we can't assure it .. but maybe that's just the best we can do for this case?

app/db/model.py
```diff
@@ -373,17 +378,57 @@ class Entity(LegacyMixin, Identifiable):
...
-    age_value: Mapped[timedelta | None]
-    age_min: Mapped[timedelta | None]
-    age_max: Mapped[timedelta | None]
-    age_period: Mapped[AgePeriod | None]
-    sex: Mapped[Sex | None]
-    weight: Mapped[float | None]  # in grams
+    DOI: Mapped[str | None] = mapped_column(String, nullable=True, index=True)
+    PMID: Mapped[int | None] = mapped_column(BigInteger, nullable=True, index=True)
+    original_source_location: Mapped[str | None] = mapped_column(String, nullable=True)
+    other: Mapped[str | None] = mapped_column(String, nullable=True)
```
eleftherioszisis, line 389: Not sure this is a helpful field. It's not cleat what this "other" refers to.

app/db/model.py
```diff
@@ -373,17 +378,57 @@ class Entity(LegacyMixin, Identifiable):
...
+    publication_year: Mapped[int | None] = mapped_column(sa.Integer, nullable=True)
+    abstract: Mapped[str | None] = mapped_column(String, nullable=True)
 
-    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+    }
+
+
+class PublishedIn(Base):
```
eleftherioszisis, line 402: Could we maybe derive from Identifiable? We should aim for most of our tables to have creation_date/update_date. There is also a PR pending that will add created_by/updated_by in Identifiable too.

app/db/model.py
```diff
@@ -395,6 +440,34 @@ def subject(cls):
         return relationship("Subject", uselist=False, foreign_keys=cls.subject_id)
 
 
+class ScientificArtifact(
+    Entity, SubjectMixin, NameDescriptionVectorMixin, LocationMixin, LicensedMixin
+):
+    """Base class for scientific artifacts."""
+
+    __tablename__ = "scientific_artifact"
```
eleftherioszisis, line 448: Add it in `app.db.types` EntityType and declare it here as `EntityType.scientific_artifcact.value`

app/db/model.py
```diff
@@ -395,6 +440,34 @@ def subject(cls):
...
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    experiment_date: Mapped[datetime | None] = mapped_column(DateTime)
+    contact_id: Mapped[uuid.UUID | None] = mapped_column(ForeignKey("person.id"), nullable=True)
+
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+    }
+
+
+class Subject(NameDescriptionVectorMixin, SpeciesMixin, Entity):
```
eleftherioszisis, line 458: It would be best if things that are not changed to not show in changes.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,91 @@
...
+    ENTITYSOURCE = "EntitySource"  # where the model is published
+    COMPONENTSOURCE = "ComponentSource"  # source of any one item used in the model
+    APPLICATION = "Application"  # any publication which uses the model
+
+
+class Author(TypedDict):
+    """The names of authors of a publication."""
+
+    name: str
+    lastname: str
```
eleftherioszisis, line 27: We can be consistent here with the Agent fields, i.e. `given_name` and `family_name`.

app/schemas/scientific_artifact.py
```diff
@@ -0,0 +1,91 @@
...
+from app.schemas.base import BrainRegionRead
+from app.schemas.entity import EntityRead
+
+
+class PublicationType(StrEnum):
+    """The type of publication for a scientific artifact."""
+
+    ENTITYSOURCE = "EntitySource"  # where the model is published
+    COMPONENTSOURCE = "ComponentSource"  # source of any one item used in the model
+    APPLICATION = "Application"  # any publication which uses the model
```
eleftherioszisis, line 20: How about simply:
```
entity_source = auto()
component_source = auto()
application = auto()
```

dkeller9: People may be contributing this from manually-constructed jsons so it might be more convenient for them to have user-interpretable strings.

eleftherioszisis: Yeah, what `auto()` does in the case of StrEnum it assigns the same value as key and makes things easier to read. So:

```python
class A(StrEnum):
    entity_source = auto()
    component_source = auto()
 ```
 
 Is equivalent to:
 
 ```python
 class A(StrEnum):
     entity_source = "entity_source"
     component_source = "component_source"
 ```
 
 Note that the above is true only for snake case and small letter keys. A.ENTITY_SOURCE for example would have a value of "entity_source" with auto().

File: app/db/model.py [hunk 13]
```diff
@@ -373,17 +378,57 @@ class Entity(LegacyMixin, Identifiable):
...
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
-    age_value: Mapped[timedelta | None]
-    age_min: Mapped[timedelta | None]
-    age_max: Mapped[timedelta | None]
-    age_period: Mapped[AgePeriod | None]
-    sex: Mapped[Sex | None]
-    weight: Mapped[float | None]  # in grams
+    DOI: Mapped[str | None] = mapped_column(String, nullable=True, index=True)
+    PMID: Mapped[int | None] = mapped_column(BigInteger, nullable=True, index=True)
+    original_source_location: Mapped[str | None] = mapped_column(String, nullable=True)
```
jdcourcol, line 388: Is that a URL ?

dkeller9: original_source_location could be a URL

jdcourcol: what is the difference with url property below ?

dkeller9: The url field below should probably be used if there is a url. @danielaegassan do you have any thoughts? Is this field redundant?

app/db/model.py
```diff
@@ -373,17 +379,58 @@ class Entity(LegacyMixin, Identifiable):
...
+    publication_year: Mapped[int | None] = mapped_column(sa.Integer, nullable=True)
+    abstract: Mapped[str | None] = mapped_column(String, nullable=True)
 
-    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+    }
+
+
+class PublishedIn(Identifiable):
```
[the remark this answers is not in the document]
dkeller9: Good idea- the change has been made.

app/db/model.py
```diff
@@ -395,6 +442,34 @@ def subject(cls):
...
 
+class ScientificArtifact(
+    Entity, SubjectMixin, NameDescriptionVectorMixin, LocationMixin, LicensedMixin
+):
+    """Base class for scientific artifacts."""
+
+    __tablename__ = __tablename__ = EntityType.scientific_artifact.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    experiment_date: Mapped[datetime | None] = mapped_column(DateTime)
+    contact_id: Mapped[uuid.UUID | None] = mapped_column(ForeignKey("person.id"), nullable=True)
```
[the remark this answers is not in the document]
dkeller9: Are non-OBI people not allowed in the person table?

app/db/model.py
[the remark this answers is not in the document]
dkeller9: The logic should be in the code that populates the database.

app/db/model.py
```diff
@@ -373,17 +378,57 @@ class Entity(LegacyMixin, Identifiable):
...
-    sex: Mapped[Sex | None]
-    weight: Mapped[float | None]  # in grams
+    DOI: Mapped[str | None] = mapped_column(String, nullable=True, index=True)
+    PMID: Mapped[int | None] = mapped_column(BigInteger, nullable=True, index=True)
+    original_source_location: Mapped[str | None] = mapped_column(String, nullable=True)
+    other: Mapped[str | None] = mapped_column(String, nullable=True)
+    title: Mapped[str | None] = mapped_column(String, nullable=True)
+    authors: Mapped[list[Author] | None] = mapped_column(JSONB, nullable=True)
+    url: Mapped[str | None] = mapped_column(String, nullable=True)
+    journal: Mapped[str | None] = mapped_column(String, nullable=True)
```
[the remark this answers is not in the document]
danielaegassan: @dkeller9, @jdcourcol, at the moment, we are using this to show citations in the UI for circuits.  I see two possible options for removing this. 

1. We pull the journal from some data base.  In this case, what do we do if DOI was not added?  Can we strongly encourage to make it mandatory somehow?  
2. We do not to show the journal in the UI.  However, note that this is already showing in the response from the agent for instance, so we would need to make it consistent.

app/db/model.py
danielaegassan: @jdcourcol @dkeller9 I think again, the issue here is what to do for sources that don't have a DOI.  So we have here multiple options for an identifier. 

It would be great if we can enforce that most times there is a DOI and only in other cases, one might want to give a different identifier e.g., url to bioarxiv.

james-isbister:
Suggestions from scientific meeting @jdcourcol :

- Take out PMID, journal, other_identifier, URL
- Make DOI mandatory 
- In future, have journal table or pull from journal table

app/db/model.py
danielaegassan: After a meeting with the science team, a general more restrictive suggestion addressing these issues is given in (https://github.com/openbraininstitute/entitycore/pull/196#issuecomment-2935215696)

jdcourcol (approved):

PR #197, opened 2025-05-26
Use NameDescriptionVectorMixin for License and ElectricalRecordingStimulus

mgeplf:


alembic/versions/20250526_100043_c4fa9c327b08_license_use_namedescriptionvectormixin.py
```diff
@@ -0,0 +1,62 @@
...
+        signature="license_description_vector",
+        on_entity="public.license",
+        is_constraint=False,
+        definition="BEFORE INSERT OR UPDATE ON license\n            FOR EACH ROW EXECUTE FUNCTION\n                tsvector_update_trigger(description_vector, 'pg_catalog.english', description, name)",
+    )
+    op.drop_entity(public_license_license_description_vector)
+
+    op.drop_index("ix_license_description_vector", table_name="license", postgresql_using="gin")
+    op.drop_column("license", "description_vector")
+    # ### end Alembic commands ###
```
GianlucaFicarelli, line 62: It should be possible to merge the 2 migration files.
Minor, because maybe we'll need to rebuild the migrations again.

PR #206, opened 2025-05-30
Remove legacy distribution metadata from assets during import

GianlucaFicarelli:


eleftherioszisis (approved):

GianlucaFicarelli:
The legacy dict has been removed from the asset meta field in staging, so the response payload is smaller already.

PR #208, opened 2025-06-02
Simulation, SimulationCampaign and SimulationReport initial model

jdcourcol:
A PR to move forward on [this.](https://github.com/openbraininstitute/obi-one/issues/132)

@eleftherioszisis : 
(Activities are in bold)
SimulationCampaign->**SimulationGeneration**->Simulation(s)->**SimulationExecution**->SimulationReport

- Adding Activity class following OPM
- adding SimulationExecution & SimulationGeneration as children of activity. 
- adding SimulationCampaign, Simulation & SimulationReport

- REST end points and tests will once model is defined.


app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+
+class SimulationCampaign(
+    NameDescriptionVectorMixin,
+    Entity,
+):
+    """Represents a simulation campaign entity in the database.
+
+    it has an asset which is the simulation campaign configuration file.
```
james-isbister, line 943: There will probably be two types of simulation campaign configuration:

1. The dump of the obi-one SimulationsForm that is used to generate the campaign. This will be useful for repopulating the GUI if a) the user has not yet generated the campaign but comes back to edit/launch it later, b) to see the parameterization of the campaign (in non-edit mode) after the campaign has been generated.
2. A file similar conceptually to the bbp-workflow campaign config, which summarizes the different simulations generated in the campaign, and can be used (for example) as the input to a blueetl analysis.

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+
+    Attributes:
+        id (uuid.UUID): Primary key for the simulation campaign, referencing the entity ID.
+    """
+
+    __tablename__ = EntityType.simulation_campaign.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    simulations = relationship("Simulation", back_populates="simulation_campaign")
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
```
james-isbister, line 952: As mentioned by @jdcourcol we should consider how the dimensions of the scan are represented (i.e. {"ca": [1.05, 1.1], "random_seed": [1, 2, 3]} in the SimulationCampaign Entity

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    __tablename__ = EntityType.simulation.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    simulation_campaign_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("simulation_campaign.id"), index=True
+    )
+    simulation_campaign: Mapped[SimulationCampaign] = relationship(
+        "SimulationCampaign",
+        uselist=False,
+        foreign_keys=[simulation_campaign_id],
+    )
```
james-isbister, line 973: We should also consider how the scan parameters for the individual simulation are represented (i.e. {"ca": 1.1, "random_seed": 2} in the SimulationCampaign Entity

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("activity.id"), primary_key=True)
+    status: Mapped[SimulationExecutionStatus] = mapped_column(
+        Enum(SimulationExecutionStatus, name="simulation_execution_status"),
+        default=SimulationExecutionStatus.created,
+    )
+
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+
+class SimulationReport(Entity):
```
james-isbister, line 995: Is there one SimulationReport per SimulationExecution which references multiple report files (i.e. spikes.h5, voltage.h5)? Or multiple SimulationReports per SimulationExecution, with each SimulationReport referencing a single file?

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+    """Represents a simulation report entity in the database.
+
+    it has an asset which is the SONATA simulation report file.
+
+    Attributes:
+        id (uuid.UUID): Primary key for the simulation report, referencing the entity ID.
+    """
+
```
jdcourcol, line 1003: There are limitation in the cloud and expectations from SONATA.
We cannot use absolute path, and relative path are relative to the simulation config.

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+class SimulationExecution(Activity):
+    """Represents a simulation execution entity in the database.
+
+    Attributes:
+        id (uuid.UUID): Primary key for the simulation execution, referencing the entity ID.
+    """
+
+    __tablename__ = ActivityType.simulation_execution.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("activity.id"), primary_key=True)
```
james-isbister, line 986: Does this mean we store just one SimulationExecution per Simulation? I think that would certainly simplify things

app/db/model.py
jdcourcol: Due to how sonata is specified, I fear we can have only one SimulationReport per SimulationExecution, and this simulation report has multiple assets.

app/db/model.py
```diff
@@ -896,3 +932,88 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+
+    Attributes:
+        id (uuid.UUID): Primary key for the simulation campaign, referencing the entity ID.
+    """
+
+    __tablename__ = EntityType.simulation_campaign.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    simulations = relationship("Simulation", back_populates="simulation_campaign")
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
```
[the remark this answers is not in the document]
jdcourcol: >I would suggest using a more general EntityReference, especially if we want to support other simulation types like ion channels later on

For the time being, we have Entity, ScientificArtefact, Circuit. We can go for ScientificArtefact.

app/db/model.py
james-isbister: The naming might just be slightly confusing then? In sonata a simulation can have multiple reports, each represented by a different file.

Perhaps the Entity could just be named differently like SimulationOutput, for example?

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: I put entity as the other entities are not merged yet.

app/db/model.py
jdcourcol: resolved with last commit.

app/db/model.py
jdcourcol: So here, we would need that especially when we list simulation campaigns. 
let's take an example:

1. in the SimulationForm we have something like (take it as an illustration not actual function) :

`param1: {"np.random.choice": { range: 5, nb_value:2 }}` 
2. in the bbp-workflow like config:
`param1: [0, 3, 4]`
3. in one of the simulation config:
`param1: 3`

What do we want to see in the list of simulation campaign ?  1) or 2) ?

I would imagine 1) as this is what the user provided.
However, in the detail view, I may want to see 2)

@james-isbister ?

app/db/model.py
```diff
@@ -246,6 +249,39 @@ class Organization(Agent):
     }
 
 
+class Activity(Identifiable):
+    __tablename__ = "activity"
+    type: Mapped[ActivityType]
+    start_time: Mapped[datetime | None]
```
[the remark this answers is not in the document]
jdcourcol: I put that initially, however it is not in the opm spec.

app/db/model.py
```diff
@@ -246,6 +249,39 @@ class Organization(Agent):
...
+    end_time: Mapped[datetime | None]
+    used_ids: Mapped[list[uuid.UUID]] = mapped_column(
+        ForeignKey("entity.id", ondelete="CASCADE"),
+        index=True,
+        default=[],
```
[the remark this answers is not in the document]
jdcourcol: yeah, I can imagine a job that is generating things from thin air. but let's say this should not be the case indeed.

app/db/model.py
jdcourcol: ok, we start then with a dictionary of list. we will have to see how we can make this searchable though in the GUI.

app/db/model.py
```diff
@@ -896,3 +944,113 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+        uselist=False,
+        foreign_keys=[simulation_campaign_id],
+    )
+    entity_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), index=True)
+    entity: Mapped[Entity] = relationship(
+        "Entity",
+        uselist=False,
+        foreign_keys=[entity_id],
+    )
+    scan_parameters: Mapped[JSON_DICT] = mapped_column(
```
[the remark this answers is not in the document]
james-isbister: For the simulation entity, they are a subset of the parameters (specifically the ones which are varied in the parent scan/campaign). It may be good for this to be made clear in the name

app/db/model.py
james-isbister: > Why do we need a default? Is it possible for example for the user to choose from the GUI a SimulationResult entity to analyze?

> Speaking of analysis, what is the input for an analysis task? Is it the generated entity of a simulation execution, i.e. a SimulationResult, or something more complex?

For analysis, we foresee that the user will select all or a subset of simulations in a campaign that have completed, and analyse the generated simulation activity (I mean spikes and traces) using BlueETL, which takes as input a reference to the campaign_config.json and an optional filter of the simulations. Note on a local filesystem it ignores the simulations in the campaign that have not yet completed.

We will need to think about how this will be achieved on the platform, either through the GUI or a notebook, and how it can also work locally if the user downloads the simulation data. In all cases, the user will be analyzing a "campaign" though even if it has just one simulation or just one or a subset of simulations have completed or are selected for analysis.

For each Simulation in the SimulationCampaign we don't want the user to have to select which SimulationResult to analyse from various completed (i.e. successful) SimulationExecutions, as there could be 100s of Simulations in a Campaign. This is why I think some kind of default completed SimulationExecution is necessary.

Also, I'm not sure we want to suggest to the user that rerunning a completed simulation could give a different output somehow if none of the parameters have changed. Once a simulation is run successfully the output should be final in the users mind. There is no real reason to re-run it. If the user does re-run it, it would be simpler if they got the sense that they were overwriting the previous Simulation. Then they would be more inclined to generate a new campaign if they really want to re-run something (with different hardware for example).

I understand that it is important to represent additional failed/cancelled Simulations for debugging, but once a simulation has run successfully, I'm not even sure the past failures need to be presented to the user (maybe in future, but its not a priority). For now, I think its enough, that they see the status of the last attempted SimulationExecution for each Simulation.

app/db/model.py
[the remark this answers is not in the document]
eleftherioszisis: If these are parameters for the simulation they should be named as such and not carry obi-one's scan context as it is not meaningful at this level.

If someone generates a single simulation manually without a campaign wouldn't it make more sense for this to be called just parameters? Why the prefix scan is important here? For the simulation it doesn't matter how the parameters are generated, just that these are the parameters to be used.

It would be less confusing to not enforce context in entities that do not need it. Please correct me if I'm wrong in my reasoning.

app/db/model.py
```diff
@@ -896,3 +944,113 @@ class CellComposition(NameDescriptionVectorMixin, LocationMixin, SpeciesMixin, E
...
+
+    it has an asset which is the simulation campaign configuration file.
+
+    Attributes:
+        id (uuid.UUID): Primary key for the simulation campaign, referencing the entity ID.
+    """
+
+    __tablename__ = EntityType.simulation_campaign.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    simulations = relationship("Simulation", back_populates="simulation_campaign")
```
eleftherioszisis, line 965: If the SimulationCampaign is used by two SimulationCampaignGeneration activities (or whichever activity registers the simulation configs) it will list the union of all simulations from all activities.

Is this what we want?

Maybe we can have a SimulationCampaignGenerationResult (or better named) that links all simulations associated with that activity and the resulting config from the generation activity?

james-isbister: Yes, some form of linking like this sounds good!

Similarly to SimulationExecution, I would say that there should be some sort of default SimulationCampaignGenerationResult, and we maybe don't need the user to rerun a SimulationCampaignGeneration if one has already completed successfully. Rather they should "duplicate" the SimulationCampaign and rerun it if they want to do this

james-isbister:
Sorry @eleftherioszisis @jdcourcol  I should have better communicated the context/plan for the GUI

The idea is that the user creates or duplicates a campaign (note they can also see the "scan parameters" here): 
Screenshot 2025-06-04 at 10 38 46

Then edits the parameters of the campaign:
Screenshot 2025-06-04 at 10 39 00

Then generates the campaign where they can view the "scan_parameters" and simulation input files of each simulation, and can select which ones to run:
Screenshot 2025-06-04 at 10 41 27

Then they can view the status of each simulation:
Screenshot 2025-06-04 at 10 39 29

And select one to view the results (files for now):
Screenshot 2025-06-04 at 10 44 39

In this context, the user does not see multiple re-runs of successfully completed simulations (due to the complications described above. Rather they should duplicate a campaign as in the first image.

Moreover, they will be able to select a subset of completed simulations and press Analyse, which will for now launch a Jupyter notebook

jdcourcol:
> Sorry @eleftherioszisis @jdcourcol I should have better communicated the context/plan for the GUI
> 
> The idea is that the user creates or duplicates a campaign (note they can also see the "scan parameters" here in the list of campaigns): Screenshot 2025-06-04 at 10 38 46
> 
> Then edits the parameters of the campaign: Screenshot 2025-06-04 at 10 39 00
> 
> Then generates the campaign where they can view the "scan_parameters" and simulation input files of each simulation, and can select which ones to run: Screenshot 2025-06-04 at 10 41 27
> 
> Then they can view the status of each simulation: Screenshot 2025-06-04 at 10 39 29
> 
> And select one to view the results (files for now): Screenshot 2025-06-04 at 10 44 39
> 
> In this context, the user does not see multiple re-runs of successfully completed simulations (due to the complications described above. Rather they should duplicate a campaign as in the first image.
> 
> Moreover, they will be able to select a subset of completed simulations and press Analyse, which will for now launch a Jupyter notebook
> 
> Also when they go back to a previously generated campaign, the parameters will be non-editable

@james-isbister . That is not incompatible with multiple execution of the simulation config. In the GUI above, we see the status "failed" for some of them. This can come from many reasons (nodes not available, tmp bug, whatever...). The user may get the support to fix the issue or retry later. then we have another execution. This execution will again be successful or not. Therefore, we can log it. Also, we have an "activity" panel where we log all the executions (successful or not).
Finally, indeed, we prefer to avoid re-running a successful execution. So the GUI should prevent the user to do that. (except may be in some cases where we want to force the re-run )

app/db/model.py
jdcourcol: >that references the one that should be seen in the GUI or used in analysis.

What we can do it to use only the successful execution. and for the time being assume that we have only 1 successful execution. (if we want to have more, I would mark the previous one as deprecated).

james-isbister: Sounds good!

app/db/model.py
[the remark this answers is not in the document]
james-isbister: > I would prefer to keep it consistent whatever the number of simulations.

I think for a campaign with one simulation it would just be an empty dictionary then, as none of the parameters are really scan parameters. Different parts of the GUI would just have to display something slightly different like "N/A" in Explore for example

james-isbister:
The outcome of all the discussions above seem good to me now @jdcourcol @eleftherioszisis

app/db/types.py
```diff
@@ -94,6 +104,14 @@ class SingleNeuronSimulationStatus(StrEnum):
     success = auto()
 
 
+class SimulationExecutionStatus(StrEnum):
+    created = auto()
```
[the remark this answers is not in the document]
pgetta: Looks good to me.

app/db/model.py
jdcourcol: ok. Let's ignore that second file for now. We will see how we integrate BlueETL in a second step.

app/db/model.py
```diff
@@ -287,6 +287,28 @@ class Organization(Agent):
...
+
+class EntityActivityGenerated(Base):
+    __tablename__ = "entity_activity_generated"
+    entity_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    activity_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("activity.id"), primary_key=True)
+    __table_args__ = (
+        UniqueConstraint(
+            "entity_id", "activity_id", name="uq_entity_activity_generated_entity_id_activity_id"
+        ),
+    )
```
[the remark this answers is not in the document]
jdcourcol: ah ok

eleftherioszisis (approved):

PR #209, opened 2025-06-03
Expose Person subject id in payload

eleftherioszisis:


mgeplf (approved):

bilalesi:
I think the entities are missing a filter for `sub_id`, similar to how we currently filter by `created_by`.
If filtering by `sub_id` is not part of the entitycore design, I might use the person endpoint instead, but it should supports filtering to retrieve the relevant `creator_id`.

PR #224, opened 2025-06-13
Sim provenance

eleftherioszisis:


app/db/utils.py
```diff
@@ -50,11 +50,16 @@ def load_db_model_from_pydantic[I: Identifiable](
...
     created_by_id: uuid.UUID,
     updated_by_id: uuid.UUID,
+    ignore_attributes: set[str] | None = None,
 ) -> I:
     data = json_model.model_dump(by_alias=True) | {
         "created_by_id": created_by_id,
         "updated_by_id": updated_by_id,
     }
-    if issubclass(db_model_class, Entity):
+    if "authorized_project_id" in db_model_class.__mapper__.columns:
```
eleftherioszisis, line 59: activities also have project ids now, so we cannot limit it to Entity derived tables anymore. Instead I check if authorized_project_id is in the columns of the current table or its parents.

jdcourcol (approved):
I don't have any particular comment. We will see if Pavlo / Gil can use it.

pgetta (approved):
Looks good to me, let's ship it!

PR #225, opened 2025-06-16
Add create endpoints for public mtype/etype classificiation

eleftherioszisis:
Allow creating etype/mtype classification that are constrained to be public for now.

This will allow to unblock the mtype registration for CNS morphologies, which should be public.


app/service/mtype_classification.py
```diff
@@ -0,0 +1,55 @@
...
+        user_context.project_id,
+    )
+    if db.execute(stmt).scalar_one() == 0:
+        L.warning("Attempting to create an annotation for an entity inaccessible to user")
+        raise HTTPException(status_code=404, detail=f"Cannot access entity {json_model.entity_id}")
+
+    if not json_model.authorized_public:
+        L.warning("Attempting to create a private classification, which is not supported.")
+        raise HTTPException(
+            status_code=404,
```
jdcourcol, line 44: may be 400 instead. same above.

jdcourcol (changes requested):

app/service/mtype_classification.py
eleftherioszisis: done

PR #226, opened 2025-06-16
S3 directory

mgeplf:


app/service/asset.py
```diff
@@ -129,3 +138,134 @@ def delete_entity_asset(
...
+        )
+
+    entity = entity_service.get_writable_entity(
+        repos,
+        user_context=user_context,
+        entity_type=entity_type,
+        entity_id=entity_id,
+    )
+
+    # or do we want to use a user supplied name, and if so, how do we handle conflicts
```
mgeplf, line 175: we need to make a decision about this, and what the semantics are:
1) do we let users "name" the directory?
2) is that how we say an asset is unique?
3) When we download that directory, is that name included in the path?

@jdcourcol thoughts?

app/schemas/asset.py
```diff
@@ -64,3 +67,24 @@ def ensure_entity_type_label_consistency(self):
 
 class AssetsMixin(BaseModel):
     assets: list[AssetRead]
+
+
+class FileList(BaseModel):
+    files: list[Path]
+    meta: dict | None
+    label: AssetLabel | None
```
mgeplf, line 75: Currently, we have `label`; ie: SWC.
These are "type" checked (ie: morph can have swc/neurolucida, etc).

What do we do with directories?
Can any entity have a directory, or only some (ie: can morph have a directory? or only circuit?)

I can go either way, but my preference is that only things we explicitly say are allowed to have directories can have a directory - ie: we say a circuit can, a simulation, can, etc.

@jdcourcol thoughts?

eleftherioszisis: Wouldn't that limit us though for cases like image directories for instance? To me a directory is like any other asset, and therefore any entity can have one.

In addition, asset metadata has the is_directory flag, which makes it clear that the asset is a directory. Labels can be used even further to constraint what is allowed specifically to each entity.

For example for circuits we can have the label "circuit_directory" so that we don't have to rely on the assumption there is only one directory per entity.

It would be better if we slowly moved away from having assumptions on what an entity contains and rather use labels to explicitly mark the assets, which the consumers then know how to get consistently.

app/routers/asset.py
```diff
@@ -202,6 +204,48 @@ def delete_entity_asset(
...
+        user_context=user_context,
+        entity_type=entity_route_to_type(entity_route),
+        entity_id=entity_id,
+        s3_client=s3_client,
+        files=files,
+    )
+    return AssetAndPresignedURLS.model_validate({"asset": model.dict(), "files": urls})
+
+
+@router.get("/{entity_route}/{entity_id}/assets/{asset_id}/list")
```
eleftherioszisis, line 228:
```
"/{entity_route}/{entity_id}/assets/directory/{asset_id}"
```
wouldn't this be more clear it is only limited to directories and more similar to the upload above? (or with "/list" added)

app/service/asset.py
eleftherioszisis: By having path="" for all directories and not allowing the user to specify a relative path, methods like:
https://github.com/openbraininstitute/entitysdk/blob/30a760a3c138528109b2a0f100079b103a49cbab/src/entitysdk/client.py#L361

Will never work for directories if someone would like to download all the assets of an entity in an output directory.

We should follow the same logic as with files, where a relative nested path is allowed. Not sure why directories should be treated differently.

jdcourcol: for file, the pattern is as follow:
{vlab_id}/{project_id}/assets/{entity-type}/{asset_id}/filename

so for directory, the pattern would be:
{vlab_id}/{project_id}/assets/{entity-type}/{asset_id}/

and the user provide a list like:
file1
path1/file2

that would give:
{vlab_id}/{project_id}/assets/{entity-type}/{asset_id}/file1
{vlab_id}/{project_id}/assets/{entity-type}/{asset_id}/path1/file2

jdcourcol: for downloading multiple assets, you don't have any unicity guarantee in any case.

app/schemas/asset.py
mgeplf: > I think we should do though, but that is not the topic here.

FYI: If a label is given for an asset, it's constrained to which entity types it can be attached.

But ok, we can ignore this for now.

eleftherioszisis (approved):

PR #227, opened 2025-06-16
import with fixes for prod

jdcourcol:
Import updated for new data from prod:

- some e-model have been deprecated: corresponding memodel, synaptome, simulation cannot be imported
- some experiment data uses "years"
- some new data have been importing with wrong mtype mainly.

### DIFF
--- a/app/cli/curate.py
+++ b/app/cli/curate.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 
 from app.cli.brain_region_data import BRAIN_REGION_REPLACEMENTS
 from app.logger import L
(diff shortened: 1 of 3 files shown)


app/cli/curate.py
```diff
@@ -463,6 +517,18 @@ def default_licenses():
...
 
+def curate_license(data):
+    if data.get("@id") == "https://creativecommons.org/licenses/by/4.0":
+        data["@id"] = "https://creativecommons.org/licenses/by/4.0/"
+    return data
+
+
+def curate_hierarchy_name(hierarchy_name):
+    if hierarchy_name == "Isocortex, layer 1":
+        return "Isocortex"
```
eleftherioszisis, line 528: Is this correct? Layer 1 is not equivalent to the entire Isocortex.

app/cli/import_data.py
```diff
@@ -344,6 +346,18 @@ def import_mtype_annotation_body(data, db):
 
 
 def import_etype_annotation_body(data, db):
+    data.extend(
+        [
+            {
+                "label": "bAC_IN",
```
eleftherioszisis, line 352: Should this etype be just "bAC" ?

app/cli/curate.py
jdcourcol: there is no such a region. and it happened on some test data.

app/cli/import_data.py
jdcourcol: that is a new etype from thalamus , Darshan told me.

eleftherioszisis (approved):
lgtm

PR #233, opened 2025-06-18
Constant regions

jdcourcol:
- The uuids of the leaf regions can be initialized with the cell_composition
- because of the seed usage, the basic cell group (not a leaf ) was colliding with a thalamic region, so I had to change it.


app/cli/import_data.py
```diff
@@ -1214,6 +1227,63 @@ def ingest(db, project_context, data_list, all_data_by_id, hierarchy_name: str):
         db.commit()
 
 
+class ImportSynaptome(Import):
```
mgeplf, line 1230: is this meant to be part of this PR?

PR #243, opened 2025-06-23
Import validations, add missing distributions.

jdcourcol:
Import code executed Friday June 20th:
- Import of validations
- fix due to the newly added constraint on content-type
- fetch of missing assets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,11 @@ organize-files:  ## Organize files locally by creating symlinks from the backup 
 	docker compose up --wait db
 	uv run -m app.cli.import_data organize-files $(if $(wildcard curated_files.txt),curated_files.txt,files.txt)
 
+fetch-missing-distributions:
+	@$(call load_env,run-local)
+	docker compose up --wait db
+	uv run -m app.cli.import_data fetch-missing-distributions files.txt  missing_distributions ./out
+
 publish: build  ## Publish the Docker image to DockerHub
 	docker compose push app
 
--- a/app/cli/import_data.py
+++ b/app/cli/import_data.py
@@ -61,6 +61,7 @@ from app.db.model import (
     SingleNeuronSynaptome,
     SingleNeuronSynaptomeSimulation,
     Species,
+    ValidationResult,
 )
 from app.db.session import configure_database_session_manager
 from app.db.types import (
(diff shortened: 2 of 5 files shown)


app/cli/utils.py
```diff
@@ -1,9 +1,12 @@
 import datetime
+import os
 import uuid
 from datetime import timedelta
+from pathlib import Path
 from typing import Any, Literal
 
 import deepdiff
+import requests
```
mgeplf, line 9: httpx is generally what we use for doing http requests now; requests happens to be in the repo as a dependency of another package, but I think we should stick w/ httpx

app/cli/import_data.py
```diff
@@ -2040,6 +2081,75 @@ def organize_files(digest_path):
...
+        if sha256_digest in src_paths:
+            continue
+
+        content_url = distribution.get("contentUrl")
+        assert content_url
+        target_file = os.path.join(out_dir, sha256_digest)
+        L.info("missing content for distribution {}".format(distribution))
+        copied, token = utils.http_copy(token, content_url, target_file)
+        if not copied:
+            L.warning("failed to copy {}".format(distribution))
```
mgeplf, line 2146: please don't use `format`; you can use %s in the logging string

app/cli/utils.py
```diff
@@ -887,3 +897,49 @@ def get_or_create_admin(db, _cache={}):
...
+    data = {
+        "client_id": client_id,
+        "client_secret": client_secret,
+        "grant_type": "client_credentials",
+        "scope": "openid",
+    }
+    header = {
+        "Content-Type": "application/x-www-form-urlencoded",
+    }
+    response = requests.post(keycloak_url, data=data, headers=header, timeout=30)
```
mgeplf, line 923: is the timeout really needed?

Makefile
```diff
@@ -69,6 +69,11 @@ organize-files:  ## Organize files locally by creating symlinks from the backup
 	docker compose up --wait db
 	uv run -m app.cli.import_data organize-files $(if $(wildcard curated_files.txt),curated_files.txt,files.txt)
 
+fetch-missing-distributions:
```
eleftherioszisis, line 72: Could it be documented where in the order of commands for importing the data does this need to be executed?

app/cli/import_data.py
jdcourcol: done

app/cli/utils.py
jdcourcol: done.

app/cli/utils.py
mgeplf: huh, interesting, thanks for the heads up

app/cli/utils.py
mgeplf: did the change get pushed?

jdcourcol: ahem. I forgot to remove the import requests.

app/cli/import_data.py
```diff
@@ -1226,6 +1227,44 @@ def ingest(db, project_context, data_list, all_data_by_id, hierarchy_name: str):
...
+            legacy_self = data["@id"]
+            rm = utils._find_by_legacy_id(legacy_id, ValidationResult, db)
+            if rm:
+                continue
+
+            entity_id = data["entity_id"]
+            entity = utils._find_by_legacy_id(entity_id, Entity, db)
+
+            db_validation_report = ValidationResult(
+                legacy_self = [legacy_self],
```
jdcourcol, line 1252: I realize that I forgot to add the legacy_id. This had the side effect of not linking the asset to the validation report.

jdcourcol:
are we good now ?

mgeplf (approved):
I'm assuming you've tested this by uploading and seeing the results were fine.

Makefile
jdcourcol: it is there now : https://github.com/openbraininstitute/entitycore/blob/manual_deployment_instructions/docs/manual_deployment.md#missing-distributions

PR #244, opened 2025-06-23
add filter by `sub_id` and other filters for person

bilalesi:


app/filters/person.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Annotated
+
+from fastapi_filter import FilterDepends
+
+from app.db.model import Person
+from app.filters.base import CustomFilter
+from app.filters.common import CreatorFilterMixin
+
+
+class PersonFilter(CustomFilter, CreatorFilterMixin):
```
eleftherioszisis, line 11: Could you derive from AgentFilter. You could add there any additional pref_label filters keys that are missing.
https://github.com/openbraininstitute/entitycore/blob/205244a8a50b7e4305b34f1e5b0a932eeac78607/app/filters/common.py#L56

app/service/person.py
```diff
@@ -20,30 +26,54 @@ def _load(query: sa.Select):
...
-    ).scalars()
+    aliases: Aliases = {
+        Agent: {
+            "created_by": created_by_alias,
+            "updated_by": updated_by_alias,
+        }
+    }
 
-    total_items = db.execute(query.with_only_columns(sa.func.count(Person.id))).scalar_one()
+    facet_keys = filter_keys = [
```
mgeplf, line 45: is the `facet_keys` needed, since there are no facets?

app/service/person.py
```diff
@@ -20,30 +26,54 @@ def _load(query: sa.Select):
...
 
-    response = ListResponse[PersonRead](
-        data=[PersonRead.model_validate(d) for d in data],
-        pagination=PaginationResponse(
-            page=pagination_request.page,
-            page_size=pagination_request.page_size,
-            total_items=total_items,
-        ),
-        facets=None,
+    name_to_facet_query_params, filter_joins = query_params_factory(
```
mgeplf, line 50: `name_to_facet_query_params` isn't needed either, I think

PR #251, opened 2025-06-24
Add CellComposition read endpoint

eleftherioszisis:
Replace hardcoded static file to getting the entity and its assets.

### DIFF
--- a/app/filters/cell_composition.py
+++ b/app/filters/cell_composition.py
@@ -1,0 +1,25 @@
+from typing import Annotated
+
+from fastapi_filter import FilterDepends
+
+from app.db.model import CellComposition
+from app.filters.base import CustomFilter
+from app.filters.common import (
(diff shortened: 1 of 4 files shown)


bilalesi (approved):
LGTM

bilalesi:
so for cell composition, should we have the ID static too, so the FE use it as ENV ?

jdcourcol:
I discussed with Eleftherios, we will add the query end point end you will be able to query it by its name , like "release 2024".

I think we can remove all the hardcoded value:
- fetch the atlas named "BlueBrain Atlas",
- get the hierarchy out of it
- get the root brain region and default selected region ids by querying using their acronym
- finally get cell composition fetching it by name

eleftherioszisis:
I added a read_many endpoint to /cell-composition to allow searching it by name.

PR #252, opened 2025-06-24
Do not show deleted assets in entity metadata

eleftherioszisis:


app/db/model.py
```diff
@@ -462,9 +462,11 @@ class Entity(LegacyMixin, Identifiable):
     contributions: Mapped[list["Contribution"]] = relationship(uselist=True, viewonly=True)
     assets: Mapped[list["Asset"]] = relationship(
         "Asset",
-        foreign_keys="Asset.entity_id",
         uselist=True,
         viewonly=True,
+        primaryjoin=lambda: sa.and_(
+            Entity.id == Asset.entity_id, Asset.status != AssetStatus.DELETED
+        ),
```
GianlucaFicarelli, line 469: For confirmation, isn't it possible to avoid the lambda, and use only an expression as in some other primaryjoins?

GianlucaFicarelli (approved):

PR #253, opened 2025-06-25
Remove octet-stream content type because it is not needed/used

eleftherioszisis:
I added it because it was being set in get_content_type, however on second thought we shouldn't really allow it unless there is a use case that requires it.

### DIFF
--- a/app/utils/files.py
+++ b/app/utils/files.py
@@ -28,7 +28,7 @@ def get_content_type(file: UploadFile) -> ContentType:
             original_content_type,
             guessed_content_type,
         )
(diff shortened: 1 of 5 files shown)


app/utils/files.py
```diff
@@ -28,7 +28,7 @@ def get_content_type(file: UploadFile) -> ContentType:
             original_content_type,
             guessed_content_type,
         )
-    str_content_type = original_content_type or guessed_content_type or "application/octet-stream"
```
mgeplf, line 31: should there be an error if neither `original_content_type` or `guessed_content_type` work?

eleftherioszisis: Good point. I will add one

PR #254, opened 2025-06-25
Add prod agents

eleftherioszisis:


app/cli/curate.py
```diff
@@ -166,7 +166,7 @@ def curate_etype(data):
         )
     elif data["label"] == "TH_cNAD_noscltb":
         data["definition"] = (
-            "Thalamus continuous non-adapting non-oscillatory low-threshold bursting electrical type"  # noqa: E501
+            "Thalamus continuous non-adapting non-oscillatory low-threshold bursting electrical type"
```
GianlucaFicarelli, line 169: Minor, we could split the string instead of adding a noqa to the full file.

PR #260, opened 2025-06-26
Real required label

mgeplf:


app/schemas/asset.py
```diff
@@ -68,6 +65,29 @@ def ensure_entity_type_label_consistency(self):
             )
             raise ValueError(msg)
 
+        label_reqs = allowed_asset_labels[self.label]
```
eleftherioszisis, line 68: Shouldn't label: `AssetLabel | None = None` be made mandatory in AssetBase?

mgeplf: yes, good eye, thanks.

PR #262, opened 2025-06-27
add sim execution status to simulations within campaign

bilalesi:
this is a to have the activity (simulation_execution) status  within the simulation response, 
I had to add `used_by_activities` to the `Entity` model, but i don't know if it's fit the original and db arch you have in mind

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -468,6 +468,15 @@ class Entity(LegacyMixin, Identifiable):
             Entity.id == Asset.entity_id, Asset.status != AssetStatus.DELETED
         ),
     )
+    generated_by: Mapped[list["Activity"]] = relationship(
+        "Activity",
+        secondary="usage",
+        primaryjoin="Entity.id == usage.c.usage_entity_id",
+        secondaryjoin="Activity.id == usage.c.usage_activity_id",
+        uselist=True,
+        viewonly=True,
+        lazy="selectin",
+    )
 
     __mapper_args__ = {  # noqa: RUF012
         "polymorphic_identity": __tablename__,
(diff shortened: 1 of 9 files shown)


app/db/model.py
```diff
@@ -468,6 +468,15 @@ class Entity(LegacyMixin, Identifiable):
             Entity.id == Asset.entity_id, Asset.status != AssetStatus.DELETED
         ),
     )
+    used_by_activities: Mapped[list["Activity"]] = relationship(
```
eleftherioszisis, line 471: Could we maybe call this generated_by?
https://www.w3.org/TR/prov-o/diagrams/starting-points.svg

eleftherioszisis:
Why do we need the simulation_execution to be within the simulation response? Can't we keep activities linking entities and not the other way around? With the reusing of entities these joins can become extremely big...

app/db/model.py
bilalesi: from what i understand the 
SimulationCampaign (Entity) -> Simulation (Entity)
Simulation (Entity) -> Usage (*..*) <- Execution (Activity)

so simulation is used by an activity not generated by !?

eleftherioszisis:
> for your suggestion is there a way to query it ?

For each simulation you could query for a SimulationExecution that has a used__id equal to the simulation's one. Not optimal but it could get you what you want without extra code.

See: https://github.com/openbraininstitute/entitycore/blob/acea0cc0b1e07f22229192cf20b0feaeb511b69a/tests/test_simulation_execution.py#L219

app/db/model.py
eleftherioszisis: No, you were right. I mistakenly thought these are generations. I apologize for the confusion. It should be used_by

eleftherioszisis:
I have created an issue with a proposed solution in #264

PR #265, opened 2025-06-30
Use OBI as Admin pref_label

eleftherioszisis:


app/cli/utils.py
```diff
@@ -970,7 +970,7 @@ def get_or_create_admin(db, _cache={}):
 
     admin = Person(
         id=admin_id,
-        pref_label="Admin",
+        pref_label="OBI",
```
GianlucaFicarelli, line 973: Also the select query above should be changed (or even better, "OBI" can be moved to a variable or constant.

eleftherioszisis: All good points. Done.

PR #269, opened 2025-07-02
upgrade pydantic 2.10.6 -> 2.11.7

mgeplf:
* saves ~ %20 for `python -c 'from app.application import app'`, possibly due to https://github.com/pydantic/pydantic/issues/6768

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -703,41 +703,42 @@ wheels = [
 
 [[package]]
 name = "pydantic"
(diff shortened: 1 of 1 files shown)


GianlucaFicarelli:
Ok we can wait to be safer, and update all the dependencies after CNS.
We should also discuss and decide if/when we want to update the dependencies regularly, to avoid big updates.

PR #277, opened 2025-07-08
Update default order by in BrainAtlasRegionFilter

GianlucaFicarelli:


app/filters/brain_atlas.py
```diff
@@ -15,7 +15,7 @@ class Constants(CustomFilter.Constants):
 
 
 class BrainAtlasRegionFilter(IdFilterMixin, CustomFilter):
-    order_by: list[str] = ["id", "creation_date"]  # noqa: RUF012
+    order_by: list[str] = ["-creation_date"]  # noqa: RUF012
```
GianlucaFicarelli, line 18: Note: I don't have a strong preference between `creation_date` and `-creation_date`, but in all the other filters using `creation_date` as default, it's descending, so it seems more consistent.

mgeplf (approved):

PR #280, opened 2025-07-11
add counts route to entity module

bilalesi:
related issue: [80](https://github.com/openbraininstitute/prod-virtual-lab/issues/80)


app/service/entity.py
```diff
@@ -1,17 +1,42 @@
...
+from app.dependencies.db import SessionDep
 from app.repository.group import RepositoryGroup
 from app.schemas.auth import UserContext, UserContextWithProjectId
+from app.schemas.entity import EntityCountRead
 
 ENTITY_TYPE_TO_CLASS: dict[EntityType, type[Entity]] = {
     EntityType[mapper.class_.__tablename__]: mapper.class_
     for mapper in Base.registry.mappers
     if hasattr(EntityType, mapper.class_.__tablename__)
 }
```
eleftherioszisis, line 20: Could you maybe move this to app/db/utils.py perhaps?

app/service/entity.py
```diff
@@ -45,3 +70,70 @@ def get_writable_entity(
...
+    should_filter_brain_region = (
+        in_brain_region
+        and in_brain_region.within_brain_region_hierarchy_id is not None
+        and in_brain_region.within_brain_region_brain_region_id is not None
+    )
+
+    filter_conditions = []
+    if should_filter_brain_region:
+        for et in entity_types:
+            if et in ENTITY_TYPES_WITH_BRAIN_REGIONS:
```
eleftherioszisis, line 98: Isn't it surprising to allow specifying a type that isn't supported by ENTITY_TYPES_WITH_BRAIN_REGIONS and yet get a result that has nothing to do with said region?

bilalesi: yeah make sense, i create a derived enum, to only allow entities with brain region as query param and we can extended it later if we have use cases for that.

app/service/entity.py
```diff
@@ -45,3 +70,70 @@ def get_writable_entity(
...
+                entity_class = ENTITY_TYPE_TO_CLASS[et]
+                q = sa.select(
+                    sa.literal(et.value).label("type"),
+                    sa.func.count(entity_class.id).label("count"),
+                )
+                q = constrain_to_accessible_entities(
+                    q, project_id=user_context.project_id, db_model_class=entity_class
+                )
+                q = in_brain_region(q, entity_class)
```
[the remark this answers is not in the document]
bilalesi: yes, I noticed that in the logs, so I replaced it with a shared `cte` using `get_family_query`. 
the logic seems straightforward, the results are correct too

eleftherioszisis (approved):
lgtm

Issue #256, opened 2025-06-26
Asset directory on S3 not deleted when asset is deleted


mgeplf:
We mark the asset as deleted in the DB, but the s3 version isn't deleted, so if the name is reused, the contents of the directory, for instance, are the union of the old and new upload.

Single files are already deleted, see https://github.com/openbraininstitute/entitycore/blob/2c8a9e762158927a94edc1cc726beca421475caf/app/routers/asset.py#L210-L211

PR #281, opened 2025-07-11
Add admin delete endpoints for all routers

eleftherioszisis:
Add admin deletion endpoints via the /admin/ prefix

All foreign key references need to be deleted first before deleting the target resource.

**Note that admin deletion of an asset deletes both s3 file and DB entry.**

### DIFF
--- a/app/service/asset.py
+++ b/app/service/asset.py
@@ -188,7 +188,7 @@ def delete_asset(
 
 def delete_asset_storage_object(asset: AssetRead, storage_client_factory: StorageClientFactory):
     """Delete asset storage object."""
-    # TODO: Handle directories?
+    # TODO: Handle directories. See https://github.com/openbraininstitute/entitycore/issues/256
     storage = storages[asset.storage_type]
     # delete the file from S3 only if not using an open data storage
     if not storage.is_open:


app/routers/circuit.py
```diff
@@ -12,3 +12,4 @@
 read_many = router.get("")(app.service.circuit.read_many)
 read_one = router.get("/{id_}")(app.service.circuit.read_one)
 create_one = router.post("")(app.service.circuit.create_one)
+delete_one = router.delete("/{id_}")(app.service.circuit.delete_one)
```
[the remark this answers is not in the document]
eleftherioszisis: sigh, I forgot to make the change in the end...

app/routers/asset.py
```diff
@@ -287,6 +287,36 @@ def delete_entity_asset(
     return AssetRead.model_validate(asset)
 
 
+@router.delete("/admin/{entity_route}/{entity_id}/assets/{asset_id}")
+def admin_delete_entity_asset(
```
eleftherioszisis, line 291: @GianlucaFicarelli would it be preferable if I moved this in the admin router instead?

app/utils/enum.py
```diff
@@ -25,3 +27,9 @@ class UpperStrEnum(StrEnum):
     def _generate_next_value_(name: str, start: int, count: int, last_values: list[str]) -> str:  # noqa: ARG004
         """Return the upper-cased version of the member name."""
         return name.upper()
+
+
+def combine_str_enums(name: str, enums: Sequence[type[StrEnum]]) -> type[StrEnum]:
+    """Combine mutliple StrEnum into a single one."""
+    members = {e.name: e.value for enum in enums for e in enum.__members__.values()}
```
GianlucaFicarelli, line 34: For confirmation, the generated enum is not sorted, but we keep the order of the concatenated enums. I think it should be fine, as long as the order is not randomly changing (because it would generate different json schemas).

app/db/types.py
```diff
@@ -69,7 +71,6 @@ class EntityType(StrEnum):
     mesh = auto()
```
GianlucaFicarelli, line 71: Unrelated to this PR, but I think that `mesh` isn't used anymore.

app/routers/asset.py
```diff
@@ -287,6 +287,36 @@ def delete_entity_asset(
...
+    repos: RepoGroupDep,
+    user_context: AdminContextDep,  # noqa: ARG001
+    storage_client_factory: StorageClientFactoryDep,
+    entity_route: EntityRoute,
+    entity_id: uuid.UUID,
+    asset_id: uuid.UUID,
+) -> AssetRead:
+    """Delete an assets associated with a specific entity.
+
+    The asset record is not deleted from the database, but its status is changed.
```
GianlucaFicarelli, line 301: The docstring should be different from the other endpoint (the record is actually deleted)

app/service/asset.py
```diff
@@ -155,6 +155,22 @@ def delete_entity_asset(
     return AssetRead.model_validate(asset)
 
 
+def remove_entity_asset(
```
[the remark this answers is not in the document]
eleftherioszisis: Is it better now?

app/db/utils.py
```diff
@@ -22,6 +22,12 @@
     if hasattr(EntityType, mapper.class_.__tablename__)
 }
 
+RESOURCE_TYPE_TO_CLASS: dict[str, type[Identifiable]] = {
+    mapper.class_.__tablename__: mapper.class_
+    for mapper in Base.registry.mappers
+    if mapper.class_.__tablename__ in set(ResourceType)
```
GianlucaFicarelli, line 28: Minor, but `set(ResourceType)` is called for each iteration, so it's not better than using `in ResourceType`

app/service/asset.py
```diff
@@ -137,24 +139,64 @@ def delete_entity_asset(
...
+                entity_id=entity_id,
+                asset_id=asset_id,
+                asset_status=AssetStatus.DELETED,
+            )
+    return AssetRead.model_validate(asset)
+
+
+def delete_asset_storage_object(asset: AssetRead, storage_client_factory: StorageClientFactory):
+    """Delete asset storage object."""
+    # TODO: Handle directories?
```
GianlucaFicarelli, line 191: We could link the ticket https://github.com/openbraininstitute/entitycore/issues/256

app/db/utils.py
eleftherioszisis: Removed it.

app/service/asset.py
eleftherioszisis: Added it

PR #282, opened 2025-07-14
Morphology model

dkeller9:
This branch includes a minimal set of commits to illustrate the data model, and is branched off of a more recent main. The primary files affected are: model.py, types.py, and morphology.py (although this is intended just to give context to the data model). "ReconstructionCellMorphology" is changed to "CellMorphology" because now the reconstruction morphologies are a subtype of cell morphology. Because the different sub-clases of morphology were not made programmatic subclasses of cell morphology for the sake of maintaining compatibility with how this structure is processed, the extra data needed for the subclasses is in the "extended_data" field. A text description of all the different morphology components is in: https://openbraininstitute.sharepoint.com/:w:/r/sites/OBI-Scientificstaff/_layouts/15/Doc.aspx?sourcedoc=%7BEEE86C08-A09D-469E-9849-CD85599E4026%7D&file=General_schema.docx&action=default&mobileredirect=true&DefaultItemOpen=1

### DIFF
--- a/app/db/types.py
+++ b/app/db/types.py
@@ -275,6 +275,7 @@ class AssetLabel(StrEnum):
     """See docs/asset-labels.md."""
 
     morphology = auto()
+    morphology_with_spines = auto()
     cell_composition_summary = auto()
     cell_composition_volumes = auto()
     single_neuron_synaptome_config = auto()


app/db/types.py
```diff
@@ -7,6 +7,16 @@
...
 
+from app.utils.enum import StrEnum
+
+
+class MorphologyStructureType(str, Enum):
+    digital_reconstruction = auto()
+    modified = auto()
+    computationally_synthesized = auto()
+    placeholder = auto()
+    generic = auto()  # for base CellMorphology
```
eleftherioszisis, line 18: base morphology?

app/schemas/morphology.py
```diff
@@ -19,29 +20,85 @@
 from app.schemas.contribution import ContributionReadWithoutEntityMixin
 from app.schemas.measurement_annotation import MeasurementAnnotationRead
 from app.schemas.species import NestedSpeciesRead, NestedStrainRead
+from app.schemas.types import PipelineType
 
 
-class ReconstructionMorphologyBase(BaseModel):
+class Protocol(BaseModel):
```
eleftherioszisis, line 26: This looks like it should be a field below instead of a base class.

app/schemas/morphology.py
```diff
@@ -19,29 +20,85 @@
...
+        Amount tissue shrunk by (not correction factor)
+    has_been_corrected_for_shrinkage: bool, optional
+        Whether data has been corrected for shrinkage
+    """
+
+    protocol_design: str
+    staining_method: str
+    slicing_thickness: float
+
+    slicing_direction: str | None = None
```
eleftherioszisis, line 62: Couldn't we constrain this with a StrEnum too?

app/schemas/morphology.py
```diff
@@ -19,29 +20,85 @@
...
+    magnification: float, optional
+        Magnification level used
+    tissue_shrinkage: float, optional
+        Amount tissue shrunk by (not correction factor)
+    has_been_corrected_for_shrinkage: bool, optional
+        Whether data has been corrected for shrinkage
+    """
+
+    protocol_design: str
+    staining_method: str
```
eleftherioszisis, line 59: This should be constrained with a StrEnum

app/schemas/morphology.py
```diff
@@ -19,29 +20,85 @@
...
+        Direction of slicing
+    magnification: float, optional
+        Magnification level used
+    tissue_shrinkage: float, optional
+        Amount tissue shrunk by (not correction factor)
+    has_been_corrected_for_shrinkage: bool, optional
+        Whether data has been corrected for shrinkage
+    """
+
+    protocol_design: str
```
eleftherioszisis, line 58: This should never constrained with a StrEnum

app/db/model.py
```diff
@@ -667,7 +667,25 @@ def measurement_annotation(cls):
         )
 
 
-class ReconstructionMorphology(
+class CellMorphologyMetadata(Base):
```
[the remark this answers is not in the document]
dkeller9: This is due to the polymorphic nature of the subclasses of CellMorphology. They could also all be optional fields of the main table, but I thought it was better to separate them to show the distinctions between the classes.

app/db/model.py
```diff
@@ -667,7 +667,25 @@ def measurement_annotation(cls):
...
+    __tablename__ = "cell_morphology_metadata"
+    id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=create_uuid)
+    cell_morphology_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("cell_morphology.id"), index=True, unique=True
+    )
+    method_description: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
+    pipeline_state: Mapped[PipelineType | None] = mapped_column(
```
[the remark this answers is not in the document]
dkeller9: fixed

app/db/types.py
dkeller9: If there are no other fields needed as for the additional types. Will remove this comment as it is confusing.

app/db/model.py
```diff
@@ -667,7 +667,25 @@ def measurement_annotation(cls):
...
+        Enum(PipelineType, name="pipeline_type"), nullable=True
+    )
+    is_related_to: Mapped[list[uuid.UUID] | None] = mapped_column(
```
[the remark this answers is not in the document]
dkeller9: Yes, perhaps this could be handled through the is_derived_from source and target table we discussed previously, in which case the is_related_field could be removed. But even if that is done, deletion of an id means that it would still need to be removed from the relationship table.

app/db/model.py
```diff
@@ -667,7 +667,25 @@ def measurement_annotation(cls):
...
+        ARRAY(UUID(as_uuid=True)), nullable=True
+    )
+    score_dict: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
```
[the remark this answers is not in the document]
dkeller9: I think it is a validation activity. @lidakanari I wonder if this should be stored as a validation result instead?

app/db/model.py
```diff
@@ -667,7 +666,26 @@ def measurement_annotation(cls):
...
+    provenance: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
+    cell_morphology = relationship("CellMorphology", back_populates="extended_data")
+
+
+class CellMorphology(
```
jdcourcol, line 688: My understanding is that it should derive from ScientificArtefact .

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: yes, it looks to me like a validation report. This is a pattern we already have for me-models

app/db/model.py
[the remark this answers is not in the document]
jdcourcol: I am not sure I understand. I read the document but I don't know what the relationship is . Is it a derivation ?

app/db/model.py
dkeller9: Now derived from ScientificArtifact.

app/schemas/morphology.py
```diff
@@ -57,5 +114,69 @@ class ReconstructionMorphologyRead(
...
+
+class ScoreDict(BaseModel):
+    x: dict[str, float]
+
+
+class PipelineType(str, Enum):
+    Raw = "Raw"
+    Curated = "Curated"
+    Unraveled = "Unraveled"
+    Repaired = "Repaired"
```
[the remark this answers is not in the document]
dkeller9: fixed

app/db/model.py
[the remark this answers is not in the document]
dkeller9: The reason for doing it this way was to avoid breaking the existing endpoints and code by introducing new subclasses of morphologies. But yes I agree that it would be more straightforward to just have the other classes outright. I will modify accordingly.

app/schemas/morphology.py
dkeller9: It was requested to make Protocol a base class of the experimental morphology method.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: I have removed the CellMorphologyMetadata table after consulting Eleftherios.

app/schemas/morphology.py
```diff
@@ -57,5 +114,69 @@ class ReconstructionMorphologyRead(
...
+    Ratified = auto()
+
+
+class DigitalReconstruction(CellMorphologyRead):
+    reconstruction_method: ExperimentalMorphologyMethod
+    pipeline_state: PipelineType
+    is_related_to: list[uuid.UUID]
+
+
+class DigitalReconstructionCreate(CellMorphologyCreate):
```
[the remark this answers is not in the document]
dkeller9: I will defer to the engineers on this one, the purpose of including the schema is just to help illustrate how the data model should be perceived.

app/db/model.py
```diff
@@ -679,7 +697,14 @@ class ReconstructionMorphology(
 
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
     location: Mapped[PointLocation | None]
-
+    structure_type = mapped_column(
+        Enum(MorphologyStructureType, name="morphologystructuretype"),
+        nullable=False,
+        default=MorphologyStructureType.generic,
```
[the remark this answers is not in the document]
dkeller9: I am having trouble getting it to accept a simpler syntax.

app/db/model.py
```diff
@@ -666,21 +668,114 @@ def measurement_annotation(cls):
...
+    # Relationship to the polymorphic MorphologyProtocol
+    morphology_method: Mapped["MorphologyProtocol"] = relationship(
+        "MorphologyProtocol",
+        foreign_keys=[morphology_method_id],
+        uselist=False,
+    )
+
+    #structured jsonb , pydantic typedict
+    # Attributes from DigitalReconstruction
+    pipeline_state: Mapped[PipelineType | None] = mapped_column(
```
jdcourcol, line 760: I think we should be more specific than "pipeline". It is blue brain repair pipeline. So may be "repair_pipeline".

app/db/model.py
```diff
@@ -666,21 +668,114 @@ def measurement_annotation(cls):
...
+        Enum(PipelineType, name="pipelinetype"), 
+        nullable=True,
+    )
+ 
+    # Attributes from ComputationallySynthesized
+
+    provenance: Mapped[str]
+#cab it be structured, maybe renamed/?
+    # Attribute from Placeholder and DigitalReconstruction
+    is_related_to: Mapped[list[uuid.UUID] | None] = mapped_column(
```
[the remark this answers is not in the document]
dkeller9: Yes, removed.

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+
+class ExperimentalMorphologyMethod(MorphologyProtocol):
+    __tablename__ = "experimental_morphology_protocol"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("morphology_protocol.id"), primary_key=True)
+    staining_method: Mapped[StainingType| None] 
+    slicing_thickness: Mapped[float]
+    slicing_direction: Mapped[SlicingDirectionType | None]
+    magnification: Mapped[float | None]
+    tissue_shrinkage: Mapped[float | None]
+    has_been_corrected_for_shrinkage: Mapped[bool | None]
```
jdcourcol, line 705: "has_been_corrected_for_shrinkage" -> "corrected_for_shrinkage"

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+    protocol_document: Mapped[str | None]
+    protocol_design: Mapped[str]
+    type: Mapped[str]  # Discriminator column for polymorphism
+
+    __mapper_args__ = {
+        "polymorphic_identity": __tablename__,
+        "polymorphic_on": "type",
+    }
+
+class ExperimentalMorphologyMethod(MorphologyProtocol):
```
jdcourcol, line 697: Method->Protocol ?

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("morphology_protocol.id"), primary_key=True)
+    staining_method: Mapped[StainingType| None] 
+    slicing_thickness: Mapped[float]
+    slicing_direction: Mapped[SlicingDirectionType | None]
+    magnification: Mapped[float | None]
+    tissue_shrinkage: Mapped[float | None]
+    has_been_corrected_for_shrinkage: Mapped[bool | None]
+
+    __mapper_args__ = {
+        "polymorphic_identity": "experimental", # A string identifier for this type
```
jdcourcol, line 708: "polymorphic_identity": "experimental" -> "polymorphic_identity": __tablename__

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+            foreign_keys="MeasurementAnnotation.entity_id",
+            uselist=False,
+            viewonly=True,
+        )
+    
+class MorphologyProtocol(Identifiable): # Inherit from Identifiable for primary key and timestamps
+    __tablename__ = "morphology_protocol"
+    protocol_document: Mapped[str | None]
+    protocol_design: Mapped[str]
+    type: Mapped[str]  # Discriminator column for polymorphism
```
jdcourcol, line 690: do we need that ? @GianlucaFicarelli  ?

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+        return relationship(
+            "MeasurementAnnotation",
+            foreign_keys="MeasurementAnnotation.entity_id",
+            uselist=False,
+            viewonly=True,
+        )
+    
+class MorphologyProtocol(Identifiable): # Inherit from Identifiable for primary key and timestamps
+    __tablename__ = "morphology_protocol"
+    protocol_document: Mapped[str | None]
```
jdcourcol, line 688: I don't understand what that is ? a URL ? a DOI ?

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+
+    __mapper_args__ = {
+        "polymorphic_identity": __tablename__,
+        "polymorphic_on": "type",
+    }
+
+class ExperimentalMorphologyMethod(MorphologyProtocol):
+    __tablename__ = "experimental_morphology_protocol"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("morphology_protocol.id"), primary_key=True)
+    staining_method: Mapped[StainingType| None]
```
jdcourcol, line 700: staining_method -> staining_type ? since that is the name of this enum

app/db/model.py
```diff
@@ -666,21 +668,114 @@ def measurement_annotation(cls):
...
+    #structured jsonb , pydantic typedict
+    # Attributes from DigitalReconstruction
+    pipeline_state: Mapped[PipelineType | None] = mapped_column(
+        Enum(PipelineType, name="pipelinetype"), 
+        nullable=True,
+    )
+ 
+    # Attributes from ComputationallySynthesized
+
+    provenance: Mapped[str]
```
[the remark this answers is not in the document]
dkeller9: We need feedback from @eleftherioszisis on what this should be. Probably a JSON to represent a dictionary.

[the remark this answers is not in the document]
jdcourcol: then I don't understand why this is not the input of an activity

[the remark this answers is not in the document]
dkeller9: If indeed it is tracked in activity, then we can remove it. Can you confirm @eleftherioszisis ?

app/db/model.py
dkeller9: Can't this identifier be user-defined as has been done here?

app/db/model.py
```diff
@@ -666,21 +668,110 @@ def measurement_annotation(cls):
...
+    method: Mapped[str] # This 'method' might need further typing based on your vocabulary
+
+    __mapper_args__ = {
+        "polymorphic_identity": "computationally_synthesized",
+    }
+
+class ModifiedMorphologyMethod(MorphologyProtocol):
+    __tablename__ = "modified_morphology_protocol"
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("morphology_protocol.id"), primary_key=True)
+    method: Mapped[MethodsType] # Assuming MethodsType is an Enum or similar
```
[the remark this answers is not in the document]
dkeller9: Yes, it is polymorphic because the methods can vary considerably.

app/db/model.py
eleftherioszisis: You need to add "polymorphic_on": "type" which is the intended column to be used here for sqlalchemy to decide which hierarchy of tables to update when an operation is made.

In addition, in MorphologyProtocol the type must be constrained by a MorphologyProtocolType StreEnum the values of which should be the children tables, e.g.

```python
class MorphologyProtocolType(StrEnum):
    experimental_morphology_protocol = auto()
```

and then in MorphologyProtocol `type: Mapped[MorphologyProtocolType]`, and in the children tables:

```python
class ExperimentalMorphologyProtocol(MorphologyProtocol):
     __tablename__ = MorphologyProtocolType.experimental_morphology_protocol.value

     ....

    __mapper_args__ = {
         "polymorphic_on": "type"
         "polymorphic_identity": __tablename__



    
```

app/db/model.py
```diff
@@ -667,7 +667,25 @@ def measurement_annotation(cls):
...
+    )
+    method_description: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
+    pipeline_state: Mapped[PipelineType | None] = mapped_column(
+        Enum(PipelineType, name="pipeline_type"), nullable=True
+    )
+    is_related_to: Mapped[list[uuid.UUID] | None] = mapped_column(
+        ARRAY(UUID(as_uuid=True)), nullable=True
+    )
+    score_dict: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
+    provenance: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
```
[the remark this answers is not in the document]
dkeller9: These aspects of provenance are already handled by the publication linkage table and the activity table, in the case of derived morphologies. I think what was originally meant by "provenance" as noted in the General Schema was the synthesis parameters. But apparently these are also handled in a separate table.

app/db/model.py
dkeller9: This is fixed in the latest push.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: Okay, removed.

app/schemas/morphology.py
dkeller9: I added a list of staining options. Can be augmented as needed.

app/db/model.py
```diff
@@ -666,21 +668,112 @@ def measurement_annotation(cls):
...
     __tablename__ = EntityType.reconstruction_morphology.value
 
-    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scientific_artifact.id"), primary_key=True)
+    
     location: Mapped[PointLocation | None]
+    generation_type: Mapped[MorphologyGenerationType] = mapped_column(
+        Enum(MorphologyGenerationType, name="morphologygenerationtype"),
+        nullable=False,
+        default=MorphologyGenerationType.placeholder,
```
eleftherioszisis, line 745: IMO this should be nullable. Registering morphologies as placeholders by default doesn't sound like the right default behavior.

app/db/model.py
```diff
@@ -666,21 +668,112 @@ def measurement_annotation(cls):
...
+    )
 
-    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+    # New foreign key for the morphology method
+    morphology_protocol_id: Mapped[uuid.UUID | None] = mapped_column(
+        ForeignKey("morphology_protocol.id"), nullable=True, index=True
+    )
+    # Relationship to the polymorphic MorphologyProtocol
+    morphology_protocol: Mapped["MorphologyProtocol"] = relationship(
```
eleftherioszisis, line 755: Given that the morphology protocol carries information about the generation_type, do we really need it as a separate column?

dkeller9: @eleftherioszisis are you proposing to remove the line "morphology_protocol: Mapped["MorphologyProtocol"] = relationship(..." ?

dkeller9: The only issue I can see with that is that placeholder might not have a protocol.

eleftherioszisis: Thanks. Let me clean and get the PR in shape first.

GianlucaFicarelli:
I think that the latest PR based on this branch is https://github.com/openbraininstitute/entitycore/pull/333

PR #285, opened 2025-07-14
Add gltf-binary type for mesh assets

### LINKED ISSUE #284: Adding a new type of brain region meshes
Currently, we are using `*.obj` meshes to display the brain regions in the explorer.
The format is quite old and text-based.

We can convert these files into [`*.glb`](https://en.wikipedia.org/wiki/GlTF) with DRACO compression which will reduce the size by 94%:
the current size of all the `*.obj` files is __372 Mb__ and the size for `*.glb` files is only __15 Mb__.

The mimetype of the GLB files is `model/gltf-binary`.

Would it be possible to have a storage for this new data? And a new entry in entitycore?

eleftherioszisis:
Closes #284


tolokoban (approved):
LGTM

PR #287, opened 2025-07-15
ElectricalCellRecording is now inheriting from ScientificArtifact

AurelienJaquier:
The three new fields that are added by scientific artifact are nullable (experiment_data, contact_id, atlas_id). We don’t need to rename or delete any columns. We already were using subject in ElectricalCellRecording, so this is not changing by using ScientificArtifact. Two fields of ElectricalCellRecording were nullable before, but are no longer nullable when using ScientificArtifact. Those fields are `assets` in ElectricalCellRecordingRead and `brain_region_id` in ElectricalCellRecordingCreate.


AurelienJaquier:
> The db model class needs to be reparented too in app/db/model.py. Same for the filters in app/filters/electrical_cell_recording.py

Ok, I have updated model and filters. Please have a look and check

eleftherioszisis:
Applying the migration I made above due to the change to a local db created from the staging dump gets:
```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "electrical_cell_recording" violates foreign key constraint "fk_electrical_cell_recording_id_scientific_artifact"
DETAIL:  Key (id)=(83273ecc-10dc-415c-b3da-c3b8e8d36d50) is not present in table "scientific_artifact".

[SQL: ALTER TABLE electrical_cell_recording ADD CONSTRAINT fk_electrical_cell_recording_id_scientific_artifact FOREIGN KEY(id) REFERENCES scientific_artifact (id)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```
This will be fun :)

eleftherioszisis:
Yeah, sure.

app/filters/electrical_cell_recording.py
```diff
@@ -4,19 +4,15 @@
...
-    EntityFilterMixin,
     ETypeClassFilterMixin,
     NameFilterMixin,
 )
-from app.filters.subject import SubjectFilterMixin
+from app.filters.scientific_artifact import ScientificArtifactFilter
 
 
 class ElectricalCellRecordingFilter(
     CustomFilter,
```
eleftherioszisis, line 14: Remove CustomFilter. You have it in the scientific artifact parent.

eleftherioszisis:
I'll let @GianlucaFicarelli take a look.

alembic/versions/20250718_151353_5307e375e19f_ephys_parent_scientific_artifact.py
```diff
@@ -0,0 +1,205 @@
...
+    conn = op.get_bind()
+
+    op.add_column(
+        "electrical_cell_recording",
+        sa.Column("subject_id", sa.UUID(), autoincrement=False, nullable=True),
+    )
+    op.add_column(
+        "electrical_cell_recording",
+        sa.Column("brain_region_id", sa.UUID(), autoincrement=False, nullable=False),
+    )
```
GianlucaFicarelli, line 126: This fails during a potential downgrade with:
> sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) column "brain_region_id" of relation "electrical_cell_recording" contains null values

alembic/versions/20250718_151353_5307e375e19f_ephys_parent_scientific_artifact.py
```diff
@@ -0,0 +1,205 @@
...
+    missing_rows = conn.execute(
+        sa.select(
+            electrical_cell_recording.c.id,
+            electrical_cell_recording.c.subject_id,
+            electrical_cell_recording.c.brain_region_id,
+            electrical_cell_recording.c.license_id,
+        ).where(~electrical_cell_recording.c.id.in_(sa.select(scientific_artifact.c.id)))
+    ).fetchall()
+
+    if missing_rows:
```
GianlucaFicarelli, line 57: If there are many records, it should be more performant to execute a single `insert into ... select ...` without retrieving the records, but no need to change the query in this case.

alembic/versions/20250718_151353_5307e375e19f_ephys_parent_scientific_artifact.py
```diff
@@ -0,0 +1,205 @@
+"""Default migration message
```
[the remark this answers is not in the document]
eleftherioszisis: Updated

alembic/versions/20250718_151353_5307e375e19f_ephys_parent_scientific_artifact.py
```diff
@@ -0,0 +1,205 @@
...
+    "scientific_artifact",
+    sa.column("id", sa.UUID()),
+    sa.column("subject_id", sa.UUID()),
+    sa.column("brain_region_id", sa.UUID()),
+    sa.column("license_id", sa.UUID()),
+)
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
eleftherioszisis: Removed

alembic/versions/20250718_151353_5307e375e19f_ephys_parent_scientific_artifact.py
eleftherioszisis: Ah indeed. Fixed it.

PR #296, opened 2025-07-21
Updated ScientificArtifact for switching circuits explore front end to entitycore

chr-pok:
Updates related to switching circuits explore front end to entitycore:
- In ScientificArtifact: https://github.com/openbraininstitute/prod-explore-functionality/issues/235
  - Added `published_in` (optional short version string of the publication)
  - Changed `contact_id` to `contact_email` (string instead of relation)


dkeller9 (commented):
Yes, the PR is consistent with what we discussed.

danielaegassan (commented):
Looks good to me

PR #297, opened 2025-07-21
Updated Publication for switching circuits explore front end to entitycore

chr-pok:
Update related to switching circuits explore front end to entitycore:
- In Publication: https://github.com/openbraininstitute/prod-explore-functionality/issues/235
  - Added `publication_month` (optional month of the publication) together with a new enum Month

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -52,6 +52,7 @@ from app.db.types import (
     EntityType,
     MeasurementStatistic,
     MeasurementUnit,
+    Month,
     PointLocation,
     PointLocationType,
     Sex,
@@ -510,6 +511,7 @@ class Publication(Entity, NameDescriptionVectorMixin):
         title (str | None): Title of the publication.
         authors (list[Author] | None): List of authors associated with the publication.
         publication_year (int | None): Year the publication was released.
+        publication_month (Month | None): Month the publication was released.
         abstract (str | None): Abstract or summary of the publication.
 
     """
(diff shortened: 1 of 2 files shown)


chr-pok:
Hey @danielaegassan and @dkeller9! Before I will ask for a technical review, could you please have a look if this change is what we had in mind from the scientific perspective? Thanks!

app/db/model.py
```diff
@@ -510,6 +511,7 @@ class Publication(Entity, NameDescriptionVectorMixin):
         title (str | None): Title of the publication.
         authors (list[Author] | None): List of authors associated with the publication.
         publication_year (int | None): Year the publication was released.
+        publication_month (Month | None): Month the publication was released.
```
eleftherioszisis, line 514: Why not use datetime instead? Is it necessary to have two columns for year and month?

chr-pok: Ok I see. So in general, we can't sort on a combination of columns? I think for year/month, this won't be required. Or is there any easy and clean alternative with month (and day) being optional?

danielaegassan: If this bring trouble, the suggestion was to just keep year.

eleftherioszisis: Yeah, sorting on combination of columns is not supported. Given that datetime is already supported by sqlalchemy I would just go with setting month/day to 1 if not available..

chr-pok:
Closing this PR without merging, since this is causing more troubles than actually adding something useful.

PR #300, opened 2025-07-22
IonChannelModel inheriting from ScientificArtifact

AurelienJaquier:
Note: This PR expects changes made in #287 , so we won't be able to merge this before #287 is merged and we have rebased on it.

- add `channelpedia_link` field to `IonChannelModel`
- make `IonChannelModel` inherit from Scientific artifact.
  - This adds a bunch of nullable fields to it: `experiment_date`, `contact_id`, `atlas_id`, and `license_id`.
  - It also replaces `SpeciesMixin` inheritance with `SubjectMixin` inheritance, meaning `IonChannelModel` now has `subject_id` field instead of `species_id`. This will have to be handled somehow.

Partially solves https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/14


app/db/model.py
```diff
@@ -949,17 +949,18 @@ def _normalize_name(self, key, value):  # noqa: PLR6301, ARG002
...
-    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scientific_artifact.id"), primary_key=True)
 
     is_ljp_corrected: Mapped[bool] = mapped_column(default=False)
     is_temperature_dependent: Mapped[bool] = mapped_column(default=False)
     temperature_celsius: Mapped[int]
     is_stochastic: Mapped[bool] = mapped_column(default=False)
     nmodl_suffix: Mapped[str]
     neuron_block: Mapped[JSON_DICT]
+    channelpedia_link: Mapped[str | None] = mapped_column(String, nullable=True, default=None)
```
eleftherioszisis, line 963: Isn't it too specific though? What if we want to link a different source?

AurelienJaquier:
Based on what was done for `Publication` and `ScientificArtifactPublicationLink`, I created `ExternalDatabase`, `ExternalDatabaseURL` and `ScientificArtifactExternalDatabaseURLLink` to make the link between ion channel model and the channelpedia (or other database) URL

AurelienJaquier:
This PR is good from my point of view. If it looks good also to you guys, could you help me with the tests?

app/db/model.py
```diff
@@ -540,6 +540,48 @@ class Publication(Entity, NameDescriptionVectorMixin):
...
+    __tablename__ = EntityType.external_data_source_page.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    external_data_source_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("external_data_source.id"), index=True, nullable=False
+    )
+    URL: Mapped[str] = mapped_column(String, index=True)
+
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+    }
```
GianlucaFicarelli, line 582: It's not wrong to normalize `ExternalDataSource` into `ExternalDataSource` and `ExternalDataSourcePage`, but every additional table adds some complexity to the model (one more record to insert when inserting, one more table to join when reading).
Are there use cases that would benefit from this separation?
Do we really need to store the information into two tables?

app/service/emodel.py
```diff
@@ -40,8 +40,7 @@ def _load(select: sa.Select[tuple[EModel]]):
         selectinload(EModel.contributions).joinedload(Contribution.role),
         joinedload(EModel.mtypes),
         joinedload(EModel.etypes),
-        selectinload(EModel.ion_channel_models).joinedload(IonChannelModel.species),
-        selectinload(EModel.ion_channel_models).joinedload(IonChannelModel.strain),
+        selectinload(EModel.ion_channel_models).joinedload(IonChannelModel.subject),
```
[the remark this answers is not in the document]
AurelienJaquier: I thought I had to do it because of the joinedload with IonChannelModel. Should I revert that change?

app/db/model.py
GianlucaFicarelli: The additional table is clean and allows to register more sources, so I'm not strongly against.
On the other side, a `source_label` column is simpler, but if it's not validated against a list of valid values (as an enum), it would allow to register any incorrect data.

Clarifying the use cases can be helpful to decide.
- How many sources do we expect?
- How often will we need to add new ones?
- Who will be allowed to add sources?
- Who will be allowed to add pages?

Let's see other opinions as well.

GianlucaFicarelli:
I'm going to split the changes in this PR in 2 parts, because they are not strictly related and they should be simpler to complete and review:
- IonChannelModel inheriting from ScientificArtifact: need also to add the migration of schema and existing data, as done in https://github.com/openbraininstitute/entitycore/pull/287
- Add External Data Source to ScientificArtifact: need also to add endpoint to create the link between the scientific artifact and the external data source, similar to publications as done in https://github.com/openbraininstitute/entitycore/pull/301

app/schemas/ion_channel_model.py
```diff
@@ -43,25 +37,32 @@ class IonChannelModelBase(BaseModel):
...
+    IonChannelModelRead,
+    AssetsMixin,
+):
     pass
 
 
 class IonChannelModelExpanded(
-    IonChannelModelWAssets, CreatedByUpdatedByMixin, ContributionReadWithoutEntityMixin
+    IonChannelModelBase,
+    ScientificArtifactRead,
```
GianlucaFicarelli, line 66: The 3 read schemas IonChannelModelRead, IonChannelModelWAssets, IonChannelModelExpanded are kept for backward compatibility (and maybe performance reasons).
iirc there was a discussion to unify them.

PR #316, opened 2025-07-30
Allow filtering electrical_cell_recording by recording_origin and recording_type

GianlucaFicarelli:
Fix https://github.com/openbraininstitute/prod-singlecell-simulation/issues/182

### DIFF
--- a/tests/test_electrical_cell_recording.py
+++ b/tests/test_electrical_cell_recording.py
@@ -12,7 +12,7 @@ from app.db.model import (
     Species,
     Subject,
 )
-from app.db.types import EntityType
+from app.db.types import ElectricalRecordingOrigin, ElectricalRecordingType, EntityType
 
 from .utils import (
     PROJECT_ID,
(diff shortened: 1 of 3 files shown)


tests/test_electrical_cell_recording.py
```diff
@@ -338,6 +338,58 @@ def test_filtering(client, models):
...
+        params={"recording_type": ElectricalRecordingType.intracellular},
+    ).json()["data"]
+    assert len(data) == len(recordings)
+
+    data = assert_request(
+        client.get,
+        url=ROUTE,
+        params={"recording_type": ElectricalRecordingType.extracellular},
+    ).json()["data"]
+    assert len(data) == 0
```
mgeplf, line 353: i find it a bit scary that all the recordings of the same type; although not enough to ask you to change it now

bilalesi (approved):

tests/test_electrical_cell_recording.py
GianlucaFicarelli: I can update the models created for tests, it shouldn't break other tests.

mgeplf (approved):

PR #319, opened 2025-07-31
Add dump and load scripts

GianlucaFicarelli:
Fix https://github.com/openbraininstitute/INFRA/issues/231

Only the public data are dumped.

See the README for instructions.

### DIFF
--- a/scripts/export/build_database_archive.sh
+++ b/scripts/export/build_database_archive.sh
@@ -2,7 +2,7 @@
 # Automatically generated, do not edit!
 set -euo pipefail
 SCRIPT_VERSION="1"
-SCRIPT_DB_VERSION="805fc8028f39"
+SCRIPT_DB_VERSION="8072d90499ad"
 echo "DB dump (version $SCRIPT_VERSION for db version $SCRIPT_DB_VERSION)"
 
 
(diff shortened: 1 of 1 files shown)


scripts/export/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+After any db migration, the script `build_database_archive.sh` should be updated by running:
+
+```
+uv run ./scripts/export/write_scripts.py
+```
+
+### Export the public data from the database
+
+The script `build_database_archive.sh` requires access to read the database, and it builds a self-extracting archive named `install_db_{YYYYMMDD}_{DB_VERSION}.run` in the working directory.
```
mgeplf: Very fancy!

I'm glad there is `install_db...run --noexec --keep`

scripts/export/write_scripts.py
```diff
@@ -0,0 +1,514 @@
...
+    mapper.class_.__tablename__: mapper
+    for mapper in Base.registry.mappers
+    if mapper.class_.__tablename__
+}
+BUILD_SCRIPT = "build_database_archive.sh"
+LOAD_SCRIPT = "load.sh"
+
+SCRIPT_VERSION = 1
+SETUP_PSQL = """
+export PATH="/opt/homebrew/opt/postgresql@17/bin:$PATH"
```
mgeplf: instead of having both the mac and linux on the PATH, perhaps test which OS it is w/ `uname`?

scripts/export/write_scripts.py
```diff
@@ -0,0 +1,514 @@
...
+    script = ScriptDirectory.from_config(config)
+    if not (head := script.get_current_head()):
+        msg = "Head revision not found"
+        raise RuntimeError(msg)
+    L.info("Alembic head revision: %s", head)
+    return head
+
+
+def get_queries() -> dict[str, str]:
+    """Return the mapping used to generate the export queries for each table."""
```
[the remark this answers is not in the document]
mgeplf: I think it's probably good enough that you're already printing the mismatch between tables and `get_queries` in `check_queries`

scripts/export/README.md
```diff
@@ -0,0 +1,95 @@
+# Exporting and importing public data
+
+## Database
+
+### Update the bash scripts
+
+After any db migration, the script `build_database_archive.sh` should be updated by running:
```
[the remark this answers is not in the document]
GianlucaFicarelli: The issue is that the dump script requires psql/pg_dump/makeself that are not available in the image.
Maybe it's easier to retrieve a specific version from the git history and execute it where it's needed?
For example from github:
```
curl -O https://raw.githubusercontent.com/openbraininstitute/entitycore/2025.10.8/scripts/export/build_database_archive.sh
```

scripts/export/write_scripts.py
GianlucaFicarelli: yeah I can improve it, although `/usr/pgsql-17` is specific to our bastion host, so I'm not sure it can be generic enough to take in account all the possible versions and locations.
Instead of checking the OS, I could check if the new path exists and add it only in that case.

mgeplf: >  .....add it only in that case.

Or just use it directly?

eleftherioszisis (approved):

PR #320, opened 2025-08-06
Add circuit hierarchy enpoint

GianlucaFicarelli:


app/routers/hierarchy.py
```diff
@@ -0,0 +1,10 @@
+from fastapi import APIRouter
+
+import app.service.hierarchy
+
+router = APIRouter(
+    prefix="/hierarchy",
+    tags=["hierarchy"],
+)
+
+router.get("/circuit")(app.service.hierarchy.read_circuit_hierarchy)
```
GianlucaFicarelli, line 10: Note: we can decide for a different endpoint name, I don't have a string opinion.
- /hierarchy/circuit: it can be useful if we think we can reuse the same logic for other entities
- /circuit-hierarchy: it's specific for circuits
- ...

mgeplf: I don't have a strong preference, I usually like things organized like `circuit/hierarchy`, but that's just how my brain works.

mgeplf (commented):
LGTM, but I'm not the end user, so I'll let Bilal have a look.

Very clean.

app/routers/hierarchy.py
GianlucaFicarelli: `/circuit/hierarchy` could work for me as well, however it should be added before `/circuit/{uuid}`, because FastAPI evaluates routes in the order they are added.

GianlucaFicarelli: If we change the route from `/hierarchy/circuit` to `/circuit/hierarchy` I can move the endpoint definition in the to `routers/circuit.py`, but keep `service/hierarchy.py`, as the logic there isn't very specific to the circuit.

It can remain generic as long as we don't add attributes that are specific to the circuit, or we'll need different schemas for the response.

app/schemas/hierarchy.py
```diff
@@ -0,0 +1,21 @@
...
+    name: str
+    parent_id: uuid.UUID | None
+    children: list["HierarchyNode"] = []
+    authorized_public: bool
+    authorized_project_id: uuid.UUID
+
+
+class HierarchyTree(BaseModel):
+    derivation_type: DerivationType
+    data: list[HierarchyNode]
```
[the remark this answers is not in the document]
mgeplf: Unless we remove things now, they will probably be required forever :)
OTOH, we haven't really been too careful about optimizing payloads thus far, so maybe we don't need to start now.

app/service/hierarchy.py
```diff
@@ -0,0 +1,120 @@
...
+    Depending on the derivation type, the hierarchy will be built differently. In particular,
+    a circuit is considered a root if it has no parents of the specified derivation type.
+
+    The hierarchy assumes the following rules for the derivations:
+
+    - A circuit can have zero or more children linked with any derivation type.
+    - A circuit can have zero or more parents, provided each parent is different, and is linked with
+      a different derivation type.
+    - A public circuit can have any combination of public and private circuits as children.
+    - A private circuit can have only private circuits with the same project_id as children.
```
mgeplf, line 102: maybe link the discussion where these rules were discussed?

tests/test_hierarchy.py
```diff
@@ -0,0 +1,485 @@
...
+        )
+        for i, override_dict in enumerate(overrides)
+    ]
+    return add_all_db(db, rows)
+
+
+@pytest.fixture
+def hierarchy(db, root_circuits, models):
+    """Build a circuit hierarchy.
+
```
mgeplf, line 132: maybe mention this is mermaidjs, in case someone doesn't recognize it :)

app/service/hierarchy.py
```diff
@@ -0,0 +1,120 @@
...
+    db: Session,
+    project_id: uuid.UUID | None,
+    entity_class: type[Entity],
+    derivation_type: DerivationType,
+) -> dict[uuid.UUID, HierarchyNode]:
+    root = aliased(entity_class, flat=True, name="root")
+    parent = aliased(entity_class, flat=True, name="parent")
+    child = aliased(entity_class, flat=True, name="child")
+    order_by = ["name", "id"]
+    subq = sa.select(sa.literal(1)).where(
```
mgeplf, line 27: perhaps name the subq with its intended purpose?

mgeplf (commented):
LGTM, just some small nitpicks.  I'll let @bilalesi do the final approval.

app/service/hierarchy.py
GianlucaFicarelli: I could call it `matching_derivation_for_root` so the condition will become `where(~sa.exists(matching_derivation_for_root))`

PR #326, opened 2025-08-08
Add timezone to scientific_artifact.experiment_date

GianlucaFicarelli:


mgeplf (approved):

GianlucaFicarelli:
With the migration, any existing experiment_date will be converted to timestamp with tz as: 2025-08-01 00:00:00 -> 2025-08-01 00:00:00+00

PR #328, opened 2025-08-12
Fix legacy attributes

GianlucaFicarelli:
- Replace ScientificArtifactFilterBase.contact_id (uuid) with contact_email (str)
- Remove ScientificArtifactBase.atlas_id because it doesn't exist in the db model

### DIFF
--- a/app/filters/scientific_artifact.py
+++ b/app/filters/scientific_artifact.py
@@ -1,4 +1,3 @@
-import uuid
 from datetime import datetime
 from typing import Annotated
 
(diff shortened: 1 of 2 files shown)


GianlucaFicarelli:
For confirmation or discussion: is it correct that the db model of `ScientificArtifact` doesn't contain `atlas_id`?
Or at the contrary is it needed for all the classes inheriting from scientific artifact?
Note that `atlas_id` has been explicitly added to the circuit class, but not to scientific_artifact.

@darshanmandge @AurelienJaquier

mgeplf (approved):
Code is fine; consider this a LGTM for that side - whether atlas_id needs to exist is something I don't know.

PR #333, opened 2025-08-14
Fixes for morphology_model branch

eleftherioszisis:


app/routers/legacy/model/utils.py
```diff
@@ -35,7 +35,7 @@
         # app.models.annotation.MTypeAnnotationBody: "Class",
         # app.models.annotation.ETypeAnnotationBody: "Class",
         Species: "Species",
-        ReconstructionMorphology: "https://neuroshapes.org/ReconstructedNeuronMorphology",
+        CellMorphology: "https://neuroshapes.org/ReconstructedNeuronMorphology",
```
GianlucaFicarelli, line 38: Unrelated to this PR and it can be done later, but we may want to remove the legacy routers that aren't exposed anyway, so we don't have to update them.

app/schemas/morphology.py
```diff
@@ -18,30 +24,96 @@
...
+
+class ProtocolMixin(BaseModel):
+    """Generic Experimental method.
+
+    Parameters:
+    -----------
+    protocol_document: str, optional
+        URL link to protocol document or publication
+    protocol_design: str
+        From a controlled vocabulary (e.g., EM, CellPatch, Fluorophore, Imp)
```
GianlucaFicarelli, line 38: According to the docstring, protocol_design should be defined as a controlled vocabulary?

app/schemas/morphology.py
```diff
@@ -18,30 +24,96 @@
...
+    slicing_thickness: float
+        Thickness of the slice in microns
+    slicing_direction: SlicingDirectionType, optional
+        Direction of slicing
+    magnification: float, optional
+        Magnification level used
+    tissue_shrinkage: float, optional
+        Amount tissue shrunk by (not correction factor)
+    has_been_corrected_for_shrinkage: bool, optional
+        Whether data has been corrected for shrinkage
```
GianlucaFicarelli, line 62: The docstring is `has_been_corrected_for_shrinkage` but the attribute is `corrected_for_shrinkage`

app/schemas/morphology.py
```diff
@@ -18,30 +24,96 @@
...
+    CellMorphologyBase,
     LicensedCreateMixin,
     AuthorizationOptionalPublicMixin,
 ):
-    species_id: uuid.UUID
-    strain_id: uuid.UUID | None = None
+    subject_id: uuid.UUID
     brain_region_id: uuid.UUID
     legacy_id: list[str] | None = None
+    morphology_protocol_id: uuid.UUID | None = None  # This would be set after the method is created
```
GianlucaFicarelli, line 112: I wonder why morphology_protocol_id would be set later?
Cannot it be made not nullable?

app/service/morphology.py
```diff
@@ -43,24 +44,25 @@
...
+        joinedload(CellMorphology.morphology_protocol),
+        selectinload(CellMorphology.contributions).selectinload(Contribution.agent),
+        selectinload(CellMorphology.contributions).selectinload(Contribution.role),
+        joinedload(CellMorphology.mtypes),
+        joinedload(CellMorphology.license),
+        joinedload(CellMorphology.subject).joinedload(Subject.species),
+        joinedload(CellMorphology.subject).joinedload(Subject.strain),
+        selectinload(CellMorphology.assets),
+        joinedload(CellMorphology.created_by),
+        joinedload(CellMorphology.updated_by),
```
GianlucaFicarelli, line 57: I think that some of the joins can be inner joins:
- brain_region
- morphology_protocol only if't made not nullable
- created_by
- updated_by

app/db/model.py
```diff
@@ -685,19 +683,86 @@ def measurement_annotation(cls):
...
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scientific_artifact.id"), primary_key=True)
 
-    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
     location: Mapped[PointLocation | None]
 
+    # New foreign key for the morphology method
+    morphology_protocol_id: Mapped[uuid.UUID | None] = mapped_column(
+        ForeignKey("morphology_protocol.id"), index=True
+    )
```
GianlucaFicarelli, line 754: Can the same protocol be reused for multiple morphologies?
Or is it a one-to-one relationship?

app/db/model.py
```diff
@@ -685,19 +683,86 @@ def measurement_annotation(cls):
...
+    # Relationship to the polymorphic MorphologyProtocol
+    morphology_protocol: Mapped["MorphologyProtocol"] = relationship(
+        "MorphologyProtocol",
+        foreign_keys=[morphology_protocol_id],
+        uselist=False,
+    )
+
+    # structured jsonb , pydantic typedict
+    # Attributes from DigitalReconstruction
```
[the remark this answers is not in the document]
GianlucaFicarelli: I removed them.

app/schemas/morphology.py
GianlucaFicarelli: I just fixed the docstring with `corrected_for_shrinkage` because this name is used also in the db model

app/db/model.py
```diff
@@ -19,7 +18,7 @@
     UniqueConstraint,
     func,
 )
-from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR
+from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR, UUID
```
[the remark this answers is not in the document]
GianlucaFicarelli: Fixed.

app/db/model.py
```diff
@@ -685,19 +683,86 @@ def measurement_annotation(cls):
...
+
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": __tablename__,
+        "polymorphic_on": "type",
+    }
+
+
+class ExperimentalMorphologyProtocol(MorphologyProtocol):
+    __tablename__ = (
+        MorphologyGenerationType.digital.value
```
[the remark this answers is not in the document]
GianlucaFicarelli: I pushed a commit with the STI approach, but reverting the model to JTI is straightforward.
However, it could affect the joins in the _load method when we need to retrieve the morphology with the protocol.

app/db/model.py
dkeller9: Yes, they could be reused. Any user should be allowed to create them. If the morphology protocol is exactly the same, then it should be reused and a new one not created.

app/routers/legacy/model/utils.py
dkeller9: Agreed.

app/schemas/morphology.py
```diff
@@ -18,30 +24,77 @@
...
+        corrected_for_shrinkage: Whether data has been corrected for shrinkage.
+    """
+
+    id: uuid.UUID
+    staining_type: StainingType | None = None
+    slicing_thickness: float
+    slicing_direction: SlicingDirectionType | None = None
+    magnification: float | None = None
+    tissue_shrinkage: float | None = None
+    corrected_for_shrinkage: bool | None = None
```
GianlucaFicarelli, line 60: Why `corrected_for_shrinkage` can be None also if it's a bool? Does it mean that the value is unknown in that case?
@dkeller9

app/schemas/morphology.py
```diff
@@ -18,30 +24,77 @@
...
+        tissue_shrinkage: Amount tissue shrunk by (not correction factor).
+        corrected_for_shrinkage: Whether data has been corrected for shrinkage.
+    """
+
+    id: uuid.UUID
+    staining_type: StainingType | None = None
+    slicing_thickness: float
+    slicing_direction: SlicingDirectionType | None = None
+    magnification: float | None = None
+    tissue_shrinkage: float | None = None
```
GianlucaFicarelli, line 59: Is there any validation to be added for the floats, e.g. min and max?
- slicing_thickness
- magnification
- tissue_shrinkage

@dkeller9

app/schemas/morphology.py
```diff
@@ -18,30 +24,77 @@
...
+
+class ProtocolMixin(BaseModel):
+    """Generic Experimental method.
+
+    Attributes:
+        protocol_document: URL link to protocol document or publication.
+        protocol_design: From a controlled vocabulary (e.g., EM, CellPatch, Fluorophore, Imp)
+    """
+
+    protocol_document: str | None = None
```
GianlucaFicarelli, line 38: Can we validate that protocol_document is a http/https url?
Or should we leave it as a free text?
@dkeller9

app/db/model.py
dkeller9: Even minor differences in protocols should be considered separate. Also, I think it's convenient to have separate endpoints.

app/db/model.py
```diff
@@ -699,18 +696,74 @@ def measurement_annotation(cls):
...
+
+    __mapper_args__ = {  # noqa: RUF012
+        "polymorphic_identity": MorphologyGenerationType.digital.value,
+    }
+
+
+class ComputationallySynthesizedMorphologyProtocol(MorphologyProtocol):
+    # __tablename__ = f"morphology_protocol_{MorphologyGenerationType.computational}"
+    # id: Mapped[uuid.UUID] = mapped_column(ForeignKey("morphology_protocol.id"), primary_key=True)
+    method_type: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
```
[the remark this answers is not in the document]
dkeller9: That is fine

app/db/types.py
```diff
@@ -404,12 +444,17 @@ class LabelRequirements(BaseModel):
...
         AssetLabel.morphology: [
             LabelRequirements(content_type=ContentType.asc, is_directory=False),
             LabelRequirements(content_type=ContentType.swc, is_directory=False),
             LabelRequirements(content_type=ContentType.h5, is_directory=False),
         ],
+        AssetLabel.morphology_with_spines: [
+            LabelRequirements(content_type=ContentType.asc, is_directory=False),
+            LabelRequirements(content_type=ContentType.swc, is_directory=False),
+            LabelRequirements(content_type=ContentType.h5, is_directory=False),
+        ],
```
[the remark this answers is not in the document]
dkeller9: The same source morphology could have both the assets morphology and morphology_with_spines, so two labels seems necessary.

app/schemas/morphology.py
dkeller9: Yes, a url validation is a good idea.

app/schemas/morphology.py
dkeller9: For the moment we can use [EM, CellPatch, Fluorophore]

dkeller9: Yes, lower case and electron_microscopy are both fine.

app/db/model.py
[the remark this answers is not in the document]
dkeller9: STI sounds good.

app/db/model.py
dkeller9: The same approach as for subject can be used. For now we won't enforce uniqueness.

app/db/types.py
```diff
@@ -8,6 +8,45 @@
...
+    mix_and_match = auto()
+    mousified = auto()
+    ratified = auto()
+
+
+class MorphologyGenerationType(StrEnum):
+    digital = auto()  # digital reconstruction
+    modified = auto()  # modifed from digital
+    computational = auto()  # computationally synthesized
+    placeholder = auto()  # could be a morphology or compartment
```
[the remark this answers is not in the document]
dkeller9: The types could be:

digital_reconstruction
modified_reconstruction
computationally_synthesized
placeholder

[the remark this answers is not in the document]
dkeller9: @dkeller9 to look into which protocol to assign for migration.

alembic/versions/20250826_153359_7f8e31734e4c_replace_reconstruction_morphology.py
```diff
@@ -0,0 +1,873 @@
+"""Replace reconstruction_morphology with cell_morphology.
```
GianlucaFicarelli, line 1: Note that this migration required mostly manual operations, so it needs to be reviewed

alembic/versions/20250905_134708_01644f002227_add_triggers.py
```diff
@@ -0,0 +1,75 @@
+"""Add triggers
```
GianlucaFicarelli, line 1: This migration instead has been generated automatically

app/db/model.py
```diff
@@ -699,18 +697,76 @@ def measurement_annotation(cls):
         )
 
 
-class ReconstructionMorphology(
+class CellMorphologyProtocol(Entity):
```
GianlucaFicarelli, line 700: The base protocol class is now called CellMorphologyProtocol instead of MorphologyProtocol to be more specific, and similar to CellMorphology.
The same for the endpoint that is called /cell-morphology-protocol

app/db/triggers.py
```diff
@@ -98,15 +116,18 @@ def unauthorized_private_reference_trigger(
...
-    unauthorized_private_reference_trigger(MEModel, "morphology_id", ReconstructionMorphology),
-    unauthorized_private_reference_function(MEModel, "emodel_id", EModel),
-    unauthorized_private_reference_trigger(MEModel, "emodel_id", EModel),
-]
+for model, field_name, target_model in [
+    (EModel, "exemplar_morphology_id", CellMorphology),
+    (MEModel, "morphology_id", CellMorphology),
+    (MEModel, "emodel_id", EModel),
+    (CellMorphology, "cell_morphology_protocol_id", CellMorphologyProtocol),
+    # TODO: enable the following triggers for subject_id
```
GianlucaFicarelli, line 124: I'll open a separate PR to enable triggers for subject_id.
We can also discuss alternatives approaches, but for the moment this is the simplest check.

app/db/utils.py
```diff
@@ -22,6 +31,15 @@
...
 
+CELL_MORPHOLOGY_GENERATION_TYPE_TO_CLASS: dict[
+    CellMorphologyGenerationType, type[CellMorphologyProtocol]
+] = {
+    CellMorphologyGenerationType[mapper.polymorphic_identity]: mapper.class_
+    for mapper in Base.registry.mappers
+    if issubclass(mapper.class_, CellMorphologyProtocol)
+    and mapper.class_ != CellMorphologyProtocol  # exclude the base class
+    and mapper.polymorphic_identity
+}
```
GianlucaFicarelli, line 42: This mapping is used to know which table is needed when inserting a new record.

GianlucaFicarelli:
The latest changes are ready for review.

app/db/types.py
```diff
@@ -10,6 +10,51 @@
...
+    custom = auto()
+
+
+class StainingType(StrEnum):
+    golgi = auto()
+    nissl = auto()
+    luxol_fast_blue = auto()
+    fluorescent_nissl = auto()
+    fluorescent_dyes = auto()
+    fluorescent_orotein_expression = auto()
```
mgeplf, line 53: `fluorescent_protein_expression`?

app/schemas/cell_morphology_protocol.py
```diff
@@ -0,0 +1,215 @@
...
+class CommonReadMixin(
+    IdentifiableMixin,
+    CreationMixin,
+    CreatedByUpdatedByMixin,
+    AuthorizationMixin,
+):
+    """Common mixin for readable schemas."""
+
+
+class ProtocolMixin:
```
mgeplf, line 32: do you think there will be other `Protocols`, I can't tell if this should be named more specifically or not.

app/db/types.py
GianlucaFicarelli: Good eye, going to fix it

app/schemas/cell_morphology_protocol.py
GianlucaFicarelli: This is an internal class in cell_morphology_protocol.py, but I'll rename it to CellMorphologyProtocolMixin

mgeplf (approved):

GianlucaFicarelli:
The last 2 commits after "Fix after merge" should be reviewed. Changes:

- Simplify CellMorphologyRead CellMorphologyCreate by inheriting from ScientificArtifactRead and ScientificArtifactCreate (and fix an issue with missing attributes, test added)
- Minor, use pytest.mark.parametrize in test_create_one
- Simplify TypeAdapter by defining a SupportsModelValidate protocol (uses an approach similar to the one in https://github.com/openbraininstitute/entitysdk/pull/126)

Issue #330, opened 2025-08-12
Add entities for EM dense reconstructions and EM cell mesh


MWolfR:
To get Marwan's Ultraliser pipeline working on the platform, we need a representation of EM-derived cell meshes. During discussions, we found that this also requires a representation of the EM release (such as MICrONS, H01, etc.) itself.

I tried to draft the changes in the branch "em_reconstruction", but frankly - I have no idea what I am doing. This needs an engineer to look at and probably fix.

As I am unfamiliar with this, I can only guess the estimated time at 2-3 days.

Acceptance criteria is that we can create entities with the new types and register them in the database.

evakennyobi:
To discuss in the first instance with @GianlucaFicarelli

PR #339, opened 2025-08-21
Add EMCellMesh and EMDenseReconstructionDataset

GianlucaFicarelli:
This is based on the branch `em_reconstruction` by @MWolfR 

Fixes https://github.com/openbraininstitute/entitycore/issues/330

Notes:
- `EMDenseReconstructionDataset` inherits from ScientificArtifact because it needs all the base attributes. However, only service admins are allowed to create a new resource, so a new fastapi dependency called AdminContextWithProjectIdDep has been created for this use case. Alternatively, we could define `EMDenseReconstructionDataset` as a global resource, but then we'll need to add all the attributes and relationships of ScientificArtifact excluding authorized_public and authorized_project_id, and this seems less clean.
- `POST /em-dense-reconstruction-dataset` can be called only by admins with a project-id to create a new EMDenseReconstructionDataset
- `POST /em-cell-mesh` can be called by anyone to create a new EMCellMesh, and needs to reference an existing em-dense-reconstruction-dataset
- `GET` endpoints can be called by every authenticated user as usual.

### DIFF
--- a/docs/asset-labels.md
+++ b/docs/asset-labels.md
@@ -4,8 +4,8 @@
 | brain_atlas_region                    | brain_atlas_region_mesh                           | application/obj    | .obj            | Brain atlas region mesh geometry object.                                                                    |
 | cell_composition                      | cell_composition_summary                          | application/json   | .json           | Region/mtype/etype densities summary                                                                        |
 | cell_composition                      | cell_composition_volumes                          | application/json   | .json           | mtype/etype voxel densities composition                                                                     |
-| em_cell_mesh                          | cell_surface_mesh                                 | application/x-hdf5 | .h5             |
-a triangle mesh describing the surface of a cell
+| em_cell_mesh                          | cell_surface_mesh                                 | application/x-hdf5 | .h5             | a triangle mesh describing the surface of a cell in h5 format
+|                                       |                                                   | application/obj    | .obj            | a triangle mesh describing the surface of a cell in obj format
 | circuit                               | sonata_circuit                                    | N/A                | N/A (directory) | SONATA circuit, but have a circuit_config.json in the root of the directory                                 |
 | circuit                               | compressed_sonata_circuit                         | application/gzip   | .gz             | Compressed SONATA circuit in GZIP format                                                                    |
 | circuit                               | circuit_connectivity_matrices                     | N/A                | N/A (directory) | Connectivity matrices in ConnectomeUtilities format, with a matrix_config.json in the root of the directory |


app/db/model.py
```diff
@@ -1436,3 +1438,116 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
...
+    Notes:
+        - Has no assets. Data access all through the specified URLs.
+    """
+
+    __tablename__ = EntityType.em_dense_reconstruction_dataset.value
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scientific_artifact.id"), primary_key=True)
+
+    protocol: Mapped[str | None]
+    fixation: Mapped[str | None]
+    staining: Mapped[str | None]  # TODO: controlled vocabulary?
```
GianlucaFicarelli, line 1490: Can a StrEnum be defined?

app/db/model.py
```diff
@@ -1436,3 +1438,116 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
...
+    slicing: Mapped[float | None]
+    shrinkage: Mapped[float | None]
+    microscope_type: Mapped[str | None]  # TODO: controlled vocabulary
```
GianlucaFicarelli, line 1493: Can a StrEnum be defined?

app/service/em_cell_mesh.py
```diff
@@ -0,0 +1,123 @@
...
+        authorized_project_id=user_context.project_id,
+        response_schema_class=EMCellMeshRead,
+        apply_operations=_load,
+    )
+
+
+def create_one(
+    user_context: UserContextWithProjectIdDep,
+    db: SessionDep,
+    ion_channel_model: EMCellMeshCreate,
```
MWolfR, line 114: ion channel?

app/service/em_cell_mesh.py
```diff
@@ -0,0 +1,123 @@
...
+) -> EMCellMeshRead:
+    return router_create_one(
+        db=db,
+        apply_operations=_load,
+        user_context=user_context,
+        json_model=ion_channel_model,
```
MWolfR, line 120: Here as well, why ion channel in cell mesh?

MWolfR (commented):
Most of the parts I am not qualified to review. But I am OK with the changes made to contributions that were originally mine. 

I am only puzzled by two references to "ion_channel_model" (see specific comments).

app/db/model.py
MWolfR: Let's keep it a string.

app/db/model.py
```diff
@@ -1436,3 +1439,116 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
...
+
+
+class EMCellMesh(ScientificArtifact):
+    """Cell surface mesh created from a dense EM reconstruction.
+
+    Attributes:
+        id (uuid.UUID): Primary key.
+        em_dense_reconstruction_dataset_id (uuid.UUID): The id of the dense em reconstruction
+            dataset that the mesh originates from.
+        release_version (int): Version of the em reconstruction dataset that was used.
```
jdcourcol, line 1522: why is the release_version not on the em reconstruction dataset ?

Does that mean the an "em reconstruction dataset" represents multiple version of the dataset ?

app/db/model.py
```diff
@@ -1436,3 +1439,116 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
...
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scientific_artifact.id"), primary_key=True)
+    em_dense_reconstruction_dataset_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("em_dense_reconstruction_dataset.id"), index=True
+    )
+    release_version: Mapped[int]
+    dense_reconstruction_cell_id: Mapped[int]
+    generation_method: Mapped[EMCellMeshGenerationMethod]
+    level_of_detail: Mapped[int]
+    generation_parameters: Mapped[JSON_DICT | None]
```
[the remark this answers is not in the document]
GianlucaFicarelli: All these attributes are required in both the Create and Read pydantic schemas (only generation_parameters is optional), so I would expect that the users know what they mean. Isn't that the case @MWolfR?

app/db/model.py
jdcourcol: is the user going to provide this version ? @MWolfR ?

PR #355, opened 2025-09-09
Add admin update endpoints

eleftherioszisis:


app/queries/common.py
```diff
@@ -354,15 +355,17 @@ def router_update_one[T: BaseModel, I: Identifiable](
     id_: uuid.UUID,
     db: Session,
     db_model_class: type[I],
-    user_context: UserContext,
+    user_context: UserContext | None,
```
mgeplf, line 358: I wonder if it would make sense to have an `EmptyUserContext` type, instead of `None`, when it comes to authentication related things, I think it's worthwhile to be more explicit that "yes, we don't need a UserContext here", since I find None can be a bit too easy to "skip over" when reading the code.

Dunno, just a thought.  Maybe @GianlucaFicarelli has input.

app/routers/ion_channel_recording.py
```diff
@@ -1,6 +1,7 @@
 from fastapi import APIRouter
 
 import app.service.ion_channel_recording
+from app.routers.admin import router as admin_router
 
 router = APIRouter(
     prefix="/ion-channel-recording",
```
mgeplf, line 7: no `ROUTE = "ion-channel-recording"` like the others?

app/schemas/activity.py
```diff
@@ -34,6 +36,7 @@ class ActivityCreate(ActivityBase, AuthorizationOptionalPublicMixin):
     generated_ids: list[uuid.UUID] = []
 
 
-class ActivityUpdate(ActivityBase, AuthorizationOptionalPublicMixin):
-    end_time: datetime | None = None
-    generated_ids: list[uuid.UUID] | None = None
+class ActivityUpdate(BaseModel):
+    start_time: datetime | Literal["<NOT_SET>"] | None = NOT_SET
```
mgeplf, line 40:
```suggestion
    start_time: datetime | Literal[NOT_SET] | None = NOT_SET
```

Might be worth making a type for it if it'll be used elsewhere.

app/service/electrical_cell_recording.py
```diff
@@ -27,9 +27,10 @@
 )
 from app.queries.factory import query_params_factory
 from app.schemas.electrical_cell_recording import (
+    ElectricalCellRecordingAdminUpdate,
     ElectricalCellRecordingCreate,
     ElectricalCellRecordingRead,
-    ElectricalCellRecordingUpdate,
+    ElectricalCellRecordingUserUpdate,
```
mgeplf, line 33: does this one really need a special `ElectricalCellRecordingUserUpdate`, some of the others don't have `User` in it, and I think that makes sense.

tests/test_electrical_recording_stimulus.py
```diff
@@ -98,8 +98,44 @@ def test_update_one(client, model_id):
     assert data["start_time"] is None
     assert data["end_time"] is None
 
+    # only admin client can hit admin endpoint
```
mgeplf, line 101: it looks like most of the tests are nearly the same, there's no way to reduce the duplication, is there?

mgeplf (commented):
Quite the slog; seems reasonable to me, couple small comments.

app/schemas/activity.py
mgeplf: weird, I checked it w/ python 3.11:
```
>>> NOT_SET = "<NOT_SET>"
>>> from typing import Literal
>>> NST = Literal[NOT_SET]
>>> NST
typing.Literal['<NOT_SET>']

```
which is why I thought it would work :)

Seems to work under 3.12 as well:
```
>>> NOT_SET = "<NOT_SET>"
>>> from typing import Any, Literal
>>> NST = Literal[NOT_SET]
>>> NST
typing.Literal['<NOT_SET>']
>>> import sys; print(sys.version_info)
sys.version_info(major=3, minor=12, micro=8, releaselevel='final', serial=0)
```

Strange.

app/routers/ion_channel_recording.py
eleftherioszisis: Fixed

PR #356, opened 2025-09-10
Circuit node population properties

james-isbister:
@MWolfR asked me to make a PR for adding simple node population properties to the Circuit entity.

These properties will be required by the UI and AI agent for designing simulation campaigns. 

Whilst some of the properties are very cheap to compute (the list of populations from the circuit config), others require downloading the nodes file which doesn't scale to larger circuits if we don't want the UI and AI agent to be slow and expensive

We would also like to standardize how such properties are fetched, hence adding them as properties of the Circuit entity. This was also deemed optimal for the AI agent @jankrepl @WonderPG @BoBer78 and also would be simpler for the UI simulation configuration I think

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1374,6 +1374,27 @@ class ScientificArtifactExternalUrlLink(Identifiable):
     )
 
 
+from enum import auto, StrEnum
+from pydantic import BaseModel
+class NodePopulationType(StrEnum):
+    biophysical = auto()
+    virtual = auto()
+
+class SpatialCoordinate(StrEnum):
+    x = auto()
+    y = auto()
+    z = auto()
+
+class CircuitMetricsNodePopulation(BaseModel):
+    number_of_nodes: int
+    name: str
+    population_type: NodePopulationType
+    property_names: list[str]
+    property_unique_values: dict[str, list[str]]
+    property_value_counts: dict[str, dict[str, int]]
+    node_location_info: dict[SpatialCoordinate, dict[str, float]] | None
+
+
 class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
     """Represents a neural circuit as a scientific artifact.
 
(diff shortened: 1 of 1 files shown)


app/db/model.py
```diff
@@ -1374,6 +1376,27 @@ class ScientificArtifactExternalUrlLink(Identifiable):
...
+
+
+class CircuitMetricsNodePopulation(BaseModel):
+    number_of_nodes: int
+    name: str
+    population_type: NodePopulationType
+    property_names: list[str]
+    property_unique_values: dict[str, list[str]]
+    property_value_counts: dict[str, dict[str, int]]
+    node_location_info: dict[SpatialCoordinate, dict[str, float]] | None
```
eleftherioszisis, line 1397: If someone selects 100k nodes we will have dictionaries with 100k keys as metadata?

james-isbister: Yes, sorry, I meant to come back to this and check whether node_location_info was a dictionary by node. I was copying some of Michael's code. I can remove that

app/db/model.py
```diff
@@ -1428,6 +1451,15 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
...
     number_connections: Mapped[int | None] = mapped_column(BigInteger)
 
+    names_of_biophys_node_populations: list[str]
+    names_of_virtual_node_populations: list[str]
+    names_of_node_populations: list[str]
+    # Probably the list of names is enough
+    # number_of_biophys_node_populations: int
+    # number_of_virtual_node_populations: int
+    biophysical_node_populations: list[CircuitMetricsNodePopulation | None]
+    virtual_node_populations: list[CircuitMetricsNodePopulation | None]
```
eleftherioszisis, line 1460: Wouldn't it be cleaner if we had one entry node_populations that lists all populations with extra info, e.g. their type?

There are circuits that have neither biophysical or virtual neurons where these lists will just be empty.

Issue #357, opened 2025-09-10
Questions about SimulationCampaign


mgeplf:
1) What is the meaning of `entity_id` on `SimulationCampaign`?

It is defined here:
https://github.com/openbraininstitute/entitycore/blob/4b58023f5f3b9a02ca5540817a7ccddacf9c975a/app/db/model.py#L1138

However, each `Simulation` also has an `entity_id` (https://github.com/openbraininstitute/entitycore/blob/4b58023f5f3b9a02ca5540817a7ccddacf9c975a/app/db/model.py#L1138); is it required that these are the same - currently the tests check for a `SimulationCampaign` with different `Simulation`s with different circuits: https://github.com/openbraininstitute/entitycore/blob/4b58023f5f3b9a02ca5540817a7ccddacf9c975a/tests/test_simulation_campaign.py#L244

Perhaps `entity_id` should be removed from `SimulationCampaign`?

2) Both `SimulationCampaign` and `Simulation` have `scan_parameters`; is the expectation that the `SimulationCampaign` ones describe the scan, and the `Simulation` one describes precisely the parameters run by the simulation?  I think it would be helpful to describe the contents of this json more clearly.

mgeplf:
> `entity_id` is usually pointing to the circuit used for the SimulationCampaign.

That's what I had assumed, however, my point is that it *may not* match the `Simulation`s; and I don't know whether that's intended or not.

Basically, I think that either:
1) Only `SimulationCampaign` should have `entity_id` if all `Simulation`s are required to use the same Circuit (or other entity)
2) `SimulationCampaign` should not have `entity_id` if the `Simulation`s can point to different entities.

Otherwise there is a mismatch - the `SimulationCampaign` entity might point at 1 circuit, but the `Simulation`s can point at another, which seems inconsistent.

g-bar:
We actually have no choice, we have to keep it in the SimulationCampaign, if anything it should be removed from the Simulation.

The reason is that in the UI we have these tables `Small microcircuit simulation` and `Paired neuron simulation`, which in reality are not simulations but simulation campaigns and in order to determine what kind they are we filter them by `circuit__scale=small` and `circuit__scaile=paired` respectively.

[...]

PR #360, opened 2025-09-11
Fixes for SimulationCampaign

eleftherioszisis:
* Add entity_id filtering
* Add circuit nested filter
* Prefix simulation circuit correctly.

### DIFF
--- a/app/filters/simulation.py
+++ b/app/filters/simulation.py
@@ -16,7 +16,6 @@ class SimulationFilterBase(NameFilterMixin, IdFilterMixin, CustomFilter):
 
 
 class NestedSimulationFilter(SimulationFilterBase):
-
     circuit: Annotated[
         NestedCircuitFilter | None,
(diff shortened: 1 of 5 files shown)


eleftherioszisis:
@g-bar needs the querying using the entity_id and the circuit filter to get all simulations per campaign with the correct pagination.

mgeplf:
> needs the querying using the entity_id and the circuit filter to get all simulations per campaign with the correct pagination.

I think we should decide on https://github.com/openbraininstitute/entitycore/issues/357#issuecomment-3274147176; if we make the `entity_id` on `SimulationCampaign` load bearing, we can't change it - and then we either need to remove it from `Simulation` or we can have both, and then mismatches, which seems like a problem.

mgeplf:
> I think we should decide on #357 [...]

In that ticket, @jdcourcol made it clear that `entity_id` needs to be on both objects, and that it can remain `mono-valued` for now.

PR #369, opened 2025-09-17
Validate types in SimulationExecutionFilter and ElectricalRecordingStimulusFilter

GianlucaFicarelli:


eleftherioszisis (approved):

app/filters/electrical_recording_stimulus.py
```diff
@@ -13,8 +14,8 @@
 class ElectricalRecordingStimulusFilter(EntityFilterMixin, NameFilterMixin, CustomFilter):
     order_by: list[str] = ["-creation_date"]  # noqa: RUF012
 
-    shape: str | None = None
-    injection_type: str | None = None
+    shape: ElectricalRecordingStimulusShape | None = None
```
mgeplf, line 17: hmm, I wonder if we have done this anywhere else by accident? (ie: not for ElectricalRecordingStimulusShape) - maybe we can get AI to audit it /s

GianlucaFicarelli: I checked manually the filters and I found only these 2 (only SimulationExecutionFilter was reported)

PR #370, opened 2025-09-19
Make project context header optional in all but create endpoints

eleftherioszisis:
* Create endpoints require a mandatory vlab/proj id pair as before
* In all other endpoints:
   - If a user passes vlab/proj id pair in the header things will work as they used to constraining the queries into that project.
   - If a user passes just their token, then the user_project_ids in the token will be used in the queries.
   - vlab/proj id context has precedence over user_project_ids

**Note**: Using the token is more permissive because the query will check the authorized_project_id of the entities against all authorized user project ids.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import uuid
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from datetime import timedelta
+from typing import NamedTuple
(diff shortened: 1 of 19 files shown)


jdcourcol:
>Create endpoints require a mandatory vlab/proj id pair as before

@eleftherioszisis  why not just the project_id ?

PR #371, opened 2025-09-23
Notebook entities v2

GianlucaFicarelli:
Quick overview, this PR defines:
- AnalysisNotebookTemplate(Entity, NameDescriptionVectorMixin)
- AnalysisNotebookEnvironment(Entity)
- AnalysisNotebookExecution(Activity)
- AnalysisNotebookResult(Entity, NameDescriptionVectorMixin)

Execution represents the activity of executing a notebook.
It can be optionally linked to a Template, containing the notebook template and requirements.txt as assets, and input specifications as JSONB.
It must be linked to an Environment, containing requirements.txt as an asset and the runtime information as JSONB.
It generates an AnalysisNotebookResult, containing the final notebook as an asset.

Other changes:
- unify some common logic in tests for activities

Fix https://github.com/openbraininstitute/prod-launch-notebook/issues/38


app/schemas/analysis_notebook_template.py
```diff
@@ -0,0 +1,76 @@
...
+class PythonDependency(BaseModel):
+    version: str  # e.g. ">=3.10,<3.12"
+
+
+class InputType(BaseModel):
+    name: str
+    type: EntityType
+    is_list: bool = False
+    count_min: Annotated[int, Field(ge=0)] = 1
+    count_max: Annotated[int | None, Field(ge=0)] = 1
```
GianlucaFicarelli, line 27: The format of the input can be discussed or changed to satisfy the known use cases, cc @MWolfR @james-isbister .
The idea here is that `name` and `type` are mandatory and define the variable names and the types that should be presented to the user.
The other fields can be used when a variable is a list.
We could also consider name as a template (for example `"var_{i}"` and present multiple variables with different names if that use case is needed.
See also the examples in conftest.py.

app/db/model.py
```diff
@@ -1618,3 +1619,101 @@ class EMCellMesh(ScientificArtifact):
...
+
+class AnalysisNotebookEnvironment(Entity):
+    """Represents the environment of the AnalysisNotebookExecution.
+
+    Attributes:
+        id: (uuid.UUID): Primary key, referencing the entity ID.
+        runtime_info: runtime variables associated with the environment, with schema.
+
+    Assets:
+        - requirements.txt produced with `pip freeze`.
```
mgeplf, line 1655: If the goal is reproducibility, perhaps [PEP-0751](https://peps.python.org/pep-0751/) / https://packaging.python.org/en/latest/specifications/pylock-toml/#pylock-toml-spec

is the way to go.
I can see using requirements.txt for the `AnalysisNotebookTemplate` though.

app/db/model.py
```diff
@@ -1618,3 +1619,101 @@ class EMCellMesh(ScientificArtifact):
     )
 
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+
+class AnalysisNotebookTemplate(Entity, NameDescriptionVectorMixin):
```
mgeplf, line 1624: Will there be other types of `NotebookTemplate`s other than `Analysis`?

app/db/model.py
```diff
@@ -1618,3 +1619,101 @@ class EMCellMesh(ScientificArtifact):
...
+
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+
+class AnalysisNotebookEnvironment(Entity):
+    """Represents the environment of the AnalysisNotebookExecution.
+
+    Attributes:
+        id: (uuid.UUID): Primary key, referencing the entity ID.
+        runtime_info: runtime variables associated with the environment, with schema.
```
mgeplf, line 1652: Maybe mention that the `schema` is called `RunTimeInfo`, so it's easier to find?

app/schemas/analysis_notebook_environment.py
```diff
@@ -0,0 +1,76 @@
...
+    AuthorizationMixin,
+    AuthorizationOptionalPublicMixin,
+    CreationMixin,
+    EntityTypeMixin,
+    IdentifiableMixin,
+)
+from app.schemas.utils import make_update_schema
+
+
+class PythonInfo(BaseModel):
```
mgeplf, line 15: will `PythonInfo`, `OsInfo` and `RuntimeInfo` only be used in analysis, or can they be moved into base.py or types.py?  I suppose it could be done in the future if they get used, but otherwise I"m wondering if they should be prefixed with `Analysis`?

app/schemas/analysis_notebook_template.py
```diff
@@ -0,0 +1,76 @@
...
+)
+from app.schemas.utils import make_update_schema
+
+
+class PythonDependency(BaseModel):
+    version: str  # e.g. ">=3.10,<3.12"
+
+
+class InputType(BaseModel):
+    name: str
```
mgeplf, line 23: do you want to version this too, perhaps the input schema will evolve over time - ie: I think it was a good idea in `Specifications`, and probably should be here too.

eleftherioszisis (approved):
On the technical side it look good.

app/schemas/analysis_notebook_template.py
```diff
@@ -0,0 +1,76 @@
...
+    type: EntityType
```
mgeplf, line 24: I think it might be clearer to call this `entity_type`

tests/test_simulation_generation.py
```diff
@@ -128,48 +139,15 @@ def test_create_one__unauthorized_entities(
...
-    assert_request(client_user_1.post, url=ROUTE, json=authorized_used, expected_status_code=200)
-
-    # user 1 is allowed to create Generation with public entity created by user2
-    authorized_used = json_data | {
-        "used_ids": [str(user1_morph_id)],
-        "generated_ids": [str(user2_public_morph_id)],
-    }
-    assert_request(client_user_1.post, url=ROUTE, json=authorized_used, expected_status_code=200)
-
-    # user 1 is allowed to create both with public entity created by user2
```
mgeplf, line 167: nice to clean all this up

mgeplf (commented):
seems reasonable to me

app/db/model.py
GianlucaFicarelli: Not sure about that, all the new classes have a prefix `AnalysisNotebook` to make clear that they are used for that.

@MWolfR do we have plan to introduce and support other types of notebooks?

app/schemas/analysis_notebook_environment.py
GianlucaFicarelli: We might need to prefix them if we make them global and not conflicting with other classes, but do we want to add a prefix also if they are local?.
We could also move them to base.py or types.py, but I cannot foresee if they can be used as they are in other cases.

app/schemas/analysis_notebook_template.py
GianlucaFicarelli: There is a `schema_version` at the top level `Specifications`, I think that InputType should follow that schema unless we want to expose InputType standalone without Specifications.

app/schemas/analysis_notebook_environment.py
GianlucaFicarelli: Yeah I can move them to types.py so we can try to reuse or adapt them later when needed.

GianlucaFicarelli:
> I would like to ask how we should use the environment in practice... For example there's that metabolism notebook from Polina which uses quite a lot of outdated library versions and which doesn't work on the latest version of JupyterLab: would we then, based on the older python version in the AnalysisNotebookEnvironment and certain version constraints in the requirements in the AnalysisNotebookEnvironment, deduce that a different container image should be used to start the notebook? At the moment we just got a single 'latest' version of https://github.com/openbraininstitute/obi-notebook-image but we could have some branch with an image with much older versions which could be used for the metabolism notebook.

Cannot the same image be used with a requirements.txt that specifies old fixed versions of the package? Does the notebook need different system libraries?

If we require different images, it can be a good idea to specify which one in the template (to run new analysis) and in the environment (for reproducibility/provenance).

mgeplf:
> Does the notebook need different system libraries?

The "Polina" notebook is a very special beast; it includes things like Julia, and other compiled in components, and I understand it's hard to even get everything to build in the first place - unfortunately it's not pure-python.

MWolfR:
For now, let's stick to one canonical environment that allows the other notebooks to work. 

For Polina's notebook -- I do not want to be held back by it. If there is a hack to get it to work, fine. Otherwise I would be willing to drop it. (But also check with Georges)

GianlucaFicarelli:
> For now, let's stick to one canonical environment that allows the other notebooks to work.

I'm already updating the schemas with optional nested objects for DockerDependency and DockerRuntimeInfo. They are not mandatory so they can skipped if needed, but can be used for better specifications or provenance.

app/db/model.py
MWolfR: Frozen requirements are OK.

GianlucaFicarelli: After more checks, I'm going to revert to requirements.txt in https://github.com/openbraininstitute/entitycore/pull/383 for the moment.
The support in pip and uv for pylock.toml is still experimental (in particular, pip still doesn't allow to use pylock.toml to install the packages).
Freeze cannot be enforced in this way, but I would still recommend it in both the template and in the result.

PR #378, opened 2025-09-26
Fix wrong type for SimulationGeneration

### LINKED ISSUE #377: SimulationGeneration type inconsistent with route, wrong admin delete endpoint
SimulationGeneration type is `simulation_campaign_generation` which is not consistent with neither the model nor the route names, which are SimulationGeneration and simulation-generation respectively.

This introduces a bug in the admin delete endpoint which autogenerates the route from type, expected a `simulation-campaign-generation` instead of a `simulation-generation` route.

eleftherioszisis:
* Removed generation and execution activities that are mistakenly listed under entities
* Used the ActivityType that has the correct names (which are also used in the tables) to create the routes.

Fixes #377


mgeplf:
do you think anyone is relying on this bug? I'm wondering what client code will have to change?

PR #384, opened 2025-09-30
add emodels relationship to ion channel models

AurelienJaquier:
This should solve the issue raised in this comment: https://github.com/openbraininstitute/prod-explore-functionality/issues/142#issuecomment-3345950795

### DIFF
--- a/app/schemas/emodel.py
+++ b/app/schemas/emodel.py
@@ -71,5 +71,4 @@ class EModelReadExpanded(EModelRead, AssetsMixin):
 
 
 # we have to do this to avoid cyclic imports
-# source: https://github.com/fastapi/fastapi/issues/153#issuecomment-1095709747
-IonChannelModelExpanded.update_forward_refs(EModelRead=EModelRead)
+IonChannelModelExpanded.model_rebuild()


AurelienJaquier:
@g-bar I think this should solve the problem we had in [your comment](https://github.com/openbraininstitute/prod-explore-functionality/issues/142#issuecomment-3345950795). Could you tell me if this PR would indeed solve it with a review?

g-bar:
I think this would work for me but I’ll let @GianlucaFicarelli and @eleftherioszisis review it.

app/service/ion_channel_model.py
```diff
@@ -55,6 +55,7 @@ def _load_expanded(q: Select[IonChannelModel]) -> Select[IonChannelModel]:
         selectinload(IonChannelModel.contributions).selectinload(Contribution.agent),
         selectinload(IonChannelModel.contributions).selectinload(Contribution.role),
         selectinload(IonChannelModel.assets),
+        selectinload(IonChannelModel.emodels),
```
eleftherioszisis, line 58: I understand EModel having a has-a relationship of ion-channel models, but the opposite isn't strange? This seems like a case of querying, not explicit relationship. In addition, this introduces a cycle. Does this really work? I am surprised that this would work without joining the relationships of EModel.

In any case, please add tests to ensure that the functionality that is added really works and it isn't just lazy loading or empty records that let tests pass without complaints.

eleftherioszisis (changes requested):

app/service/ion_channel_model.py
AurelienJaquier: For the [display of ion channel model in explore](https://github.com/openbraininstitute/prod-explore-functionality/issues/142) we need to get all emodels that use the ion channel.
If we want to use querying instead of relationship, what should we have to change in entitycore / entitysdk?

Issue #372, opened 2025-09-23
Make project context (vlab, proj) ids optional


eleftherioszisis:
In most endpoints a user is required to provide a project and virtual lab id in the header of the request. This project context is then validated against their user groups and used to constrain queries. The rest of their user groups is ignored in the queries.

However, that context imposes a constraint (on a specific proj id) that may not be always desirable. Therefore, I propose to make the project context optional for all endpoints, except for the create ones that require it to register the entity.

To maintain backwards compatibility and stricter access to resources, when a project context is provided the behavior should remain as before, constraining all queries to the specific project context.

If not provided then the user's groups, as obtained from keycloak, should be used to determine if a user has access to a resource. Naturally, this makes the behavior of the endpoints more permissive allowing the user to perform queries with respect of all their project ids.

**TLDR:**
* If a user passes vlab/proj id pair in the header things will work as they used to constraining the queries into that project.
* If a user passes just their token, then the user_project_ids in the token will be used in the queries.

Draft here: #370 

@jdcourcol @mgeplf @GianlucaFicarelli

GianlucaFicarelli:
Note that this is a breaking change for retrieving public resources because to retrieve public entities:
- now the frontend doesn't pass virtual-lab-id/project-id in the header
- with this change, the frontend should specify in the requests the parameter `authorized_public=true` as added in https://github.com/openbraininstitute/prod-platform-architecture/issues/125 when retrieving entities. However, this parameter must not be specified when retrieving global resources (species, strain, brain_region...) because they are public by definition and they don't contain this attribute.

I think the main use case for this proposal is to allow the clients to retrieve all their private entities across multiple projects in a single request. Can this be useful in the new UI or in some planned functionality? Are there other use cases?

Side note just for clarity: the retrieval of resources that are public OR belonging to a specific project cannot be expressed with a regular AND filter, but this isn't going to change with this proposal.

GianlucaFicarelli:
In my previous comment I was considering only the `get_many` endpoint, but it's clearer if we consider each type of endpoint that we have now:

- read_many
- create_one
- read_one/{id}
- update_one/{id}
- delete_one/{id}
- admin_read_one/{id}
- admin_update_one/{id}
- admin_delete_one/{id}

In the endpoints above:

- The `read_many` can return in a single call:
  - public entities belonging to any project, or
  - public OR private entities, both belonging to a specific project, or
  - public entities belonging to any project, OR private entities belonging to a specific project: do we still need that?
[...]

PR #388, opened 2025-10-01
Introduce service maintainer role for update/delete

eleftherioszisis:
Add support for a special Keycloak group `/service/entitycore/maintainer` that allows users to:

* update public entities within authorized projects
* delete public entities within authorized projects
* hard delete assets within authorized projects

To allow hard deletion from regular endpoints a new boolean parameter "hard" has been introduced that only service maintainers and admins can set to True.

### DIFF
--- a/app/db/auth.py
+++ b/app/db/auth.py
@@ -15,38 +15,30 @@ def constrain_to_writable_entities[Q: Query | Select](
     user_context: UserContext,
     db_model_class: Any = Entity,
 ) -> Q:
-    if user_context.project_id:
-        return _constrain_to_writable_project_id(query, db_model_class, user_context)
+    """Constrain query to writable entities.
 
-    return _constrain_to_writable_project_ids(query, db_model_class, user_context)
+    Permisions:
+    - Users have acces to authorized private entities
+    - Maintainers have access to all authorized entities
+    - Admins are not handled by this function and will be treated as regular users
 
-
-def _constrain_to_writable_project_id[Q: Query | Select](
-    query: Q, db_model_class: Any, user_context: UserContext
-) -> Q:
-    if user_context.is_service_maintainer:
-        return query.where(
-            db_model_class.authorized_project_id == user_context.project_id,
-        )
-
-    return query.where(
-        db_model_class.authorized_public == false(),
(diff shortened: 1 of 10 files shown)


jdcourcol (approved):
Thank you. I am good with that.

app/db/auth.py
```diff
@@ -10,6 +10,47 @@
 from app.schemas.auth import UserContext
 
 
+def constrain_to_writable_entities[Q: Query | Select](
+    query: Q,
+    user_context: UserContext,
+    db_model_class: Any = Entity,
+) -> Q:
+    if user_context.project_id:
```
mgeplf, line 18: instead of dispatching on this, can't the there just be a single `_constrain_to_writable_project_ids`, and it makes a iterable out of `user_context.project_id` so it can be used in `in_`?  I would hope that either sqlalchemy or the postgres query planner would see a list of one element and optimize the `in_` into an `==`

Frankly, I'm a bit confused by the existence of both `project_id` and `project_ids`;  could `UserContextWithProjectId` override `project_ids` to make a list of the single one?

app/db/auth.py
```diff
@@ -10,6 +10,47 @@
...
+        return _constrain_to_writable_project_id(query, db_model_class, user_context)
```
mgeplf, line 19: Instead of dispatching on `user_context.project_id` and having two queries, couldn't the scalar one be changed  into a list, and the same `db_model_class.authorized_project_id.in_(project_ids)` be used?

I suspect that the sqlalchemy or postgres query planner would change a list of a single element into a `==`.

I'm also a bit perplexed by the existence of having both `project_id` and `user_project_ids`.  Won't `user_project_ids` always contain `project_id`?

app/db/auth.py
eleftherioszisis: Indeed. I refactored it.

Header context is optional (UserContextWithProjectId is only used in create endpoints. UserContext allows header context to be optional). That means that if the user provides it, then it will constrain queries within that project, otherwise it will use the token ones.

See https://github.com/openbraininstitute/entitycore/issues/372 for the discussion.

mgeplf (approved):

PR #391, opened 2025-10-02
add asset label for ion channel model figure

AurelienJaquier:


eleftherioszisis:
When adding a new label, you need to also add the new label also here along with its requirements:
https://github.com/openbraininstitute/entitycore/blob/1d75af2996f5031359db8b829d808eb0680f17af/app/db/types.py#L507-L511

Issue #389, opened 2025-10-01
Question about electrical-cell-recording


dkeller9:
I am trying to use this code to read from electrical-cell-recording but get an error 401 that the authorization token is missing. I know I am passing a valid token however. Why might this error be occurring?

Code as follows:

```
import requests

# Define the URL and parameters
url = "https://staging.openbraininstitute.org/api/entitycore/electrical-cell-recording/"
params = {
    "page": 1,
}

# Define headers with Bearer token (most common)
headers = {
    "accept": "application/json",
    "Authorization": f"Bearer {token}"
}

# Make the GET request
response = requests.get(url, params=params, headers=headers)

# Check if the request was successful
if response.status_code == 200:
    data = response.json()
    print("Success!")
    print(data)
else:
    print(f"Error: {response.status_code}")
    print(response.text)
```

GianlucaFicarelli:
The issue is that you are calling `electrical-cell-recording/` with a trailing `/`, that is redirected to the correct url `electrical-cell-recording` without trailing `/`, but the auth headers aren't probably sent to the new url.
If you enable debug logging with `logging.basicConfig(level=logging.DEBUG, force=True)` you'll see multiple calls

```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): staging.openbraininstitute.org:443
DEBUG:urllib3.connectionpool:https://staging.openbraininstitute.org:443 "GET /api/entitycore/electrical-cell-recording/?page=1 HTTP/1.1" 307 0
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): staging.openbraininstitute.org:80
DEBUG:urllib3.connectionpool:http://staging.openbraininstitute.org:80 "GET /api/entitycore/electrical-cell-recording?page=1 HTTP/1.1" 301 134
DEBUG:urllib3.connectionpool:https://staging.openbraininstitute.org:443 "GET /api/entitycore/electrical-cell-recording?page=1 HTTP/1.1" 401 96
Error: 401
{"error_code":"NOT_AUTHENTICATED","message":"The authorization token is missing","details":null}
```

If you remove the trailing url you'll see only one call as it should be:
```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): staging.openbraininstitute.org:443
DEBUG:urllib3.connectionpool:https://staging.openbraininstitute.org:443 "GET /api/entitycore/electrical-cell-recording?page=1 HTTP/1.1" 200 507189
```

[...]

PR #395, opened 2025-10-03
Disable redirect_slashes

GianlucaFicarelli:
Disable redirect_slashes because:

- Redirecting doesn't always work well with some clients because of the required headers. For example requests, see https://github.com/openbraininstitute/entitycore/issues/389
- It causes the client to hit the endpoint multiple times (3)

This is a potentially breaking change if any client is misconfigured to use an endpoint with a trailing slash.

cc @bilalesi @pgetta @g-bar @jankrepl @WonderPG @BoBer78

### DIFF
--- a/app/application.py
+++ b/app/application.py
@@ -128,6 +128,7 @@ app = FastAPI(
         StarletteHTTPException: http_exception_handler,
     },
     root_path=settings.ROOT_PATH,
+    redirect_slashes=False,
 )
 app.add_middleware(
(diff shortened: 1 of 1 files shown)


eleftherioszisis (approved):
Thanks. It is strange this is enabled by default in the first place.

mgeplf (approved):
well, this solves a long running problem w/ my muscle memory always putting a trailing slash :)

PR #397, opened 2025-10-03
Fixes for ion channel model asset labes

eleftherioszisis:
* Add ion_channel_model prefix to asset labels
* Make migration


AurelienJaquier (approved):
thanks!

Issue #408, opened 2025-10-16
Add new entity types required for circuit extraction


chr-pok:
New entities required in entitycore for representing circuit extractions:

- **`CircuitExtractionCampaign`**: Represents the obi-one `CircuitExtractionScanConfig` for defining circuit extraction tasks over a range of N parameter value combinations
  - Asset(s): obi-one multi-config file (.json)
- **`CircuitExtractionConfig`**: Represents the obi-one `CircuitExtractionSingleConfig` for defining a single circuit extraction task
  - Reference(s): Input `Circuit`, corresponding `CircuitExtractionCampaign`
  - Asset(s): obi-one single-config file (.json)
- **`CircuitExtractionConfigGeneration(Activity)`**: Activity representing the generation of N `CircuitExtractionConfig` entities from a `CircuitExtractionCampaign` entity
- **`CircuitExtractionExecution(Activity)`**: Activity representing the execution of a single circuit extraction task, i.e., generating an output `Circuit` using a `CircuitExtractionConfig` entitiy
- **`CircuitExtractionExecutionStatus(Enum)`**: Type enum for representing the status of the `CircuitExtractionExecution`

**Note**: The output of a `CircuitExtractionExecution` is a new `Circuit` entity with a derivation of type "circuit_extraction" from the input `Circuit` entity; this is redundant and may be removed in the future

chr-pok:
The required entity types are being implemented in the `entitycore` branch [circuit-extraction](https://github.com/openbraininstitute/entitycore/tree/circuit-extraction). A draft [PR#409](https://github.com/openbraininstitute/entitycore/pull/409) has been created.

evakennyobi:
Is there an update here? thanks

PR #409, opened 2025-10-16
New entity types required for circuit extraction

chr-pok:
The following entity types are required for representing circuit extractions:

- **`CircuitExtractionCampaign`**: Represents the obi-one `CircuitExtractionScanConfig` for defining circuit extraction tasks over a range of N parameter value combinations
  - Asset(s): obi-one multi-config file (.json)
- **`CircuitExtractionConfig`**: Represents the obi-one `CircuitExtractionSingleConfig` for defining a single circuit extraction task
  - Reference(s): Input `Circuit`, corresponding `CircuitExtractionCampaign`
  - Asset(s): obi-one single-config file (.json)
- **`CircuitExtractionConfigGeneration(Activity)`**: Activity representing the generation of N `CircuitExtractionConfig` entities from a `CircuitExtractionCampaign` entity
- **`CircuitExtractionExecution(Activity)`**: Activity representing the execution of a single circuit extraction task, i.e., generating an output `Circuit` using a `CircuitExtractionConfig` entitiy
- **`CircuitExtractionExecutionStatus(Enum)`**: Type enum for representing the status of the `CircuitExtractionExecution`

**Note**: The output of a `CircuitExtractionExecution` is a new `Circuit` entity with a derivation of type "circuit_extraction" from the input `Circuit` entity; this is redundant and may be removed in the future

Related issue: https://github.com/openbraininstitute/entitycore/issues/408

### DIFF
--- a/alembic/versions/20251024_012210_dfe1f3c450ba_circuit_extraction_campaign.py
+++ b/alembic/versions/20251024_012210_dfe1f3c450ba_circuit_extraction_campaign.py
@@ -1,8 +1,8 @@
 """circuit_extraction_campaign
 
-Revision ID: 3dd02d4efbb9
+Revision ID: dfe1f3c450ba
 Revises: 805fc8028f39
-Create Date: 2025-10-24 01:10:55.777114
+Create Date: 2025-10-24 01:22:10.138266
 
 """
 
@@ -17,7 +17,7 @@ from sqlalchemy import Text
 import app.db.types
 
 # revision identifiers, used by Alembic.
-revision: str = "3dd02d4efbb9"
+revision: str = "dfe1f3c450ba"
 down_revision: Union[str, None] = "805fc8028f39"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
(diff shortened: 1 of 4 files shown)


eleftherioszisis:
Does circuit extraction make sense as a campaign? Why not just a single task of extracting a circuit? Does it really need to be a collection?

chr-pok:
> Does circuit extraction make sense as a campaign? Why not just a single task of extracting a circuit? Does it really need to be a collection?

Yes, we would like to follow the same logic as with simulation campaigns and ion channel modelling. We have a form in obi-one for it which already supports parameter scans, like extracting different node sets from a given circuit, or extracting a specific node set from different circuits. So these are use cases we would like to support.

chr-pok:
Hey @eleftherioszisis! I have added filters, services and routers by basically copy-pasting what we have for simulation campaigns so this may require some polishing since I am not really familiar with these. Also, I don't know exactly how to fix the tests that are failing atm, so would be great if you could have a look.

Please note that the DB model for circuit extraction is slightly different (intentionally) in the sense that it does not contain an `entity_id` on the level of the campaign (since a campaign may use different circuits), but a `circuit_id` on the level of a single extraction config.

Moreover, a campaign does not contain references the single extraction config entities and vice-versa, since the idea is that all `CircuitExtractionConfig` entities belonging to a `CircuitExtractionCampaign` should be accessible though its corresponding `CircuitExtractionConfigGeneration` activity, thereby reducing redundancy in the DB model (this was discussed recently).

PR #424, opened 2025-10-27
Ensure that project_id is always considered in read_one

GianlucaFicarelli:
Hotfix based on 2025.10.7, not to be merged but to be cherry picked on the main branch later.
Based on the same approach used for router_update_one, router_delete_one, router_update_activity_one.

Test added.
Opened as draft so it cannot be merged by mistake.


GianlucaFicarelli:
Released as 2025.10.8, closing without merging.

PR #435, opened 2025-11-06
Add skeletonization config and execution

GianlucaFicarelli:
Using the same approach used for:
IonChannelModelingConfig (Entity) -> SkeletonizationConfig
IonChannelModelingExecution (Activity) -> SkeletonizationExecution

The input of SkeletonizationExecution should be SkeletonizationConfig.
The output of SkeletonizationExecution should be a CellMorphology with the right CellMorphologyProtocol.

The campaign can come later and it can be done similarly to 
IonChannelModelingCampaign (Entity)
IonChannelModelingConfigGeneration (Activity)

Since the patterns are similar, we can decide to refactor some of the code later to reuse the common logic.

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -2072,8 +2072,6 @@ class SkeletonizationConfig(Entity, NameDescriptionVectorMixin):
 class SkeletonizationExecution(Activity):
     """Represents the execution of the skeletonization.
 
-    It stores the execution status of an ion channel modeling.
-
     Inputs (used):
         - SkeletonizationConfig
     Outputs (generated):


app/db/model.py
```diff
@@ -2047,3 +2048,46 @@ class AnalysisNotebookResult(Entity, NameDescriptionVectorMixin):
...
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    scan_parameters: Mapped[JSON_DICT] = mapped_column(default={}, server_default="{}")
+
+    __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
+
+
+class SkeletonizationExecution(Activity):
+    """Represents the execution of the skeletonization.
+
+    It stores the execution status of an ion channel modeling.
```
mgeplf: copy pasta?

eleftherioszisis (approved):

PR #445, opened 2025-11-13
Improve performance of the query used for pagination

GianlucaFicarelli:
Fix https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/32
See the issue for more context.


eleftherioszisis:
An order of magnitude! That must have been extremely satisfying. :D
I am good with the changes. I will approve once rebased.

Issue #381, opened 2025-09-29
Add `created_by__sub_id` filter


WonderPG:
Currently, to retrieve an entity that a platform user has created, the frontend must do the following
```
keycloak sub -> GET ALL person -> GET ALL entity belonging to the person
```
since the entities of a user can only be retrieved through the `created_by__id`, which is the `person_id`. This implies one extra request. Assuming all user that want to list their entity in a page have a keycloak sub, it would be useful to be able to retrieve an entity based on the `created_by__sub_id` which would be the keycloak sub that one can infer directly from the bearer token. This way listing entities would come down to:
```
keyckoal sub -> GET ALL entity belonging to the person
```
saving an extra request.

It seems like @bilalesi could be interested by this feature.

evakennyobi:
Can wait for resources for now

eleftherioszisis:
Changes are deployed to staging.

PR #447, opened 2025-11-17
Add created_by__sub_id filter

eleftherioszisis:
#381

### DIFF
--- a/app/filters/base.py
+++ b/app/filters/base.py
@@ -146,7 +146,10 @@ class CustomFilter[T: DeclarativeBase](Filter):
                     if aliases and self.Constants.model in aliases:
                         alias = aliases[self.Constants.model]
                         if isinstance(alias, dict):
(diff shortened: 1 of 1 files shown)


app/filters/base.py
```diff
@@ -145,7 +145,11 @@ def filter[T: DeclarativeBase](self, query: Select[tuple[T]], aliases: Aliases |
                     # { CODE is different from fastapi_filter here
                     if aliases and self.Constants.model in aliases:
                         alias = aliases[self.Constants.model]
-                        model_field = getattr(alias, field_name)
+                        if isinstance(alias, dict):
+                            # When alias is a dict, we don't use it for direct field access
```
mgeplf: I still find this comment confusing; I can see that it's comparing the alias to dict, but I don't understand why `we don't use it for direct field access`

eleftherioszisis: I think it's more clear now. wdyt?

PR #450, opened 2025-11-17
Run locally without OPENAI_API_KEY

mgeplf:
* to run locally without an OPENAI_API_KEY we can set the key to "random", which will create a random value for each embedding that is created

### DIFF
--- a/alembic/versions/20251118_173010_0ce2b30e50d1_add_notice_text.py
+++ b/alembic/versions/20251118_173010_0ce2b30e50d1_add_notice_text.py
@@ -1,0 +1,30 @@
+"""Add notice_text
+
+Revision ID: 0ce2b30e50d1
+Revises: 4b526a1ded96
+Create Date: 2025-11-18 17:30:10.650052
+
+"""
+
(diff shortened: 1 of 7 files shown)


app/utils/embedding.py
GianlucaFicarelli: Shouldn't we add `OPENAI_API_KEY=random` to `.env.run-local` and `.env.run-docker` so the new random generation can work out of the box?

app/utils/embedding.py
```diff
@@ -27,6 +30,8 @@ def generate_embedding(text: str, model: str = "text-embedding-3-small") -> list
             http_status_code=500,
         )
 
+    if settings.OPENAI_API_KEY == "random":
+        return [random.random() for _ in range(EmbeddingMixin.SIZE)]  # noqa: S311
```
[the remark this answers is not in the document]
mgeplf: done

app/utils/embedding.py
mgeplf: yes, it works, values are somewhat non-nonsensical though, which is to be expected.

PR #451, opened 2025-11-18
Add within_brain_region_direction to InBrainRegionQuery

mgeplf:
* instead of a boolean `within_brain_region_ascendants`; this new
  parameter allows for `ascendants`, `descendants`, like before, but
  also `ascendants_and_descendants`.
* `within_brain_region_ascendants` should be deprecated once everything
  depending on it is removed


eleftherioszisis (approved):
lgtm

PR #455, opened 2025-11-20
allow None for temperature in ion channel model

AurelienJaquier:
As described in [this ticket](https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/31), we do not have temperature data for some of the models. So I would like to allow `None` as a value for `temperature_celsius` to represent that.


eleftherioszisis:
> In the filters, we have `temperature_celsius__lte` and `temperature_celsius__gte`. Will allowing `None` as a value create problems with the filters?

Good question. Adding a couple of checks about that in tests will verify and ensure the correct behavior.

In addition, given that you make a change in the database model, you need to make a migration.

```
MESSAGE=optional_temperature make migration
```

Issue #361, opened 2025-09-11
Deletion cascades for entities


eleftherioszisis:
Currently deleting an entity is quite tedious because everything that refers to it needs to be deleted first.

However, one can argue that it would make sense to cascade deletion to some of their associated resources, such as:

* Contributions
* Classifications, both public and private
* Assets
* Annotations
* Measurement Annotations

Especially for assets, currently if the user has deleted (marked as deleted) an asset, it is not possible to find it in order to delete it and remove the foreign key reference to the entity.

@mgeplf @jdcourcol @GianlucaFicarelli

jdcourcol:
I agree. However, how would you "identify" what are the related you need to delete along ?

eleftherioszisis:
Yes, it is deployed to prod. It affects how entities are deleted. Users that are doing deletions using the entitycore/entitysdk API can say if it works as expected. @dkeller9 was interested in this functionality.

PR #458, opened 2025-11-25
Add deletion cascades for entities

eleftherioszisis:
#361 

Deletion cascades added (listed entries will automatically be deleted when foreign key to referenced entity is deleted):

**Entity**
----------
* Contribution
* Annotation
* Measurement Annotation
* MTypeClassification (public & private)
* ETypeClassification (public & private)
* Asset (metadata and s3 stored object)
* Generation (Activity generated_by)
* Derivation if output (already implemented)

**ScientificArtifact**
---------------------
* ScientificArtifactPublicationLink
* ScientificArtifactExternalUrlLink

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1395,7 +1395,8 @@ class Asset(Identifiable):
     meta: Mapped[JSON_DICT]  # not used yet. can be useful?
     label: Mapped[AssetLabel]
     entity_id: Mapped[uuid.UUID] = mapped_column(
-        ForeignKey("entity.id", ondelete="CASCADE"), index=True
+        ForeignKey("entity.id"),  # Asset deletion cascades through ORM app.db.events
+        index=True,
     )
     storage_type: Mapped[StorageType]
 
--- a/alembic/versions/20251222_095114_fa068879e345_deletion_cascades.py
+++ b/alembic/versions/20251222_095114_fa068879e345_deletion_cascades.py
@@ -1,8 +1,8 @@
 """deletion_cascades
 
-Revision ID: 8112fa7ffb32
+Revision ID: fa068879e345
 Revises: 4ff95da5bdf2
-Create Date: 2025-12-18 10:59:21.650527
+Create Date: 2025-12-22 09:51:14.816400
 
 """
 
@@ -16,7 +16,7 @@ from sqlalchemy import Text
 import app.db.types
 
 # revision identifiers, used by Alembic.
-revision: str = "8112fa7ffb32"
+revision: str = "fa068879e345"
 down_revision: Union[str, None] = "4ff95da5bdf2"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
@@ -28,15 +28,6 @@ def upgrade() -> None:
     op.create_foreign_key(
         op.f("fk_annotation_entity_id_entity"),
         "annotation",
-        "entity",
-        ["entity_id"],
-        ["id"],
-        ondelete="CASCADE",
-    )
-    op.drop_constraint(op.f("fk_asset_entity_id_entity"), "asset", type_="foreignkey")
-    op.create_foreign_key(
-        op.f("fk_asset_entity_id_entity"),
-        "asset",
         "entity",
         ["entity_id"],
         ["id"],
@@ -210,10 +201,6 @@ def downgrade() -> None:
     op.drop_constraint(op.f("fk_contribution_entity_id_entity"), "contribution", type_="foreignkey")
     op.create_foreign_key(
         op.f("fk_contribution_entity_id_entity"), "contribution", "entity", ["entity_id"], ["id"]
-    )
-    op.drop_constraint(op.f("fk_asset_entity_id_entity"), "asset", type_="foreignkey")
-    op.create_foreign_key(
-        op.f("fk_asset_entity_id_entity"), "asset", "entity", ["entity_id"], ["id"]
     )
     op.drop_constraint(op.f("fk_annotation_entity_id_entity"), "annotation", type_="foreignkey")
     op.create_foreign_key(
--- a/scripts/export/build_database_archive.sh
+++ b/scripts/export/build_database_archive.sh
@@ -2,7 +2,7 @@
 # Automatically generated, do not edit!
 set -euo pipefail
 SCRIPT_VERSION="1"
-SCRIPT_DB_VERSION="8112fa7ffb32"
+SCRIPT_DB_VERSION="fa068879e345"
 echo "DB dump (version $SCRIPT_VERSION for db version $SCRIPT_DB_VERSION)"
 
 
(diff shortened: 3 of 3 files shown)


app/routers/asset.py
```diff
@@ -291,7 +290,9 @@ def delete_entity_asset(
         asset_id=asset_id,
         hard_delete=hard,
     )
-    asset_service.delete_asset_storage_object(asset, storage_client_factory)
+
+    # Note: Asset storage object is deleted in app.db.events.delete_asset_storage_object_event
+
```
jdcourcol: so here as well: https://github.com/openbraininstitute/entitycore/blob/68fb574ec7370cc9f1f5259b816226ae35762dbc/app/routers/admin.py#L94C3-L94C77

app/db/model.py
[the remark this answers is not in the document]
eleftherioszisis: No, the requirement was to delete a derivation only if its generated is deleted.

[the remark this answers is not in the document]
jdcourcol: That is strange. same for usage.
To me **usage**, **derivation**, and **generation** are relationships. If the entity part is missing, then the relationship should be deleted.

The pattern being:

Entity<-**Usage**->Activity<-**Generation**->Entity

and

Entity<-**Derivation**->Entity

so if you delete an Entity the relationship is useless.


@MWolfR , what do you think ? Am I missing something ?

app/db/events.py
```diff
@@ -0,0 +1,11 @@
...
+
+from app.db.model import Asset
+from app.service.asset import delete_asset_storage_object
+from app.utils.s3 import get_s3_client
+
+
+@event.listens_for(Asset, "after_delete")
+def delete_asset_storage_object_event(_mapper, _connection, target):
+    """Trigger s3 file/directory deletion after an Asset is deleted."""
+    delete_asset_storage_object(asset=target, storage_client_factory=get_s3_client)
```
GianlucaFicarelli: According to https://docs.sqlalchemy.org/en/20/core/event.html#events-and-multiprocessing the callback is executed in a subprocess, is that correct?
What happens if the deletion fails?

app/db/model.py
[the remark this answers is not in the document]
GianlucaFicarelli: This seems important for public entities, that can have been used to generate other private or public entities in different projects.
With the cascade, the deletion of the used entity would automatically trigger the deletion of the links to the activity, changing the `used` relationships of the activity, instead of raising an error.
For private entities, the automatic deletion would affect activities in the same project so it seems less important.

From the last discussion and the document "Entity deletion endpoints" I understood that for **derivations**:
- if the used entity is deleted, an error should be raised
- if the generated entity is deleted, the derivation should be automatically deleted
- if the derivation itself is deleted, nothing else happens (but I don't see an endpoint implementing the deletion yet, it should probably be added)

For **usage**:
- if the used entity is deleted, an error should be raised
- if the activity is deleted, the link between activity and used entity should be automatically deleted

For **generation**:
- if the generated entity is deleted, the link between activity and generated entity should be deleted (although this can leave an activity without output, unless we delete the activity in the application logic)
- if the activity is deleted, the link between activity and generated entity should be deleted (the document says: _Exception 2: When an Activity or Derivation is deleted, all its outputs are deleted, together with their assets_. -> This can be surprising for the user, do we really want to delete the generated entities as well?)

app/db/events.py
```diff
@@ -0,0 +1,41 @@
...
+            delete_asset_storage_object(
+                asset=asset,
+                storage_client_factory=get_s3_client,
+            )
+        except Exception:  # noqa: BLE001
+            L.error(
+                "Failed to delete storage object for Asset id=%s full_path=%s",
+                asset.id,
+                asset.full_path,
+            )
```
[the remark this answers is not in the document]
eleftherioszisis: Indeed. Added it

app/db/events.py
```diff
@@ -0,0 +1,48 @@
...
+from app.logger import L
+from app.utils.s3 import delete_asset_storage_object, get_s3_client
+
+# from app.service.asset import delete_asset_storage_object
+
+ASSETS_TO_DELETE_KEY = "assets_to_delete_from_storage"
+
+
+@event.listens_for(Asset, "before_delete")
+def asset_before_delete(_mapper, _connection, target: Asset):
```
GianlucaFicarelli: Instead of `asset_before_delete` a more meaningful name would be clearer, and a docstring can be added.

app/db/events.py
```diff
@@ -0,0 +1,48 @@
...
+                s3_key=asset.full_path,
+                storage_client_factory=get_s3_client,
+            )
+        except Exception as e:  # noqa: BLE001
+            L.exception(
+                "Failed to delete storage object for Asset id={} full_path={} storage_type={}",
+                asset.id,
+                asset.full_path,
+                asset.storage_type,
+                exc_info=e,
```
GianlucaFicarelli: I don't think `exc_info=e` is needed or useful, since we are already logging the exception with L.exception

app/routers/admin.py
```diff
@@ -89,5 +87,5 @@ def delete_entity_asset(
         entity_id=entity_id,
         asset_id=asset_id,
     )
-    asset_service.delete_asset_storage_object(asset, storage_client_factory)
+    # Note: Asset storage object is deleted in app.db.events.delete_asset_storage_object_event
```
GianlucaFicarelli: The function isn't called app.db.events.delete_asset_storage_object_event anymore. It can be enough to reference app.db.events, or events in general.

tests/db/test_events.py
```diff
@@ -0,0 +1,244 @@
...
+        args, _ = mock_warning.call_args
+        assert "not attached to a session" in args[0]
+
+
+@pytest.fixture
+def mock_storage_delete():
+    """Patch storage deletion and S3 client creation."""
+    with (
+        patch("app.db.events.delete_asset_storage_object") as mock_delete,
+        patch("app.utils.s3.get_s3_client", return_value=Mock()),
```
GianlucaFicarelli: Minor, but for consistency delete_asset_storage_object, get_s3_client can be mocked from the same module 
app.db.events since this is the module being tested.

app/db/events.py
GianlucaFicarelli: Actually, with loguru the message would be like:

> 2025-12-19 10:46:07.374 | ERROR    | app.db.events:delete_assets_from_storage:44 - Failed to delete storage object for Asset id=310e9c8b-c778-433e-9fb6-2501b6a42162 full_path=public/a98b7abc-fc46-4700-9e3d-37137812c730/0dbced5f-cc3d-488a-8c7f-cfb8ea039dc6/assets/reconstruction_morphology/ff4e7aa7-0ce1-419e-90c4-9903b1a2bb00/AA1002.h5 storage_type=aws_s3_internal [exc_info=name 'aaaa' is not defined]
Traceback (most recent call last):
>....

However, if we want to log the str or repr of the exception in the message before the traceback, it would be better to explicitly add it to the message.

app/routers/admin.py
eleftherioszisis: fixed

tests/db/test_events.py
eleftherioszisis: fixed

PR #467, opened 2025-12-03
Add ilike_search for name and description

eleftherioszisis:
* Use fastapi-filter ilike's search with multiple columns.
* Create filter mixin that adds the filter parameter "ilike_search" and modifies fastapi-filter's `Constants.search_field_name` to "ilike_search" and `Constants.search_model_fields` to `["name", "description"]`
* Allow overriding the default `search_model_fields` by explicitly listing the desired ones in `cls.Constants`

Related issue: https://github.com/openbraininstitute/prod-explore-functionality/issues/358

### DIFF
--- a/app/filters/common.py
+++ b/app/filters/common.py
@@ -47,6 +47,8 @@ class ILikeSearchFilterMixin:
             # otherwise set default fields
             cls.Constants.search_model_fields = ILIKE_SEARCH_FIELDS  # pyright: ignore [reportAttributeAccessIssue]
 
+        search_model_fields = cls.Constants.search_model_fields  # pyright: ignore [reportAttributeAccessIssue]
+
         # Add filter key and annotation for ilike_search
         cls.__annotations__[ILIKE_SEARCH_FIELD_NAME] = str | None
 
@@ -60,7 +62,7 @@ class ILikeSearchFilterMixin:
                         "and ? for exactly one character. All other characters are treated as "
                         "literals. "
                         "Examples: 'test*' matches 'testing', 'file?.txt' matches 'file1.txt'. "
-                        f"search_model_fields: {', '.join(cls.Constants.search_model_fields)}"
+                        f"search_model_fields: {', '.join(search_model_fields)}"
                     ),
                     default=None,
                 ),


jdcourcol:
how does that work if we want to use and additional property for a type. say I want the ilike to apply to the `protocol_document` property too for `CellMorphologyProtocol`

jdcourcol:
I have another comment. My understanding is that we add a "%" at the beginning and at the end of the provided string.
Can the user provide something like  "foo%ar", that would match "foobar", "fooBar", "foocar"...
If so , can he escape the "%" character ?

jdcourcol (approved):
I am fine. I don't know if you want someone else to review though.

app/filters/common.py
```diff
@@ -29,6 +29,29 @@ class NameFilterMixin:
...
+        """Add ilike search on multiple columns.
+
+        Ensure that this mixin is added to a filter corresponding to a model with
+        ILIKE_SEARCH_FIELDS available.
+        """
+        # Add filter key and annotation for ilike_search
+        cls.__annotations__[ILIKE_SEARCH_FIELD_NAME] = str | None
+        setattr(cls, ILIKE_SEARCH_FIELD_NAME, None)
+
+        # Set fastapi-filter custom serach name and fields
```
GianlucaFicarelli: Minor, typo
```suggestion
        # Set fastapi-filter custom search name and fields
```

app/filters/common.py
```diff
@@ -29,6 +29,29 @@ class NameFilterMixin:
...
+        # See example: https://github.com/arthurio/fastapi-filter/blob/8c07dd55dfa63f09ae70eb980d51714323809906/examples/fastapi_filter_mongoengine.py#L91-L92
+        cls.Constants.search_field_name = ILIKE_SEARCH_FIELD_NAME  # pyright: ignore [reportAttributeAccessIssue]
+
+        # allow search_model_fields overrides
+        if hasattr(cls.Constants, "search_model_fields") and cls.Constants.search_model_fields:  # pyright: ignore [reportAttributeAccessIssue]
```
GianlucaFicarelli: Minor, it should be equivalent to:

```suggestion
        if getattr(cls.Constants, "search_model_fields", None):
```

app/filters/cell_morphology_protocol.py
```diff
@@ -39,6 +40,11 @@ class Constants(CellMorphologyProtocolFilterBase.Constants):
             "creation_date",
             "update_date",
         ]
+        search_model_fields = [  # noqa: RUF012
+            "name",
+            "description",
+            "protocol_document",
```
GianlucaFicarelli: For confirmation, I see in the comments that overriding `search_model_fields` in CellMorphologyProtocolFilter is an example. Do we need to keep it?

Issue #460, opened 2025-11-28
Redesign deletion permissions


eleftherioszisis:
This is a spinoff ticket from #361 

To simplify the user experience, I propose removing soft deletes as well as the hard-delete toggle. All deletions would instead be performed as hard deletes, governed by the following deletion permission rules:

**Deletion Permission Rules**
* Project admins may delete any private entities within their projects.
* Service maintainers may delete private or public entities within their projects.
* Project members may delete only the private entities they themselves created.

[Proposal Document](https://openbraininstitute-my.sharepoint.com/:w:/g/personal/eleftherios_zisis_openbraininstitute_org/IQAr67NEgkx4TLMkFKRq9_c6AckgvaKaabb4ZQJW66nL-Sw?e=3L7h8E)

PR #479, opened 2025-12-15
Redesign deletion permissions

eleftherioszisis:
* Change deletion permissions according to #460 
* Separate user and admin common delete routers
* Update asset deletion and remove "hard" deletion switch. All asset deletion will be hard governed by the new permissions rules

### DIFF
--- a/tests/routers/test_asset.py
+++ b/tests/routers/test_asset.py
@@ -805,6 +805,31 @@ def test_delete_entity_asset__authorization__admin(
         assert ErrorResponse.model_validate(data).error_code == expected_error
 
 
+@pytest.mark.parametrize(
+    ("is_public", "expected_status_code", "expected_error"),
+    [
+        (False, 200, None),
+        (True, 403, ApiErrorCode.ENTITY_FORBIDDEN),
+    ],
+)
+def test_delete_entity_asset__authorization__user_own(
+    clients, create_entity_asset, is_public, expected_status_code, expected_error
+):
+    """Test project member may deleted their own private assets."""
+    entity_id, entity_type, asset_id = create_entity_asset(
+        clients.user_3, authorized_public=is_public
+    )
+
+    data = assert_request(
+        client_method=clients.user_3.delete,
+        url=f"{route(entity_type)}/{entity_id}/assets/{asset_id}",
+        expected_status_code=expected_status_code,
+    ).json()
+
+    if expected_error is not None:
+        assert ErrorResponse.model_validate(data).error_code == expected_error
+
+
 def test_delete_entity_asset(client, entity, asset):
     response = client.delete(f"{route(entity.type)}/{entity.id}/assets/{asset.id}")
     assert response.status_code == 200, f"Failed to delete asset: {response.text}"


app/schemas/auth.py
```diff
@@ -52,6 +53,35 @@ def from_jwt(cls, token: HTTPAuthorizationCredentials) -> Self | None:
         return None
 
 
+class UserRole(StrEnum):
```
GianlucaFicarelli: Maybe UserProjectRole instead of UserRole, to be more specific about it?

app/schemas/auth.py
```diff
@@ -170,7 +203,22 @@ class UserContext(UserContextBase):
...
-    user_project_ids: list[UUID] = []
+    user_project_groups: list[UserProjectGroup] = []
+
+    @property
+    def user_project_ids(self) -> list[UUID]:
+        """Return all projects that the user has access to regardless of role."""
+        return [g.project_id for g in self.user_project_groups]
+
+    @property
+    def project_member_ids(self):
```
GianlucaFicarelli: The names `project_member_ids` and `project_admin_ids` are not immediately clear to me that they contain project ids, maybe we can find clearer names?

tests/utils.py
```diff
@@ -1283,9 +1275,9 @@ def _create_model_id(client, data):
 
     def _assert_not_found(client, model_id):
         data = assert_request(
-            client.delete, url=f"{route}/{model_id}", expected_status_code=404
+            client.delete, url=f"{route}/{model_id}", expected_status_code=403
         ).json()
-        assert data["error_code"] == "ENTITY_NOT_FOUND"
+        assert data["error_code"] == "ENTITY_FORBIDDEN"
```
GianlucaFicarelli: The (private) method is called _assert_not_found, but now it checks that the status is 403

eleftherioszisis: done

app/db/auth.py
```diff
@@ -18,7 +18,6 @@ def constrain_to_writable_entities[Q: Query | Select](
     """Constrain query to writable entities.
 
     Permisions:
-    - Users have acces to authorized private entities
     - Maintainers have access to all authorized entities
```
GianlucaFicarelli: Not changed in this PR, but it would be clearer something like "Service maintainers have write access to all the entities in the allowed projects"

app/queries/common.py
```diff
@@ -451,33 +458,66 @@ def router_delete_one[T: BaseModel, I: Identifiable](
...
-    if user_context and (
-        id_model_class := get_declaring_class(db_model_class, "authorized_project_id")
-    ):
-        query = constrain_to_writable_entities(
-            query=query,
-            user_context=user_context,
-            db_model_class=id_model_class,
+    obj = crud.get_identifiable_one(db=db, db_model_class=db_model_class, id_=id_)
+
+    if user_context and not _is_user_authorized_for_deletion(db, user_context, obj):
```
GianlucaFicarelli: `if user_context` seems redundant now that user_context cannot be None

eleftherioszisis: Ah yes. I forgot to remove it, thanks!

GianlucaFicarelli (approved):

PR #492, opened 2025-12-23
Use scope=function in get_db dependency

GianlucaFicarelli:
This change might need more discussion and testing, but it's mainly intended to ensure that:
- in case of error during the commit, the client knows that the operation failed
- in case of deletion of assets from S3 via events, the client can be sure that the deletion operation has finished, so it's possible to upload a new asset with the same name without risk of race conditions


mgeplf (approved):
Agreed, I'm wondering why this isn't the default.

PR #493, opened 2026-01-05
add `number_neurons` to Simulation`

mgeplf:
* this will be used to estimate the duration and cost of the simulation, as well as the number of nodes to be used
* should be added by whatever is creating the simulation campaign/simulations; for example obi-one

### DIFF
--- a/alembic/versions/20260106_092249_316cadee1c4f_add_number_neurons_to_simulation.py
+++ b/alembic/versions/20260106_092249_316cadee1c4f_add_number_neurons_to_simulation.py
@@ -25,22 +25,26 @@ depends_on: Union[str, Sequence[str], None] = None
 def upgrade() -> None:
     op.add_column("simulation", sa.Column("number_neurons", sa.BigInteger(), nullable=True))
 
-    op.execute(text("""
+    op.execute(
+        text("""
         UPDATE simulation
         SET number_neurons = 1
         WHERE entity_id IN (
             SELECT id FROM entity WHERE type = 'memodel'
         )
-    """))
+    """)
+    )
 
-    op.execute(text("""
+    op.execute(
+        text("""
         UPDATE simulation
         SET number_neurons = circuit.number_neurons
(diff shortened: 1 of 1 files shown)


app/db/model.py
```diff
@@ -1548,6 +1548,7 @@ class Simulation(Entity, NameDescriptionVectorMixin):
           belongs to.
         entity_id (uuid.UUID): Foreign key referencing the entity ID.
         entity (Entity): The entity this simulation is associated with.
+        number_neurons (int): Number of neurons to be simulated
```
GianlucaFicarelli: Wouldn't be `number_of_neurons` more natural than `number_neurons`, although longer?
Is it preferred to use the term neurons, or should we use nodes?

alembic/versions/20260106_092249_316cadee1c4f_add_number_neurons_to_simulation.py
```diff
@@ -0,0 +1,55 @@
...
+
+    op.execute(
+        text("""
+        UPDATE simulation
+        SET number_neurons = circuit.number_neurons
+        FROM circuit, entity
+        WHERE simulation.entity_id = circuit.id
+        AND simulation.entity_id = entity.id
+        AND entity.type = 'circuit'
+    """)
```
GianlucaFicarelli: Is it correct to set simulation.number_neurons = circuit.number_neurons?
I was thinking that the number of neurons used in the simulation can be less than the ones in the circuit.

mgeplf: That is true, but I don't have a good way to query how many are actually used without loading the circuit, so this is an upper bound for the simulations that exist.

To get the real number, I think we'd have to add libsonata, and probably some other code, to figure out what the actual values are - since this is mainly going to be used going forward, I'm not sure it's worth the effort to have an accurate tally - what do you think?

PR #498, opened 2026-01-06
Fix incorrect returned values when getting `ascendants_and_descendants`

mgeplf:
* When `ascendants_and_descendants` are chosen, and there are entities in the precise `within_brain_region_brain_region_id`, they would be retrieved twice - once from `ascendants`, and again from the `descendants`
* pagination would limit the returned objects, but could include the duplicates, which were then removed by the `unique` in `db.execute(data_query).scalars().unique()`
* this resulted in the number of objects returned being lower than what was requested in the `page_size`

### DIFF
--- a/app/filters/brain_region.py
+++ b/app/filters/brain_region.py
@@ -49,7 +49,7 @@ def get_family_query(
                     direction=WithinBrainRegionDirection.ascendants,
                 ).c.id
             )
-            .union_all(
+            .union(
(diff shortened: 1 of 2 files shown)


bilalesi (approved):
I was skeptical about `union_all`, thanks @mgeplf

GianlucaFicarelli (approved):

PR #504, opened 2026-01-12
Implement s3 Delegation & Multipart Upload for files.

eleftherioszisis:
Multipart Asset Upload Flow
**1. Initiate Upload**
`POST /{entity_type}/{entity_id}/assets/upload/initiate`
Initiates an S3 multipart upload and returns presigned URLs for each part.
A database asset is created with status = "uploading" and upload_meta containing the S3 upload_id and part metadata.
**2. Upload Parts**
The client uploads file parts directly to S3 using the provided presigned URLs.
**3. Complete Upload**
`POST /{entity_type}/{entity_id}/assets/{asset_id}/upload/complete`
Completes the multipart upload.
The asset is updated to status = "created" and upload_meta is cleared.
**4. Delete / Abort Upload**
`DELETE /{entity_type}/{entity_id}/assets/{asset_id}`
Deletes the asset record triggering the asset deletion event. If the asset is in the uploading state, the corresponding S3 multipart upload is aborted.

### DIFF
--- a/app/service/asset.py
+++ b/app/service/asset.py
@@ -392,9 +392,15 @@ def entity_asset_upload_initiate(
     content_type: str,
     preferred_part_count: int,
 ) -> AssetReadWithUploadMeta:
-    """Initiate multipart upload.
+    """Initiate a multipart upload for an existing entity asset.
 
-    Note: This function does not check user permissions.
+    This function prepares and starts a multipart upload to S3 for the specified
+    asset and returns the asset metadata along with upload details (e.g., upload ID,
+    part size configuration).
+
+    It is intended to be called immediately after the asset record has been
+    created in the router. Because of that, it does not perform any additional
+    existence or permission checks.
     """
     with ensure_result(f"Asset {asset_id} not found", error_code=ApiErrorCode.ASSET_NOT_FOUND):
         asset = repos.asset.get_entity_asset(
@@ -456,6 +462,16 @@ def entity_asset_upload_complete(
     storage: StorageUnion,
     s3_client: S3Client,
 ) -> AssetRead:
+    """Initiate a multipart file upload for an existing entity asset.
+
+    This function prepares and starts a multipart upload to S3 for the specified
+    asset and returns the asset metadata along with upload details (e.g., upload ID,
+    part size configuration).
+
+    It is intended to be called immediately after the asset record has been
+    created in the router. Because of that, it does not perform any additional
+    existence or permission checks.
+    """
     asset_db = get_writable_entity_db_asset(
         repos,
         user_context=user_context,


app/schemas/asset.py
```diff
@@ -175,3 +187,27 @@ class DetailedFileList(BaseModel):
...
+        str | None,
+        Field(
+            description=(
+                "Content type of file. "
+                "If not provided it will be deduced from the file's extension."
+            )
+        ),
+    ] = None
+    label: AssetLabel
+    preferred_part_count: Annotated[int, Field(description="Hint of desired part count.")] = (
```
mgeplf: is this hint needed? I'm not sure we need this flexibility.

app/utils/common.py
```diff
@@ -5,3 +5,8 @@ def is_ascii(s: str) -> bool:
     except UnicodeEncodeError:
         return False
     return True
+
+
+def clip(x: int, min_value: int, max_value: int) -> int:
+    """Clamp x to the inclusive range [min_value, max_value]."""
```
mgeplf: is it clip or clamp?

app/config.py
```diff
@@ -48,6 +48,13 @@ class Settings(BaseSettings):
     S3_MULTIPART_THRESHOLD: int = 5 * 1024**2  # bytes  # TODO: decide an appropriate value
     S3_PRESIGNED_URL_EXPIRATION: int = 600  # seconds  # TODO: decide an appropriate value
 
+    S3_MULTIPART_UPLOAD_MAX_SIZE: int = 5 * 1024**3  # TODO: Set appropriate upper file size limit
+    S3_MULTIPART_UPLOAD_MIN_PART_SIZE: int = 5 * 1024**2
+    S3_MULTIPART_UPLOAD_MAX_PART_SIZE: int = 5 * 1024**3
```
mgeplf: this seems very high, no?

app/service/asset.py
```diff
@@ -344,3 +353,139 @@ def list_directory(
...
+    )
+
+    # parts that are expected to be uploaded via the presigned urls
+    parts = asset.upload_meta.parts
+
+    # verify that expected and uploaded agree
+    uploaded_part_numbers = {p["PartNumber"] for p in uploaded_parts}
+    expected_part_numbers = {p.part_number for p in parts}
+
+    if uploaded_part_numbers != expected_part_numbers:
```
GianlucaFicarelli: This seems a good point where to check if the declared filesize matches with the actual filesize.
If they don't match, it seems ok to update the asset metadata with the real value.
Ideally, we should update also the sha256 if it's available in the S3 metadata, although I don't think it is.

app/utils/s3.py
```diff
@@ -166,6 +172,100 @@ def generate_presigned_url(
...
+):
+    return s3_client.generate_presigned_url(
+        "upload_part",
+        Params={
+            "Bucket": bucket,
+            "Key": s3_key,
+            "UploadId": upload_id,
+            "PartNumber": part_number,
+        },
+        ExpiresIn=settings.S3_PRESIGNED_URL_EXPIRATION,
```
GianlucaFicarelli: Maybe the default S3_PRESIGNED_URL_EXPIRATION of 600 seconds can be too low for multipart uploads?

app/routers/asset.py
GianlucaFicarelli: This can be discussed separately, but maybe we want to support multipart uploads for big files in directory uploads? The files are already uploaded directly to s3, so it seems lower priority, unless there is some size limitation on s3 directly.

app/routers/asset.py
```diff
@@ -335,23 +344,105 @@ def entity_asset_directory_list(
...
+    # create asset to fail early if full path already in progress or registered
+    asset_read = asset_service.create_entity_asset(
+        repos=repos,
+        user_context=user_context,
+        entity_type=entity_type,
+        entity_id=entity_id,
+        filename=json_model.filename,
+        content_type=content_type,
+        size=json_model.filesize,
+        sha256_digest=json_model.sha256_digest,
```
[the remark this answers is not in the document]
eleftherioszisis: I added a check for the sum of the uploaded part sizes against the one provided when uploading the file. With the sha256 I couldn't find a good way because the final SHA (if configured to calculate one) is a composite of the part SHAs and not of the entire file object.

app/routers/asset.py
```diff
@@ -225,6 +228,12 @@ def download_entity_asset(
         entity_id=entity_id,
         asset_id=asset_id,
     )
+    if asset.status == AssetStatus.UPLOADING:
+        raise ApiError(
+            message="Cannot download an uploading asset, because it is incomplete.",
+            error_code=ApiErrorCode.ASSET_UPLOAD_INCOMPLETE,
+            http_status_code=HTTPStatus.FORBIDDEN,
```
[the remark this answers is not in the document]
eleftherioszisis: Fixed

app/service/asset.py
```diff
@@ -344,3 +375,140 @@ def list_directory(
...
+    s3_client: S3Client,
+    s3_key: str,
+    bucket: str,
+    filesize: int,
+    content_type: str,
+    preferred_part_count: int,
+) -> AssetRead:
+    """Initiate multipart upload.
+
+    Note: This function does not check user permissions.
```
mgeplf: Maybe mention why it's safe to not check permissions? where should they be checked?

mgeplf (approved):

app/service/asset.py
```diff
@@ -344,3 +377,146 @@ def list_directory(
...
+    """
+    with ensure_result(f"Asset {asset_id} not found", error_code=ApiErrorCode.ASSET_NOT_FOUND):
+        asset = repos.asset.get_entity_asset(
+            entity_type=entity_type,
+            entity_id=entity_id,
+            asset_id=asset_id,
+        )
```
GianlucaFicarelli: In theory there shouldn't be the need to read the db asset  since it has been just created.
Have you considered already the possibility to reuse that asset?
However, it could need some changes because the function `create_entity_asset` returns `AssetRead` and not the db model `Asset`. It was done in this way so that the service layer returns pydantic model without exposing db models from the repository layer.

app/db/events.py
```diff
@@ -22,22 +23,55 @@ def collect_asset_for_storage_deletion(_mapper, _connection, target: Asset):
...
+    Note: Due to the nature of the operation that iterates over all assets it is important to not
+    throw an error even if one of the external side-effect fail. Otherwise, after the rollback
+    there might be db assets that are not deleted but their s3 files are.
+
+    Instead with capturing the errors it is ensured that db assets are always deleted even if that
+    may result in orphan files or multipart uploads that have failed to be deleted.
+
+    TODO: Add a cleanup function on a schedule that would remove s3 orphans from time to time.
+    """
     to_delete = session.info.pop(ASSETS_TO_DELETE_KEY, set())
```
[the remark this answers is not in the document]
eleftherioszisis: Agreed. Added it.

app/routers/asset.py
GianlucaFicarelli: Now the get many assets endpoint will return assets in both status CREATED and UPLOADING.
It would be useful to state it explicitly in a docstring (especially if it's reflected in the openapi docs), because a client looping over the assets could fail to download an asset that is still uploading.

GianlucaFicarelli (commented):
LGTM but I'll let Mike review the last changes.

PR #510, opened 2026-01-16
Auditing/versioning v2

GianlucaFicarelli:
This PR adds auditing/versioning using an external library sqlalchemy-history (fork of sqlalchemy-continuum).
Only sqlalchemy-continuum would support native_versioning (faster, using potgresql triggers), but the defined triggers have some limitations with polymorphism (they don't create a version of all the tables, and they don't set transaction_end on the previous version), so they would require some changes.

To consider all the deletions through the sqlalchemy models, the ON DELETE CASCADE defined in the db schema have been removed and replaced with a corresponding relationship with cascade.
This is less efficient, especially when deleting entities with many children, but the deletion is not the usual case so it could be acceptable. We can also reconsider the use of triggers later.

The tests are a lot slower because we have twice the number of tables, and we truncate them after each test. Using rollback instead of truncate could improve the performance, but it requires some changes in the instantiation of the sessions during tests.
Alternatively, we can disable versioning during tests that don't check versioning, and truncate only the non-version tables.

### DIFF
--- a/app/service/versioning.py
+++ b/app/service/versioning.py
@@ -65,7 +65,6 @@ from app.schemas.measurement_label import MeasurementLabelRead
 from app.schemas.memodel_calibration_result import MEModelCalibrationResultRead
 from app.schemas.publication import PublicationRead
 from app.schemas.role import RoleRead
-from app.schemas.scientific_artifact import ScientificArtifactRead
 from app.schemas.scientific_artifact_external_url_link import ScientificArtifactExternalUrlLinkRead
 from app.schemas.scientific_artifact_publication_link import ScientificArtifactPublicationLinkRead
 from app.schemas.simulation import (
@@ -137,7 +136,6 @@ RESOURCE_TYPE_TO_READ_SCHEMA = {
     "person": PersonRead,
     "publication": PublicationRead,
     "role": RoleRead,
-    "scientific_artifact": ScientificArtifactRead,
     "scientific_artifact_external_url_link": ScientificArtifactExternalUrlLinkRead,
     "scientific_artifact_publication_link": ScientificArtifactPublicationLinkRead,
     "simulation": SimulationRead,
--- a/app/utils/routers.py
+++ b/app/utils/routers.py
@@ -8,6 +8,7 @@ NOT_ROUTABLE_RESOURCES = {
     EntityType.analysis_software_source_code,
     EntityType.electrical_recording,
     EntityType.me_type_density,
+    EntityType.scientific_artifact,
     GlobalType.ion,
 }
 
--- a/tests/test_routes.py
+++ b/tests/test_routes.py
@@ -49,6 +49,9 @@ def routes():  # noqa: C901
             "/error",
             "/openapi.json",
             "/",
+            "/{resource_route}/{resource_id}/version-count",
+            "/{resource_route}/{resource_id}/version/{version_num}",
+            "/{resource_route}/{resource_id}/version/{version_num}/changeset",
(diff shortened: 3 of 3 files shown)


app/routers/cell_morphology.py
```diff
@@ -20,3 +20,7 @@
 admin_update_one = admin_router.patch(f"/{ROUTE}/{{id_}}")(
     app.service.cell_morphology.admin_update_one
 )
+
+read_mtypes_history = router.get("/{id_}/mtypes/history")(
+    app.service.cell_morphology.read_mtypes_history
+)
```
GianlucaFicarelli: This is just an example, it's going to be removed.

pyproject.toml
```diff
@@ -60,6 +61,9 @@ include = ["app"]
 [tool.uv]
 required-version = ">=0.8.0"
 
+[tool.uv.sources]
+sqlalchemy-history = { git = "https://github.com/GianlucaFicarelli/sqlalchemy-history", branch = "obi" }
```
mgeplf: should this be forked into OBI?

app/db/model.py
```diff
@@ -78,10 +81,23 @@
...
 
+# enable versioning
+make_versioned(
+    user_cls="Person",
+    plugins=[
+        UserContextPlugin(),
+        TransactionChangesPlugin(),
+    ],
+    options={
+        "enum_prefix": "",
```
mgeplf: can you add a comment what this `enum_prefix` is for?

app/db/utils.py
```diff
@@ -22,14 +22,14 @@
     for mapper in Base.registry.mappers
     if issubclass(mapper.class_, MeasurableEntityMixin)
     and issubclass(mapper.class_, Entity)
-    and mapper.class_.__tablename__
+    and getattr(mapper.class_, "__tablename__", None)
```
mgeplf: is this to skip the versioned table? Or something else?

Dockerfile
```diff
@@ -18,7 +18,8 @@ apt-get install -qyy \
     gcc \
     libc6-dev \
     libpq-dev \
-    ca-certificates
+    ca-certificates \
+    git
```
mgeplf: this is for convenience, right? not something to do w/ versioning?

app/db/utils.py
```diff
@@ -68,13 +68,14 @@ def is_polymorphic_subclass(mapper_class: type[Identifiable]) -> bool:
         return False
 
     # If polymorphic_identity is different from __tablename__, it's a subclass
-    return mapper_args["polymorphic_identity"] != mapper_class.__tablename__
+    return mapper_args["polymorphic_identity"] != getattr(mapper_class, "__tablename__", None)
```
eleftherioszisis: Would it worth to create little helper funcs with docstrings to make these more readable?

app/dependencies/db.py
```diff
@@ -23,5 +30,6 @@ def _get_repo_group(db: "SessionDep") -> RepositoryGroup:
 # operation function is finished, before the response is sent back to the client.
 # In this way, the commit of the db transaction and the associated events
 # are executed before the response is sent back to the client.
-SessionDep = Annotated[Session, Depends(get_db, scope="function")]
+PlainSessionDep = Annotated[Session, Depends(get_db, scope="function")]
+SessionDep = Annotated[Session, Depends(get_db_with_user_context, scope="function")]
```
eleftherioszisis: Maybe UserSessionDep?

Issue #511, opened 2026-01-19
Unify execution status across different activities


eleftherioszisis:
Use the same activity execution status for all activities:

```python
class ActivityStatus:
    created = auto()
    pending = auto()
    running = auto()
    done = auto()
    error = auto()
    cancelled = auto()
```

PR #513, opened 2026-01-20
Unify activity status

eleftherioszisis:
Activity status moved to activity table and unified with ActivityStatus:

```
class ActivityStatus(StrEnum):
    created = auto()
    pending = auto()
    running = auto()
    done = auto()
    error = auto()
    cancelled = auto()
```

Issue: #511

**Activities the status of which moved to activity parent table:**
* circuit_extraction_execution
* skeletonization_execution
* ion_channel_modeling_execution
* simulation_execution

**Status columns removed**
* single_neuron_simulation.status
* single_neuron_synaptome_simulation.status

**Activities without prior status column set to "done" because activity parent table has a status now**
 * analysis_notebook_execution
 * calibration
 * circuit_extraction_config_generation
 * ion_channel_modeling_config_generation
 * simulation_generation
 * skeletonization_config_generation
 * validation

Given that this is a breaking change, fyi @james-isbister @chr-pok @AurelienJaquier @ilkilic

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1104,7 +1104,6 @@ class SingleNeuronSimulation(LocationMixin, NameDescriptionVectorMixin, Entity):
     seed: Mapped[int]
     injection_location: Mapped[STRING_LIST] = mapped_column(default=[])
     recording_location: Mapped[STRING_LIST] = mapped_column(default=[])
-    status: Mapped[ActivityStatus]
     # TODO: called used ?
     me_model_id: Mapped[uuid.UUID] = mapped_column(
         ForeignKey(f"{EntityType.memodel}.id"), index=True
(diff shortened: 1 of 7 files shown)


app/db/model.py
```diff
@@ -765,10 +759,7 @@ class MEModel(
     __tablename__ = EntityType.memodel.value
     id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
 
-    validation_status: Mapped[ValidationStatus] = mapped_column(
-        Enum(ValidationStatus, name="me_model_validation_status"),
-        default=ValidationStatus.created,
-    )
+    validation_status: Mapped[ActivityStatus]
```
GianlucaFicarelli: validation_status is an attribute of me_model, and not the status of an activity, so it seems strange to define it on the entity and use the same enum.
Where is validation_status used and how is it updated?
Should it be replaced with ValidationResult?

jdcourcol: We will have to ask @pgetta . I don't see a place in the code where that flag is being updated.

jdcourcol: later

eleftherioszisis: ok, I reverted validation_status as requested.

PR #533, opened 2026-02-19
Pin alembic<1.18 for compatibility with alembic-postgresql-enum

GianlucaFicarelli:
```
See https://github.com/Pogchamp-company/alembic-postgresql-enum/issues/126
```


mgeplf (approved):
eeeeh, rough

PR #536, opened 2026-02-20
Log request context in a middleware

GianlucaFicarelli:
Since the previous [PR](https://github.com/openbraininstitute/entitycore/pull/535) added context to the application logs but not to the uvicorn logs, this PR adds a middleware for modifying the context and logging the requests, and changes the level of the uvicorn logger from INFO to WARNING to avoid logging duplicate information.

It's not possible to use the logger_context dependency as it was done in the previous PR, because it affects only the application logs emitted in the function using that dependency, so it won't add context to the logs in the middleware or uvicorn.

Now the logs are enriched with the following:

- Extra logging keys for non-authenticated endpoints:
    - request_id
- Extra logging keys for authenticated endpoints:
    - request_id
    - user_id

In addition to the above keys, thanks to the middleware the log of the request includes the following keys in the extra dict:
- forwarded_for
- process_time
- status_code
- client

In this way, when the logs are emitted as json, it should be easier to process them to extract the above information.

We might also consider adding other tools like opentelemetry in the future.

### DIFF
--- a/app/logger.py
+++ b/app/logger.py
@@ -87,6 +87,11 @@ def configure_logging() -> int:
     """Configure logging."""
 
     def patcher(record: "loguru.Record") -> None:
+        """Add request context (request_id, user_id) to all log records.
+
+        This function is automatically applied to every log message across all modules,
+        enriching them with contextual information from the current request.
+        """
         ctx = request_context_provider.get({})
         record["extra"].update(ctx)
 


app/logger.py
```diff
@@ -84,6 +85,11 @@ def str_formatter(record: "loguru.Record") -> str:
 
 def configure_logging() -> int:
     """Configure logging."""
+
+    def patcher(record: "loguru.Record") -> None:
```
mgeplf: maybe add a doc on what this does, and where it's invoked?

Issue #523, opened 2026-02-03
Generic Campaign, Config, ConfigGeneration, Execution entities


james-isbister:
## Description
As discussed in the product meeting, we would like to consider the possibility of creating generic Campaign, Config, ConfigGeneration and Execution entities which can be used by any task, with an enum specifying the type of task. 

For example, rather than having CircuitExtractionCampaign, CircuitExtractionConfigGeneration, CircuitExtractionConfig and CircuitExtractionExecution, we would have, say:

Campaign
- task_type: "CircuitExtraction"

ConfigGeneration
- task_type: "CircuitExtraction"

Config
- task_type: "CircuitExtraction"
- entities: list[Entity]

Execution
- task_type: "CircuitExtraction"

The idea is to first transfer to using such entities for tasks which have not yet been used in production. So: CircuitExtraction, Skeletonization, and we leave the entities related to Simulations and IonChannelModeling as they are for now. We also add enum keys for EMSynapseMapping and IonChannelSimulation as useful test cases. IonChannelSimulation for example, requires that Config.entities to be a list for example.

**Possible questions:**
- It would probably be redundant specifying task_type for all the 4 entity types. But at the same time it might be useful to query for any of these 4 entity types (i.e. Campaign, ConfigGeneration, ..) with a given task_type. Or maybe they're only needed for a subset of the 4 entity types?

- Is it sufficient that the Config points to a list of "entities", and the Campaign does not. We reasoned that the entities list is not used in the ConfigGeneration step and staging a part subset of coordinates in a Campaign might only need to stage the "entities" specific to those coordinates (so the Campaign.entities is redundant on its own). CircuitExtraction already follows this pattern - not having an entities property on the Campaign - so maybe people have already thought about this.

- Perhaps the initial implementation can be to migrate CircuitExtraction and Skeletonization entities along these lines

james-isbister:
Thoughts and comments are useful @chr-pok @GianlucaFicarelli @eleftherioszisis @mgeplf @jdcourcol @MWolfR @AurelienJaquier 

We would like to propose this on Friday, if there is a clear path forward

jdcourcol:
- For sure `campaign` should have the task_type as they can exist standalone (no generation happened).
- Config as well as it can be standalone.
- on both activities, it seems useless, as the input is enough to tell us.

jdcourcol:
[...]

PR #540, opened 2026-02-25
New ion channel model simulation campaign entity

AurelienJaquier:
Ion channel model simulations will use (almost) the same workflow as Circuit and ME-Model simulations. So it will use the same Simulation and SimulationResults entities as well as SimulationExecution and SimulationGeneration activities.
But since it can have multiple ion channel models as input, opposed to the single circuit used as input in circuit simulations, I have to create this new IonChannelModelSimulationCampaign to refflect that.
I also make entity_id in Simulation optional, since we do not need it for ion channel model simulations (we already know the ion channel models used thanks to IonChannelModelSimulationCampaign).

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -1292,7 +1292,7 @@ class SimulationCampaignBase(
             server_default="{}",
         )
 
-    __mapper_args__ = {
+    __mapper_args__ = {  # noqa: RUF012
         "polymorphic_on": type,
     }
 


AurelienJaquier:
The issue is that I need the `simulation_campaign_id` in `Simulation` to be compatible with both `SimulationCampaign` and `IonChannelModelSimulationCampaign`. I could not find a way to achieve this without refactoring `SimulationCampaign` to be a child of `SimulationCampaignBase`.
Is there a way to do it without refactoring?

GianlucaFicarelli:
> The issue is that I need the `simulation_campaign_id` in `Simulation` to be compatible with both `SimulationCampaign` and `IonChannelModelSimulationCampaign`. I could not find a way to achieve this without refactoring `SimulationCampaign` to be a child of `SimulationCampaignBase`. Is there a way to do it without refactoring?

Instead of defining IonChannelModelSimulationCampaign and subclass SimulationCampaignBase, would the generic campaign approach work for that use case, as defined in https://github.com/openbraininstitute/entitycore/issues/523 ?
It can be simpler than subclassing and I can proceed on on that PR.

PR #542, opened 2026-03-01
Add generic task_config and task_activity

GianlucaFicarelli:
This PR adds more generic models compared to https://github.com/openbraininstitute/entitycore/pull/532.
Only TaskConfig and TaskActivity are defined, and can replace the previous models according to this mapping:

- Campaign -> TaskConfig[xxx__campaign]
- TaskConfig -> TaskConfig[xxx__config]
- TaskConfigGeneration -> TaskActivity[xxx__config_generation]
- TaskExecution -> TaskActivity[xxx__execution]

where:

task_config_type (enum):
- circuit_simulation__campaign
- circuit_simulation__config
...

task_activity_type (enum):
- circuit_simulation__config_generation
- circuit_simulation__execution
...

This would allow to define workflows of task_config with more than 2 activities.
This generalization is possible only if:
    - all the task_configs (campaign and config) share the same attributes (scan_parameters, parent_id, input)
    - all the task_activities (config_generation and execution) share the same attributes (executor, execution_id)

### DIFF
--- a/app/db/types.py
+++ b/app/db/types.py
@@ -440,6 +440,7 @@ class AssetLabel(StrEnum):
     ion_channel_model_thumbnail = auto()
     circuit_extraction_config = auto()
     skeletonization_config = auto()
+    task_config = auto()
 
 
 class LabelRequirements(BaseModel):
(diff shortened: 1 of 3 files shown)


chr-pok:
> This PR adds more generic models compared to https://github.com/openbraininstitute/entitycore/pull/532.
> ...

Thanks for this PR! Since campaigns and individual configs are conceptually different, I am not quite sure if this level of generalization is really needed. I personally would prefer to keep them (as well as the corresponding activities) separate, as in the alternative [PR#532](https://github.com/openbraininstitute/entitycore/pull/532). This would allow us to be more flexible w.r.t. specific attributes required for campaigns vs. single configs (which will probably be displayed/filtered differently in the UI, etc.). Or is there a compelling reason why these even more generic models would be beneficial?

PR #551, opened 2026-03-09
allow ion channel model in filters of simulation campaign

AurelienJaquier:
to allow for ion channel simulations: they will also use simulation campaigns, but the linked entity_id will be the one of the first ion channel model simulated.


bilalesi (commented):
I think this should be enough, but I’ll let @eleftherioszisis confirm

PR #563, opened 2026-03-23
additional sorting fields

jdcourcol:
Added ordering fields for circuit:
- experiment_date
- published_in

BoutonDensity:
- mean, sample size, number of measurement

MEModel:
- MType, EType


app/filters/density.py
```diff
@@ -71,6 +95,10 @@ class Constants(CustomFilter.Constants):
             "subject__species__name",
             "subject__age_value",
             "mtype__pref_label",
+            "subject__strain__name",
+            "measurement_mean__value",
+            "measurement_sem__value",
```
GianlucaFicarelli: I find clearer to use "standard_error" rather than "sem", especially because it matches the same name used for filtering/ordering, that is `MeasurementStatistic.standard_error`.
Any reason to prefer sem?

mgeplf (approved):

PR #564, opened 2026-03-23
Update update_date and updated_by_id in update endpoints

GianlucaFicarelli:
The attributes updated_by_id and update_date are defined in the base models inheriting from Identifiable (abstract).
Because of Joined Table Inheritance, any change in a joined table (e.g. cell_morphology.name) should cause an update in the base table (e.g. entity.update_date and entity.created_by_id)

This is the summary of the main changes:

updated the common functions 
- router_update_one (for entities and global resources)
- router_update_activity_one (for activities)

updated update_one in each service:
- add check_authorized_project=True (for models with authorized_project_id)
- or check_authorized_project=False (for admin-only models without it)

update admin_update_one in each service:
- always check_authorized_project=False
- add user_context: AdminContextDep parameter
- and pass user_context=user_context instead of user_context=None

Fix https://github.com/openbraininstitute/prod-platform-architecture/issues/186

### DIFF
--- a/app/service/brain_region.py
+++ b/app/service/brain_region.py
@@ -103,26 +103,12 @@ def update_one(
         user_context=user_context,
         json_model=json_model,
         response_schema_class=BrainRegionRead,
-        check_authorized_project=False,
-    )
-
-
-def admin_update_one(
-    user_context: AdminContextDep,
-    db: SessionDep,
-    id_: uuid.UUID,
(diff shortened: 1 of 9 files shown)


mgeplf (approved):
nice; tedious to add `check_authorized_project` everywere

eleftherioszisis (approved):

PR #568, opened 2026-04-07
Simplify query parameters for within_brain_region_* call

mgeplf:
* as agreed in: https://github.com/openbraininstitute/entitycore/issues/567
* `within_brain_region_hierarchy_id` isn't required any more; if it's included it is checked to make sure it matches `within_brain_region_brain_region_id`
* remove deprecated `within_brain_region_ascendants`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,12 +55,12 @@ build:  ## Build the Docker image
 publish: build  ## Publish the Docker image to DockerHub
 	docker compose push app
 
-TESTS ?= tests
+# use `PYTEST_ADDOPTS=` to narrow the pytests run, and add options 
 test-local:  ## Run tests locally
 	@$(call load_env,test-local)
 	docker compose up --wait db-test
 	uv run -m alembic upgrade head
-	uv run -m pytest $(TESTS)
+	uv run -m pytest
 	uv run -m coverage xml
 	uv run -m coverage html
 
--- a/Makefile
+++ b/Makefile
@@ -55,11 +55,12 @@ build:  ## Build the Docker image
 publish: build  ## Publish the Docker image to DockerHub
 	docker compose push app
 
+TESTS ?= tests
 test-local:  ## Run tests locally
 	@$(call load_env,test-local)
 	docker compose up --wait db-test
 	uv run -m alembic upgrade head
-	uv run -m pytest
+	uv run -m pytest $(TESTS)
 	uv run -m coverage xml
 	uv run -m coverage html
 
--- a/app/dependencies/common.py
+++ b/app/dependencies/common.py
@@ -5,14 +5,15 @@ from typing import Annotated, NotRequired, TypedDict
 import sqlalchemy as sa
 from fastapi import Depends, Query
 from fastapi.dependencies.models import Dependant
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 from sqlalchemy.orm import DeclarativeBase, InstrumentedAttribute, Session
 from starlette.requests import Request
 
+from app.db.model import BrainRegion
 from app.db.types import DerivationType
 from app.errors import ApiError, ApiErrorCode
 from app.filters.base import CustomFilter
-from app.filters.brain_region import WithinBrainRegionDirection, filter_by_hierarchy_and_region
+from app.filters.brain_region import WithinBrainRegionDirection, filter_by_region
 from app.queries.filter import filter_from_db
 from app.queries.types import ApplyOperations
 from app.schemas.types import Facet, Facets, PaginationRequest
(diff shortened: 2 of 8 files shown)


app/dependencies/common.py
```diff
@@ -149,41 +150,111 @@ def __call__(self, q: sa.Select[tuple[T]], vector_col: InstrumentedAttribute):
...
     """Handle parameters for within_brain_region_* query params.
 
-    Using `within_brain_region_ascendants` will be deprecated; future usage
-    should only be through `within_brain_region_direction`.
-
-    During the transition, `within_brain_region_direction` will take precedence if it's there
-
+    Only `within_brain_region_direction` and `within_brain_region_brain_region_id`
+    are required. `within_brain_region_hierarchy_id` is optional, and only
+    used to check that the chosen brain_region_id matches the correct hierarchy.
```
GianlucaFicarelli: I don't see any added value in `within_brain_region_hierarchy_id` if `within_brain_region_brain_region_id` is unique and enough to identify the region and the hierarchy. Cannot we just remove it?

Makefile
```diff
@@ -55,11 +55,12 @@ build:  ## Build the Docker image
 publish: build  ## Publish the Docker image to DockerHub
 	docker compose push app
 
+TESTS ?= tests
 test-local:  ## Run tests locally
 	@$(call load_env,test-local)
 	docker compose up --wait db-test
 	uv run -m alembic upgrade head
-	uv run -m pytest
+	uv run -m pytest $(TESTS)
```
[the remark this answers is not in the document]
mgeplf: done

app/service/entity.py
```diff
@@ -65,20 +65,18 @@ def count_entities_by_type(
...
-        and in_brain_region.within_brain_region_hierarchy_id is not None
-        and in_brain_region.within_brain_region_brain_region_id is not None
+        in_brain_region and in_brain_region.within_brain_region_brain_region_id is not None
     )
 
     filter_conditions = []
     if should_filter_brain_region:
+        # should be checked by model_validator
+        assert in_brain_region.within_brain_region_brain_region_id  # noqa: S101
+        assert in_brain_region.within_brain_region_direction  # noqa: S101
```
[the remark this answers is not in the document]
mgeplf: done

app/dependencies/common.py
mgeplf: yup, certainly; JDC mentioned having `pass both but we raise an error if the provided hierarchy does not match the one for the region`

It would mean I don't need a db handle, which would be nice, since it's annoying to have to do a db round-trip to check.

OTOH, it would require a change in the frontend to drop the `within_brain_region_hierarchy_id`

GianlucaFicarelli: Yeah that's right, removing `within_brain_region_hierarchy_id` is a breaking change.
We can decide to keep a dummy parameter (that is just ignored) until the transition is complete, or we need to ensure that all the clients are updated at the same time, that can be more complex.

mgeplf: Sure, works for me.  I'll confirm w/ JDC.

app/dependencies/common.py
```diff
@@ -149,41 +149,68 @@ def __call__(self, q: sa.Select[tuple[T]], vector_col: InstrumentedAttribute):
...
     """Handle parameters for within_brain_region_* query params.
 
-    Using `within_brain_region_ascendants` will be deprecated; future usage
-    should only be through `within_brain_region_direction`.
-
-    During the transition, `within_brain_region_direction` will take precedence if it's there
-
+    Only `within_brain_region_direction` and `within_brain_region_brain_region_id` are required.
+    `within_brain_region_hierarchy_id` exists for historical reasons, and is
+    not used or verified: see https://github.com/openbraininstitute/entitycore/issues/567
```
GianlucaFicarelli, line 154: Minor, we could explicitly mention that it should be considered deprecated and it will be removed in a future release. We should then coordinate with the frontend and ML before removing the parameter.
It can be a good idea to open a ticket for that.

docs/API.md
```diff
@@ -279,12 +278,10 @@ For example, in the following, if `RegionA-1` is the target region, all the outl
...
 
+Only the `within_brain_region_brain_region_id` and `within_brain_region_direction` query parameters are required.
 
-Originally, there was a boolean: `within_brain_region_ascendants: bool = False`.
-If `within_brain_region_direction` is included, it overrides `within_brain_region_ascendants`.
-Otherwise, `within_brain_region_direction` is used, with the behavior that `within_brain_region_ascendants=True` is equivalent to `within_brain_region_direction=ascendants`, and `within_brain_region_ascendants=True`is equivalent to `within_brain_region_direction=descendants`.
-This will be deprecated.
-
+Note: If `within_brain_region_hierarchy_id` is included, it is not an error, but it is not checked to be correct against `within_brain_region_brain_region_id`.
+See: https://github.com/openbraininstitute/entitycore/issues/567
```
GianlucaFicarelli, line 284: Similarly, I would say that within_brain_region_hierarchy_id should be considered deprecated and it will be removed.

GianlucaFicarelli (approved):
LGTM, just a comment about removing within_brain_region_hierarchy_id in the future

app/dependencies/common.py
mgeplf: done; where is the best place to open a ticket for that?

docs/API.md
mgeplf: done

PR #572, opened 2026-04-13
Admin read_many endpoints

eleftherioszisis:
* Added explicit GlobalRoute, EntityRoute, AssociationRoute, AgentRoute in app/routers/types.py and refactored routers to use them.
* Created a default APIRouter in app/routers/common.py with default CRUD routes and refactored routers to use it wherever possible.
* On the service side a check_authorized_project allows toggling between filtering and not filtering by project_id
* Add missing create_one, update_one, delet_one endpoints for CellComposition
* Add missing update_one, delete_one endpoints for ExternalUrl
* Add mising update_one endpoint to Person, Organization, and Consortium


app/routers/common.py
```diff
@@ -0,0 +1,39 @@
...
+    """Attach admin-specific routes for a resource to the admin router."""
+    router.get(f"/{route}")(service.admin_read_many)
+    router.get(f"/{route}/{{id_}}")(service.admin_read_one)
+    router.patch(f"/{route}/{{id_}}")(service.admin_update_one)
+
+
+def create_user_router(
+    route: str,
+    service: UserCrudService,
+    before_routes: list[Callable[[APIRouter], Any]] | None = None,
```
GianlucaFicarelli, line 43: Minor and not strictly needed, but for completeness there could be also after_routes and it can be used for passing additional routes, as the ones in brain_atlas and brain_region_hierarchy.

app/routers/contribution.py
```diff
@@ -1,17 +1,19 @@
...
-read_many = router.get("")(app.service.contribution.read_many)
-read_one = router.get("/{id_}")(app.service.contribution.read_one)
-create_one = router.post("")(app.service.contribution.create_one)
+read_many = router.get("")(service.read_many)
+read_one = router.get("/{id_}")(service.read_one)
+create_one = router.post("")(service.create_one)
 
-admin_read_one = admin_router.get(f"/{ROUTE}/{{id_}}")(app.service.contribution.admin_read_one)
+admin_read_one = admin_router.get(f"/{ROUTE}/{{id_}}")(service.admin_read_one)
+admin_read_many = admin_router.get(f"/{ROUTE}")(service.admin_read_many)
```
GianlucaFicarelli: It seems it's not possible to:
- user update, admin update: probably ok, if it doesn't make sense to update any field
- user delete: is there a reason to not allow it? Depending on the case we can add a comment with the reason, or a TODO to be done in a different PR, or add it although not strictly in scope with this PR.

A comment explaining the exceptions can be useful here and also in other endpoints that don't call create_user_router and register_default_admin_routes

app/routers/common.py
```diff
@@ -0,0 +1,39 @@
...
+    router.post("")(service.create_one)
+    router.patch("/{id_}")(service.update_one)
+    router.delete("/{id_}")(service.delete_one)
+
+
+def register_default_admin_routes(router: APIRouter, service: AdminCrudService, route: str) -> None:
+    """Attach admin-specific routes for a resource to the admin router."""
+    router.get(f"/{route}")(service.admin_read_many)
+    router.get(f"/{route}/{{id_}}")(service.admin_read_one)
+    router.patch(f"/{route}/{{id_}}")(service.admin_update_one)
```
GianlucaFicarelli, line 36: With this common function, wouldn't be clearer if we move here the admin delete endpoint from admin.py?

app/routers/mtype_classification.py
```diff
@@ -1,7 +1,7 @@
 from fastapi import APIRouter
 
-import app.service.mtype_classification
 from app.routers.admin import router as admin_router
+from app.service import mtype_classification as service
 
 ROUTE = "mtype-classification"
```
GianlucaFicarelli: It should be `AssociationRoute.mtype_classification`

app/routers/common.py
eleftherioszisis: Added

app/schemas/cell_composition.py
```diff
@@ -1,21 +1,50 @@
...
     IdentifiableMixin,
     EntityTypeMixin,
+    BrainRegionReadMixin,
+    CreatedByUpdatedByMixin,
     ContributionReadWithoutEntityMixin,
     AssetsMixin,
-    NameDescriptionMixin,
 ):
-    pass
+    species: NestedSpeciesRead
```
GianlucaFicarelli: Good catch, so species was missing although defined in the db model?
Shouldn't we add also
```
strain: NestedStrainRead | None = None
```

We can also consider adding a mixin with species and strain, since they are used in multiple places.

app/schemas/service.py
```diff
@@ -0,0 +1,16 @@
...
+    read_one: Callable[..., Any]
+    create_one: Callable[..., Any]
+    update_one: Callable[..., Any]
+    delete_one: Callable[..., Any]
+
+
+class AdminCrudService(Protocol):
+    admin_read_many: Callable[..., Any]
+    admin_read_one: Callable[..., Any]
+    admin_update_one: Callable[..., Any]
```
GianlucaFicarelli, line 16: Related to the previous comment, `admin_delete_one` should be added here as well if it's added to register_default_admin_routes

app/service/scientific_artifact_external_url_link.py
```diff
@@ -133,18 +135,28 @@ def read_many(
...
             ScientificArtifactExternalUrlLink.scientific_artifact_id
             == scientific_artifact_alias.id,
         ).join(
             external_url_alias,
             ScientificArtifactExternalUrlLink.external_url_id == external_url_alias.id,
-        ),
-        project_id=user_context.project_id,
-        db_model_class=scientific_artifact_alias,
-    )
+        )
```
GianlucaFicarelli: Since the internal join is the same, couldn't this be rewritten to reuse the same statement with something like:
```python
    base_join_query = lambda q: q.join(
        scientific_artifact_alias,
        ScientificArtifactExternalUrlLink.scientific_artifact_id == scientific_artifact_alias.id,
    ).join(
        external_url_alias,
        ScientificArtifactExternalUrlLink.external_url_id == external_url_alias.id,
    )

    if check_authorized_project:
        filter_query = lambda q: constrain_to_readable_entities(
            base_join_query(q),
            project_id=user_context.project_id,
            db_model_class=scientific_artifact_alias,
        )
    else:
        filter_query = base_join_query
```

app/service/scientific_artifact_publication_link.py
```diff
@@ -133,18 +135,28 @@ def read_many(
...
             ScientificArtifactPublicationLink.scientific_artifact_id
             == scientific_artifact_alias.id,
         ).join(
             publication_alias,
             ScientificArtifactPublicationLink.publication_id == publication_alias.id,
-        ),
-        project_id=user_context.project_id,
-        db_model_class=scientific_artifact_alias,
-    )
+        )
```
GianlucaFicarelli: This can be rewritten similarly.

app/utils/routers.py
```diff
@@ -1,43 +1,16 @@
...
-        "ResourceRoute",
-        {item.name: _convert_resource_type_to_route(item.name) for item in ResourceType},
-    )
-else:
-    EntityRoute = StrEnum
-    GlobalRoute = StrEnum
-    ResourceRoute = StrEnum
-    ActivityRoute = StrEnum
+from app.db.types import EntityType, ResourceType
+from app.routers.types import EntityRoute, ResourceRoute
```
GianlucaFicarelli, line 4: There are functions that still import EntityRoute, ResourceRoute from this file, but it would be cleaner if they are imported directly from app.routers.types where they are now defined (example: admin.py)

GianlucaFicarelli (commented):
A few more comments, nice cleanup

app/service/derivation.py
```diff
@@ -81,6 +83,51 @@ def read_many(
...
+        derivation_type=derivation_type,
+        pagination_request=pagination_request,
+        entity_filter=entity_filter,
+        check_authorized_project=True,
+    )
+
+
+def admin_read_many(
+    *,
+    user_context: UserContextDep,
```
GianlucaFicarelli:
```suggestion
    user_context: AdminContextDep,
```

app/routers/contribution.py
eleftherioszisis: I added the rest of the endpoints for contribution. wdyt?

app/routers/common.py
eleftherioszisis: Added

app/routers/mtype_classification.py
eleftherioszisis: Fixed

app/schemas/cell_composition.py
eleftherioszisis: Indeed. Added it

app/service/scientific_artifact_external_url_link.py
eleftherioszisis: done

app/service/scientific_artifact_publication_link.py
eleftherioszisis: done

app/utils/routers.py
eleftherioszisis: fixed

app/service/derivation.py
eleftherioszisis: fixed

app/routers/contribution.py
GianlucaFicarelli: Thanks, it looks ok to me

GianlucaFicarelli (approved):

PR #578, opened 2026-04-15
Add sorting by memodel mtype and etype to single_neuron_synaptome

GianlucaFicarelli:
Examples:

```
http://127.0.0.1:8000/single-neuron-synaptome?order_by=me_model__etype__pref_label
http://127.0.0.1:8000/single-neuron-synaptome?me_model__etype__pref_label=AAA&order_by=me_model__etype__pref_label
```

### DIFF
--- a/tests/test_ordering.py
+++ b/tests/test_ordering.py
@@ -92,5 +92,8 @@ def test_sort_unsupported_nested_ordering_part():
         order_by=["subject__bad_part__name"],
         subject=NestedSubjectFilter(),
     )
-    with pytest.raises(ValueError, match="Unsupported ordering part 'bad_part'"):
+    with pytest.raises(
+        ValueError,
+        match="Unsupported ordering part 'bad_part' in 'subject__bad_part__name'",
+    ):
         invalid_filter.sort(sa.select(CellMorphology))
--- a/tests/test_ordering.py
+++ b/tests/test_ordering.py
@@ -1,4 +1,9 @@
+import pytest
+import sqlalchemy as sa
+
 from app.db.model import CellMorphology, License
+from app.filters.cell_morphology import CellMorphologyFilter
+from app.filters.subject import NestedSubjectFilter
 
 from tests.utils import PROJECT_ID, add_all_db, check_sort_by_field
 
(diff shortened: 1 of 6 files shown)


tests/test_ordering.py
```diff
@@ -77,3 +82,15 @@ def test_cell_morphology_ordering(db, client, subject_id, license_id, brain_regi
     data = response.json()["data"]
     assert len(data) == count // 2
     check_sort_by_field(data, "name", how="descending")
+
+
+def test_sort_unsupported_nested_ordering_part():
+    """Test that sorting by a nested field with an unsupported intermediate part raises an error."""
+
+    # Use model_construct to bypass the ordering_model_fields validator
```
mgeplf, line 90: huh, didn't know about that, neat

tests/test_single_neuron_synaptome.py
```diff
@@ -347,6 +348,22 @@ def test_facets(db, client, faceted_ids):
         "updated_by": [
             {"id": str(agent.id), "label": agent.pref_label, "count": 4, "type": agent.type}
         ],
+        "etype": [
+            {
+                "count": 4,
+                "id": ANY,
+                "label": "etype-emodel-0",
+                "type": "me_model.etype",
```
mgeplf, line 356: isn't this type a bit strange? Shouldn't it be just `etype`?

tests/test_ordering.py
```diff
@@ -77,3 +82,15 @@ def test_cell_morphology_ordering(db, client, subject_id, license_id, brain_regi
     data = response.json()["data"]
     assert len(data) == count // 2
     check_sort_by_field(data, "name", how="descending")
+
```
mgeplf, line 85: do you think it's worth adding an explicit test for this new feature?

tests/test_single_neuron_synaptome.py
GianlucaFicarelli: I think that `me_model.etype` is correct because it's a nested attribute.
In theory there could be multiple nested fields with the same name, but in different nested objects: for example, `me_model.emodel.etype`, although it's not exposed as a facet.
The key of the dict is just `etype` and the UI shows just `etype` as well, because it could be confusing otherwise.

GianlucaFicarelli: Note that it's the same for species and strain, see https://github.com/openbraininstitute/entitycore/blob/2737d9ba2dd9cb04bfef7d15216d38f036300ea1/tests/test_cell_morphology.py#L394-L411

tests/test_ordering.py
GianlucaFicarelli: The feature isn't really a new feature, because it was already possible to sort by nested fields like `subject__species__name`. Now the nested parts are just resolved differently (with the custom filter instead of the db model)
In tests/test_single_neuron_synaptome.py::test_sorting_filtering all the possible fields defined in the filter are tested already, do you think we should have a dedicated test here as well?

mgeplf: Ok, fair enough, that makes sense.

tests/test_single_neuron_synaptome.py
mgeplf: Ah, good point, I had briefly looked, but couldn't find an example.

mgeplf (approved):

eleftherioszisis (approved):

PR #586, opened 2026-04-21
Add protocol to emodel.exemplar_morphology and memodel.emodel.morphology

GianlucaFicarelli:
Other: Always use selectinload to fetch the protocols, since most morphologies should use the same

### DIFF
--- a/app/service/emodel.py
+++ b/app/service/emodel.py
@@ -47,21 +47,23 @@ def _load(select: sa.Select[tuple[EModel]]):
     return select.options(
         joinedload(EModel.species),
         joinedload(EModel.strain),
-        joinedload(EModel.exemplar_morphology),
+        joinedload(EModel.exemplar_morphology).selectinload(
+            CellMorphology.cell_morphology_protocol
+        ),
         joinedload(EModel.brain_region),
         selectinload(EModel.assets),
         selectinload(EModel.contributions).joinedload(Contribution.agent),
         selectinload(EModel.contributions).joinedload(Contribution.role),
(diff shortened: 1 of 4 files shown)


app/service/emodel.py
```diff
@@ -47,21 +47,23 @@ def _load(select: sa.Select[tuple[EModel]]):
...
-        selectinload(EModel.ion_channel_models).joinedload(IonChannelModel.brain_region),
-        selectinload(EModel.ion_channel_models).selectinload(IonChannelModel.assets),
+        selectinload(EModel.ion_channel_models).options(
+            joinedload(IonChannelModel.subject).options(
+                selectinload(Subject.species),
+                selectinload(Subject.strain),
+            ),
+            joinedload(IonChannelModel.brain_region),
+            selectinload(IonChannelModel.assets),
+        ),
```
GianlucaFicarelli, line 66: This is just to simplify the syntax, the result should be the same,

app/service/emodel.py
```diff
@@ -47,21 +47,23 @@ def _load(select: sa.Select[tuple[EModel]]):
     return select.options(
         joinedload(EModel.species),
         joinedload(EModel.strain),
-        joinedload(EModel.exemplar_morphology),
+        joinedload(EModel.exemplar_morphology).selectinload(
+            CellMorphology.cell_morphology_protocol
+        ),
```
GianlucaFicarelli, line 52: This is needed to fetch the protocol

eleftherioszisis (approved):

PR #587, opened 2026-04-21
Add Extracellular electrode array

eleftherioszisis:
Issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/153


app/db/types.py
```diff
@@ -1103,3 +1104,7 @@ class ExecutorType(StrEnum):
     single_node_job = auto()
     distributed_job = auto()
     jupyter_notebook = auto()
+
+
+class ElectrodeType(StrEnum):
+    placeholder = auto()
```
eleftherioszisis: @james-isbister please provide the supported list

james-isbister: As I just mentioned in the ticket, i think starting with a very small list would be good, and we can build it up, as we add new electrode configurations in obi-one. For now we could go with:
- Neuropixels 1.0
- Neuropixels 2.0
- Neuropixels Ultra
- Custom

james-isbister: Edited the list so there are 4

eleftherioszisis: would the following work for you?
```
neuropixels_v1
neuropixels_v2
neuropixels_ultra
custom
```

james-isbister:
I saw earlier there was a comment about the asset label "weights_file" but now I don't see it, or what it might have been changed to?

app/db/types.py
```diff
@@ -442,6 +443,7 @@ class AssetLabel(StrEnum):
     skeletonization_config = auto()
     task_config = auto()
     lod_mesh_block = auto()
+    weights = auto()
```
eleftherioszisis: @james-isbister here. I would prefer if this was less broad.

james-isbister: I don't have strong feelings about what to call it. Some suggestions would be:
- electrode_weights
- electrode_contribution_weights
- compartment_contribution_weights

Or some combination of those.

Essentially it is a file for N electrodes in the electrode array.
For each electrode there are M weights for summing the activity of M compartments in the circuit
So MxN values

eleftherioszisis: would `electrode_array_weight_matrix` work?

app/db/model.py
```diff
@@ -2335,4 +2336,18 @@ class TaskActivity(Activity, ExecutionActivityMixin):
...
+class SimulatableExtracellularRecordingArray(Entity, NameDescriptionVectorMixin):
+    __tablename__ = EntityType.simulatable_extracellular_recording_array
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+
+    electrode_type: Mapped[ElectrodeType]
+
+    circuit_id: Mapped[uuid.UUID] = mapped_column(ForeignKey(f"{EntityType.circuit}.id"))
+
+    circuit = relationship("Circuit", uselist=False, foreign_keys=circuit_id)
```
GianlucaFicarelli: Could you add the type annotation as `Mapped["Circuit"]`? With that, it should be enough to write the relationship as:

```suggestion
    circuit: Mapped["Circuit"] = relationship(foreign_keys=circuit_id)
```

app/db/types.py
```diff
@@ -800,6 +802,18 @@ class LabelRequirements(BaseModel):
...
+            LabelRequirements(
+                content_type=ContentType.h5,
+                is_directory=False,
+                description=(
+                    "File with matrix of N electrodes X M weights . "
+                    "Weights sum the activity of M compartments in the circuit"
+                ),
+            ),
+        ]
+    },
```
GianlucaFicarelli, line 816: When adding an asset label we should execute also `make update-asset-labels` to update asset-labels.md

app/db/types.py
```diff
@@ -800,6 +802,18 @@ class LabelRequirements(BaseModel):
...
         ],
     },
+    EntityType.simulatable_extracellular_recording_array: {
+        AssetLabel.electrode_array_weight_matrix: [
+            LabelRequirements(
+                content_type=ContentType.h5,
+                is_directory=False,
+                description=(
+                    "File with matrix of N electrodes X M weights . "
+                    "Weights sum the activity of M compartments in the circuit"
```
GianlucaFicarelli: There is a space before the dot, and a dot should be added at the end.
Some syntax is validated by generate_asset_labels_table.py

app/schemas/simulatable_extracellular_recording_array.py
```diff
@@ -0,0 +1,58 @@
...
+from app.schemas.contribution import ContributionReadWithoutEntityMixin
+from app.schemas.utils import make_update_schema
+
+
+class SimulatableExtracellularRecordingArrayBase(BaseModel, NameDescriptionMixin):
+    model_config = ConfigDict(from_attributes=True)
+
+    electrode_type: ElectrodeType
+
+    circuit_id: uuid.UUID
```
GianlucaFicarelli, line 24: For confirmation, is it ok to return just the circuit_id in the Read model?

eleftherioszisis: Switched to circuit_id only

app/db/types.py
```diff
@@ -1115,3 +1129,11 @@ class TargetSimulator(StrEnum):
...
     learning_engine = "LearningEngine"
     brian2 = "Brian2"
+
+
+class ElectrodeType(StrEnum):
+    neuropixels_v1 = auto()
+    neuropixels_v2 = auto()
+    neuropixels_ultra = auto()
+    custom = auto()
+    placeholder = auto()
```
GianlucaFicarelli: For discussion, in the main ticket it's [mentioned](https://github.com/openbraininstitute/prod-circuit-simulation/issues/153#issuecomment-4294628275) that we expect thousands of ElectrodeType, and we'd like the users to be able to create new ones.
If the number of types has this magnitude and it needs to be dynamic, it seems better to create an additional table for the type, so the users can create the new types, and specify it when creating a new extracellular electrode array.
With that approach, the electrode type could extend entity, similar to how we defined the cell morphology protocol and use it in cell morphology. We can also discuss it offline cc @james-isbister @eleftherioszisis

eleftherioszisis: oops, I forgot to remove the placeholder. Did it in the latest commit.

eleftherioszisis:
@james-isbister does this PR meet the ticket requirements?

james-isbister (approved):
Looks good to me, thanks!

PR #588, opened 2026-04-22
Add target_simulator in Circuit

eleftherioszisis:
Issue: https://github.com/openbraininstitute/prod-circuit-simulation/issues/148


app/db/types.py
```diff
@@ -1103,3 +1103,15 @@ class ExecutorType(StrEnum):
...
+class TargetSimulator(StrEnum):
+    """Target simulator values as used in libsonata.
+
+    See: https://github.com/openbraininstitute/libsonata/blob/5701031d206867e17d0296c5acbf206df40d0b3b/src/config.cpp#L129-L132
+    """
+
+    neuron = "NEURON"
+    coreneuron = "CORENEURON"
+    learning_engine = "LearningEngine"
+    brian2 = "brian2"
```
GianlucaFicarelli: I don't see brian2 in the linked sonata code 🤔

app/db/model.py
```diff
@@ -1821,6 +1822,12 @@ class Circuit(ScientificArtifact, NameDescriptionVectorMixin):
     number_synapses: Mapped[int] = mapped_column(BigInteger)
     number_connections: Mapped[int | None] = mapped_column(BigInteger)
 
+    target_simulator: Mapped[TargetSimulator] = mapped_column(
+        sa.Enum(TargetSimulator, values_callable=lambda x: [i.value for i in x]),
+        nullable=False,
+        server_default=TargetSimulator.coreneuron,
```
GianlucaFicarelli: Why is coreneuron the default? should we consider defining an undefined value instead?
It seems that libsonata defines also `invalid` but I didn't check how it's used.

app/schemas/circuit.py
```diff
@@ -25,6 +25,7 @@ class CircuitBase(BaseModel, NameDescriptionMixin):
 
     root_circuit_id: uuid.UUID | None = None
     atlas_id: uuid.UUID | None = None
+    target_simulator: TargetSimulator | None = None
```
GianlucaFicarelli: In the db model `target_simulator` is not nullable, so it should be defined not nullable here as well.

app/db/types.py
mgeplf: it's coming!
https://github.com/openbraininstitute/libsonata/pull/41

app/db/model.py
eleftherioszisis: Maybe making it nullable makes most sense. I would expect that not all circuits are simulatable. @mgeplf  thoughts?

mgeplf: I don't think we ever use the `invalid` key; we throw an error if an unknown simulator is used, and use `NEURON` as the default if `target_simulator` isn't specified - thus, the only way it would be detectably nullable is if one parses the .json oneself.

GianlucaFicarelli: In that case should we use the same `NEURON` default?
The default can be defined in the pydantic schema only, to avoid duplications.
In any case, it should be defined how to handle the migration for existing circuits: should we use the same value for all of them, or need to inspect the config to find the correct value?

eleftherioszisis: Switched to NEURON. I believe we should set a default and update later if necessary.

app/schemas/circuit.py
eleftherioszisis: Fixed. Set default value instead.

app/db/types.py
GianlucaFicarelli: Good 👍  In this case before merging this PR it's better to update the link after the libsonata PR is merged. Alternatively, we could link directly the sonata docs https://sonata-extension.readthedocs.io/en/latest/sonata_config.html#target-simulator

eleftherioszisis: Updated

GianlucaFicarelli (approved):

PR #595, opened 2026-04-29
Leverage pydantic unset functionality for update instead of custom NOT_SET

### LINKED ISSUE #497: Investigate replacing NOT_SET manual sentinel with pydantic's experimental MISSING
Check if we can delegate the manual not set processing functionality to pydantic using its MISSING sentinel instead of using our own custom one.

eleftherioszisis:
Leverage pydantic's `model_fields_set` that keeps track of the fields that have been set upon instantiation. That allows to differentiate from fields that are not updated by the user from fields that have been update with None.

This PR replaces the custom implementation with NOT_SET sentinels with pydantic's implicit functionality and allows removing NOT_SET from the schemas.

Fixes #497

### DIFF
--- a/app/schemas/cell_morphology_protocol.py
+++ b/app/schemas/cell_morphology_protocol.py
@@ -19,7 +19,7 @@ from app.schemas.base import (
     NameDescriptionMixin,
 )
 from app.schemas.types import SerializableHttpUrl
-from app.schemas.utils import NOT_SET, make_update_schema
+from app.schemas.utils import make_update_schema
 
 
 class CommonReadMixin(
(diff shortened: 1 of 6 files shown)


app/schemas/cell_morphology_protocol.py
```diff
@@ -325,5 +325,6 @@ def model_validate(cls, obj: Any, *args, **kwargs) -> CellMorphologyProtocolUser
         """Return the correct instance of the protocol."""
         model = cls._adapter.validate_python(obj, *args, **kwargs)
         # generation_type was needed only as a discriminator. It should not be updated.
-        model.generation_type = NOT_SET
+        model.generation_type = None
```
GianlucaFicarelli, line 328: Is `model.generation_type = None` needed even if calling `model.model_fields_set.discard()`?
What if both are removed, wouldn't `generation_type` be always the same as the original model, so not updated? 
In any case it doesn't hurt and it feels safer.

GianlucaFicarelli (approved):

PR #597, opened 2026-05-04
AIND ephys preprocessing & spikesorting TaskConfigTypes, TaskActivityTypes

james-isbister:
* Add TaskConfigType and TaskActivityType enum values

* Question: does anything need to be done in entitysdk after this?


app/db/types.py
```diff
@@ -160,6 +160,10 @@ class TaskConfigType(StrEnum):
     ion_channel_simulation__config = auto()
     em_synapse_mapping__campaign = auto()
     em_synapse_mapping__config = auto()
+    aind_ephys_preprocessing__campaign = auto()
+    aind_ephys_preprocessing__config = auto()
+    aind_ephys_spikesorting__campaign = auto()
+    aind_ephys_spikesorting__config = auto()
```
GianlucaFicarelli, line 166: Do you mind executing the db migration with a message like this?
```
make migration MESSAGE="Add AIND ephys types"
```

In entitysdk, instead, the values need to be updated with
```
tox -e generate-server-schemas --recreate
```
but it should be done after the deployment of entitycore to staging, or it can be done by pointing to a local instance of entitycore with:
```
DB_API_URL=http://127.0.0.1:8000/openapi.json tox -e generate-server-schemas --recreate
```

GianlucaFicarelli (approved):

Issue #582, opened 2026-04-16
Add admin endpoints for downloading/uploading files


eleftherioszisis:
## Description
Add endpoints for downloading and uploading files using the admin routes

## Acceptance criteria
- [ ] Download file as admin
- [ ] Upload file as admin

evakennyobi:
Can be closed @eleftherioszisis ? thanks

eleftherioszisis:
Not yet. I have implemented only the downloading part. Will do the uploading part too shortly.

eleftherioszisis:
Changes have been deployed to staging with entitycore 2026.5.6. entitysdk client changes available for entitysdk>=v0.17.0

PR #598, opened 2026-05-05
Add upload file endpoint for admins and make project_id optional for asset creation

eleftherioszisis:
Issue: #582

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -89,6 +89,8 @@ class Settings(BaseSettings):
 
     OPENAI_API_KEY: SecretStr | None = None
 
+    VIRTUAL_LAB_API_URL: str = "https://staging.cell-a.openbraininstute.org/api/virtual-lab-manager"
+
     @field_validator("DB_URI", mode="before")
     @classmethod
     def build_db_uri(cls, v: str, info: ValidationInfo) -> str:
--- a/app/routers/admin.py
+++ b/app/routers/admin.py
@@ -1,14 +1,16 @@
 import uuid
 from typing import Annotated
 
-from fastapi import APIRouter, Query
+from fastapi import APIRouter, Form, Query, UploadFile, status
 from starlette.responses import RedirectResponse
 
 from app.config import storages
-from app.db.types import StorageType
+from app.db.types import AssetLabel, StorageType
+from app.dependencies.auth import AdminContextDep
 from app.dependencies.common import PaginationQuery
 from app.dependencies.db import RepoGroupDep, SessionDep
 from app.dependencies.s3 import StorageClientFactoryDep
+from app.dependencies.virtual_lab_api import AdminVirtualLabClientDep
 from app.filters.asset import AssetFilterDep
 from app.routers.types import EntityRoute
 from app.schemas.asset import (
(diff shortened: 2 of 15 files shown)


app/service/asset.py
```diff
@@ -131,12 +136,12 @@ def get_writable_entity_db_asset(
     return asset
 
 
-def create_entity_asset(  # noqa: PLR0913
+def persist_entity_asset(  # noqa: PLR0913
```
GianlucaFicarelli: The name `persist_entity_asset` doesn't make very clear to me the difference with`create_entity_asset`.
What about something like `create_entity_asset_unverified`?

app/service/asset.py
```diff
@@ -131,12 +136,12 @@ def get_writable_entity_db_asset(
...
     repos: RepositoryGroup,
     *,
+    entity: Entity,
     user_context: UserContextWithProjectId,
     entity_type: EntityType,
```
GianlucaFicarelli: Also `entity_type` seems redundant if entity is passed.

app/service/asset.py
```diff
@@ -148,13 +153,8 @@ def create_entity_asset(  # noqa: PLR0913
...
-        repos,
-        user_context=user_context,
-        entity_type=entity_type,
-        entity_id=entity_id,
-    )
+    """Insert an asset row; the parent entity must already be loaded."""
+    entity_id = entity.id
     full_path = full_path or build_s3_path(
         vlab_id=user_context.virtual_lab_id,
         proj_id=user_context.project_id,
```
GianlucaFicarelli: proj_id should be the one of the entity, and vlab_id should be set accordingly.
It can be error prone to set them from `user_context.project_id` and `user_context.virtual_lab_id` because they aren't validated to be consistent with the entity when the admin endpoint is called.
Although it would add a dependency with virtual-lab-api, it would be possible to retrieve the correct virtual lab from the endpoint `/api/virtual-lab-manager/virtual-labs/projects/{project_id}/virtual-lab` that @bilalesi added in https://github.com/openbraininstitute/virtual-lab-api/pull/149 and that is accessible only to admins.

In the medium term we could still think to revisit the template used for storing assets, by keeping only the project_id and removing the virtual_lab_id, so we won't need the virtual_lab_id anymore in entitycore.

app/service/asset.py
```diff
@@ -196,6 +196,128 @@ def create_entity_asset(  # noqa: PLR0913
...
+            f"Exception: {e}"
+        )
+        raise ApiError(
+            message=msg,
+            error_code=ApiErrorCode.ASSET_INVALID_CONTENT_TYPE,
+            http_status_code=HTTPStatus.UNPROCESSABLE_ENTITY,
+        ) from None
+
+
+def _complete_small_upload_to_s3(
```
GianlucaFicarelli: I don't see a great improvement in moving this code to `_complete_small_upload_to_s3`, instead of leaving it in `upload_entity_asset`.

app/service/asset.py
```diff
@@ -196,6 +196,128 @@ def create_entity_asset(  # noqa: PLR0913
...
+    *, s3_client: S3Client, storage: StorageUnion, file: UploadFile, asset_read: AssetRead
+) -> AssetRead:
+    if not upload_to_s3(
+        s3_client,
+        file_obj=file.file,
+        bucket_name=storage.bucket,
+        s3_key=asset_read.full_path,
+    ):
+        raise HTTPException(status_code=500, detail="Failed to upload object")
```
GianlucaFicarelli: If simpler, we could make `upload_to_s3` raise an exception instead of returning False, although HTTPException doesn't really belong to the layer of utils/s3.py.

app/service/asset.py
eleftherioszisis: Removed

app/service/asset.py
eleftherioszisis: Renamed

app/service/asset.py
eleftherioszisis: Consolidated

app/routers/admin.py
```diff
@@ -124,6 +125,29 @@ def download_entity_asset(
     )
 
 
+@router.post("/{entity_route}/{entity_id}/assets", status_code=status.HTTP_201_CREATED)
+def upload_entity_asset(
+    *,
+    repos: RepoGroupDep,
+    entity_route: EntityRoute,  # noqa: ARG001
```
GianlucaFicarelli: It doesn't seem a good idea to ignore the specified entity_route, or it would be possible to call `/{any_entity_route}/{unrelated_id}/assets` and the asset would be registered anyway.
It could be used to verify that the type of the retrieved entity is consistent with the route.

app/service/asset.py
eleftherioszisis: I added vlab manager client to get the vlab id for admins.

app/config.py
```diff
@@ -89,6 +89,8 @@ class Settings(BaseSettings):
 
     OPENAI_API_KEY: SecretStr | None = None
 
+    VIRTUAL_LAB_API_URL: str = "https://staging.cell-a.openbraininstute.org/api/virtual-lab-manager"
```
GianlucaFicarelli, line 92: Just a reminder that VIRTUAL_LAB_API_URL needs to be added to terraform

GianlucaFicarelli (approved):
LGTM

PR #600, opened 2026-05-06
Add create_extracellular_array Activty and Config types

james-isbister:


alembic/versions/20260506_095926_129ca3e7dc4e_add_create_extracellular_array_activity_.py
```diff
@@ -0,0 +1,148 @@
...
+            "ion_channel_simulation__campaign",
+            "ion_channel_simulation__config",
+            "em_synapse_mapping__campaign",
+            "em_synapse_mapping__config",
+            "aind_ephys_preprocessing__campaign",
+            "aind_ephys_preprocessing__config",
+            "aind_ephys_spikesorting__campaign",
+            "aind_ephys_spikesorting__config",
+            "create_silico_extracellular_array__campaign",
+            "create_silico_extracellular_array__config",
```
eleftherioszisis: Verbs are not consistent with the naming schema. Would `in_silico_extracellular_array_creation__*` work? However, without knowing the specifics "silico_extracellular_array" seems too broad to me.

james-isbister: Fair point on the consistency. It's for creating the extracellular array entities we made the other week

james-isbister:
Adjusted the names to match the name of the [entity](https://github.com/openbraininstitute/entitycore/blob/a29707f0d95b5bac3feaab1df5bd9aa142d23572/app/db/model.py#L2347) which the task creates

Also re-run the migration

@eleftherioszisis

alembic/versions/20260506_095926_129ca3e7dc4e_add_create_extracellular_array_activity_.py
eleftherioszisis: Yeah, but what does the task do? It does create that entity but what is the name of the process that takes place? IMHO the name should name the process the result of which is the simulatable_extracellular_recording_array. Does "simulatable_extracellular_recording" as a process make sense in this context? Or maybe something that includes "weight_calculation"? Not sure what's best.

james-isbister: Ah damn, I meant for it to be simulatable_extracellular_recording_array_**creation**__*

Would that work? Or "extracellular_recording_array_weights_calculation__*", perhaps

eleftherioszisis: I believe the second suggestion is more descriptive if that's ok with you.

james-isbister: That was too long to be allowed as a string enum :D

So I made it "extracellular_recording_weights_calculation__*"

james-isbister:
Can we merge this now @eleftherioszisis ?

eleftherioszisis (approved):

PR #602, opened 2026-05-08
Updating the `Derivation` entity for circuit manipulations

darshanmandge:
- Add `label` field to `Derivation`.
The optional `label` field for the SONATA nodes file `model_template` entry, by convention, is formatted as `hoc:<template_name>` (e.g., `hoc:cADpyr_L5TPC`, `hoc:bAC_L23BC`).
- Add `emodel_circuit` derivation type to link EModels to Circuits
A new `DerivationType.emodel_circuit` value that models the relationship where an EModel (`used` entity) is assigned to neurons in a Circuit (`generated` entity).
- This is required to link e-models to circuits and circuit manipulations.

cc: @chr-pok Here is the initial PR. You can add the remaining `DerivationType`s for circuit manipulation and any other modifications needed.


app/db/model.py
```diff
@@ -1671,6 +1671,7 @@ class Derivation(Base):
     used: Mapped["Entity"] = relationship(foreign_keys=[used_id])
     generated: Mapped["Entity"] = relationship(foreign_keys=[generated_id], passive_deletes=True)
     derivation_type: Mapped[DerivationType]
+    label: Mapped[str | None] = mapped_column(nullable=True)
```
GianlucaFicarelli: Sqlalchemy can use the type annotation, it shouldn't be needed to add `mapped_column(nullable=True)`

```suggestion
    label: Mapped[str | None]
```

tests/test_derivation.py
```diff
@@ -158,6 +160,51 @@ def test_create_one(client, derivation_type, root_circuit, circuit):
...
         "derivation_type": derivation_type,
+        "label": None,
+    }
+
+
+def test_create_emodel_circuit_with_label(client, root_circuit, circuit):
+    """Link an emodel (used) to a circuit (generated) with the SONATA model_template label.
+
+    The ``used_id`` should be an EModel id in real usage; here we reuse a circuit fixture
+    since the generic /derivation endpoint does not constrain the entity types.
```
GianlucaFicarelli: It's better to use an emodel instead of a circuit, or this test will fail as soon as we add the validation on the type.

app/db/types.py
```diff
@@ -248,15 +248,25 @@ class DerivationType(StrEnum):
     """Represents the type of derivation relationship between two entities.
 
     Attributes:
+        circuit_customization: Indicates that a circuit entity was derived from another circuit by
+          customizing certain components. The optional ``label`` field specifies the type of
+          customization, such as ``synaptic_modification``, ``new_emodels``,
+          ``emodel_modification``, ``population_modification``.
```
GianlucaFicarelli, line 254: The labels for circuit_customization are correct according to the definition [in this comment](https://github.com/openbraininstitute/prod-build-circuit/issues/30#issuecomment-4397262402).
However, `new_emodels` doesn't follow the same pattern of the others.
Something like `emodel_addition` or `emodel_creation` sounds more consistent with the other names.
@chr-pok what do you think?

GianlucaFicarelli (commented):
The changes look overall good. We'll need to add validations:
- validate that the label is valid, depending on the derivation_type (ideally in this PR)
- validate that the types for used and generated are valid (separate PR)

app/db/types.py
chr-pok: Good idea, I have changes it to "emodel_addition" for consistency with the others.

tests/test_derivation.py
darshanmandge: Done in the latest commit.

> validate that the label is valid, depending on the derivation_type (ideally in this PR)

This was also added.

darshanmandge:
> validate that the types for used and generated are valid (separate PR)

@GianlucaFicarelli Should validation happen at the level of:
- Pydantic schema (DerivationCreate):  but it only has IDs, so we'd need a DB lookup, which doesn't fit cleanly in a validator.
- Service layer (app/service/derivation.py::create_one), fetches both entities anyway, and can check their types there.

GianlucaFicarelli:
> > validate that the types for used and generated are valid (separate PR)
> 
> @GianlucaFicarelli Should validation happen at the level of:
> 
> * Pydantic schema (DerivationCreate):  but it only has IDs, so we'd need a DB lookup, which doesn't fit cleanly in a validator.
> * Service layer (app/service/derivation.py::create_one), fetches both entities anyway, and can check their types there.

The service layer seems preferable to me in that case because of the db lookups.
We don't even need to load the full specific subclass, but checking entity.type should be enough, unless it makes the code more complex.
It should be done only when the derivation is created, since there isn't an update endpoint.
Are you willing to open a PR for that? If not, we can see later.

darshanmandge:
>  Are you willing to open a PR for that? If not, we can see later.

I can create a PR to get started. We can update and merge it later. Anyway, it is not essential for the derivation entity to work.

GianlucaFicarelli (approved):
LGTM

Issue #571, opened 2026-04-10
Failure when applying filters to the brain-region endpoint


GianlucaFicarelli:
## Environment
Logged in staging, verified locally in entitycore 2026.4.0.

## Description
The brain-region endpoint fails with some filters:

1. `http://127.0.0.1:8000/brain-region?semantic_search=thalamus&page=1&page_size=10&id=e3e70682-c209-4cac-a29f-6fbed82c07cd&hierarchy_id=e3e70682-c209-4cac-a29f-6fbed82c07cd&order_by=name&species__id=e3e70682-c209-4cac-a29f-6fbed82c07cd&strain__id=e3e70682-c209-4cac-a29f-6fbed82c07cd`

-> `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateAlias) table name "brain_region_hierarchy" specified more than once`

2. `http://127.0.0.1:8000/brain-region?strain__id=e3e70682-c209-4cac-a29f-6fbed82c07cd`

-> `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) missing FROM-clause entry for table "species"`

## To Reproduce

Call the urls above.

## Expected behavior
Return the json payload with the list of results.

## Acceptance criteria
- [ ] The endpoint 1 above succeeds
- [ ] The endpoint 2 above succeeds

GianlucaFicarelli:
The fix will be released with entitycore 2026.5.2

evakennyobi:
Can be closed based on the above?

GianlucaFicarelli:
The fix in staging works, but entitycore 2026.5.2 has been released on 13/5 and it's not deployed to production yet. We can verify and close the ticket after the next deployment to production.

evakennyobi:
I think can be closed since the latest push

PR #603, opened 2026-05-08
Fix brain-region filter, add additional filters to brain-atlas and brain-atlas-region

GianlucaFicarelli:
- Fix brain region filter in https://github.com/openbraininstitute/entitycore/issues/571 for species/strain (DuplicateAlias and missing FROM-clause)
- Fix brain-atlas filter for species/strain (cartesian product)
- Add filters to:
  - BrainAtlas:
    - hierarchy_id
    - species (fixed)
    - strain (fixed)
  - BrainAtlasRegion:
    - is_leaf_region
    - brain_atlas_id
    - brain_region_id

Note: an alternative to adding the *id filters is to add full nested filters, so that the filters would have the double underscore separator and become:
- brain_atlas_id -> brain_atlas__id, brain_atlas__name, ...
- brain_region_id -> brain_region__id, brain_region__name, ...

However, they would require additional joins and it seems unnecessary for the moment, also considering that the returned payload doesn't contain those nested entities but just the ids.

Filtering by `brain_atlas_id` would have been useful in https://github.com/openbraininstitute/migration-scripts/pull/19

### DIFF
--- a/app/service/brain_region.py
+++ b/app/service/brain_region.py
@@ -1,7 +1,8 @@
 import uuid
+from typing import TYPE_CHECKING
 
 import sqlalchemy as sa
-from sqlalchemy.orm import joinedload, raiseload
+from sqlalchemy.orm import aliased, joinedload, raiseload
 
 import app.queries.common
 from app.db.model import BrainRegion, BrainRegionHierarchy, Species, Strain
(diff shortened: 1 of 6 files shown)


eleftherioszisis (approved):
lgtm

app/service/brain_region.py
```diff
@@ -32,13 +36,21 @@ def read_many(
     semantic_search: str | None = None,
 ) -> ListResponse[BrainRegionRead]:
     db_model_class = BrainRegion
+    brh_species_alias = aliased(BrainRegionHierarchy, flat=True, name="brh_species")
```
mgeplf, line 39: are the names needed?

mgeplf (approved):

app/service/brain_region.py
GianlucaFicarelli: The names aren't required, but when specified sqlalchemy will use those in the query instead of brain_region_hierarchy_1 and brain_region_hierarchy_2, so it's easier to inspect the query.

PR #608, opened 2026-05-18
Add multipart upload for directories

### LINKED ISSUE #589: Support for large file upload in directory assets
## Description
Currently, large file upload in directory assets does not support multi-part transfer and fails for files larger than a few GB. Since this functionality is required when uploading new circuits (e.g., human circuit in [prod-human-data-and-models#29](https://github.com/openbraininstitute/prod-human-data-and-models/issues/29)) or when extracting large subcircuits (e.g., long slice from `nbS1` in [prod-build-circuit#28](https://github.com/openbraininstitute/prod-build-circuit/issues/28)), the upload functionality for directory assets should be extended to support large files.

Here an example from [prod-build-circuit#28](https://github.com/openbraininstitute/prod-build-circuit/issues/28) (Job ID: `019db5bf-d6ce-74b4-9950-75c6f95d0a34`):
```
entitysdk.exception.EntitySDKError: Uploading these files failed: {'morphologies/merged-morphologies.h5': 'https://entitycore-data-staging.s3.amazonaws.com/private/e6030ed8-a589-4be2-80a6-f975406eb1f6/2720f785-a3a2-4472-969d-19a53891c817/assets/circuit/1edc30fd-c785-49c3-ad2e-50851d07887a/sonata_circuit/morphologies/merged-morphologies.h5?AWSAccessKeyId=[REDACTED]&Signature=[REDACTED]&x-amz-security-token=[REDACTED]&Expires=1776904867'}
```

## Technical implementation & specs
Proposed file size limit: 200GB (which is the max. disk space in the launch-system)

## Acceptance criteria
- [ ] Directory assets with files up to 200GB can be uploaded

GianlucaFicarelli:
This PR adds support for multipart upload of directory assets, enabling upload of large files within directories using presigned URLs and S3 multipart upload.

### New endpoints

- `POST /{entity_type}/{entity_id}/assets/directory/multipart-upload/initiate` — Initiates a multipart upload for a directory. Creates a parent directory asset and child assets for each file, returning presigned URLs for all parts.
- `POST /{entity_type}/{entity_id}/assets/{asset_id}/directory/multipart-upload/complete` — Completes the multipart upload for all files in the directory and transitions the parent asset to `CREATED` status.

### Key changes

**Schema & model changes:**
- Added `parent_id` column to `Asset` model with self-referential FK and `children` relationship (cascade delete)
- Removed `AssetStatus.DELETED` because unused: assets are always hard-deleted from the DB
- Replaced partial unique indexes with regular unique indexes on `asset.full_path` and `(path, entity_id)`
- Entity's `assets` relationship now filters by `parent_id IS NULL` instead of `status != DELETED`
- Added `MultipartDirectoryUpload`, `NestedInitiateUploadRequest`, `MultipartDirectoryAssetAndPresignedURLS` schemas
- Added `AssetLabel.directory_child` and `ContentType.other` (`application/octet-stream`)
- Replaced `validate_path`/`validate_full_path` with `validate_relative_path` and `validate_path_component`
- Moved `DirectoryUpload` validation (empty files, duplicates, path traversal) into the Pydantic schema layer

**Service refactoring:**
- Extracted `create_entity_asset_unverified` (skips entity permission check) from `create_entity_asset` (from https://github.com/openbraininstitute/entitycore/pull/598)
- Extracted `validate_uploadfile_for_small_entity_post` and `upload_entity_asset` from the router into the service layer (from https://github.com/openbraininstitute/entitycore/pull/598)
- Renamed `delete_asset` to `delete_asset_unverified`, removed `delete_asset_storage_object` (unused, handled via `app.db.events`)
- Asset storage deletion for directories handled in `app.db.events` (skips S3 key deletion for directories, children cascade)

**Validation improvements:**
- `InitiateUploadRequest.filename` now uses `PathComponentStr` (single component only)
- `InitiateUploadRequest.filesize` now validated at schema level with `lt=S3_MULTIPART_UPLOAD_MAX_SIZE`
- `InitiateUploadRequest.content_type` now typed as `ContentType | None` instead of `str | None`
- `DirectoryUpload.directory_name` uses `PathComponentStr` with `min_length=1`
- `DirectoryUpload.files` uses `RelativePathStr` with `min_length=1` and uniqueness validator


alembic/versions/20260515_173039_c7490722832a_multipart_upload_for_directories.py
```diff
@@ -0,0 +1,255 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "c7490722832a"
+down_revision: Union[str, None] = "b8ebf9a4e3da"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Purge assets in status DELETED
+    op.execute(text("""DELETE FROM asset WHERE status='DELETED'"""))
```
mgeplf: why does this need a `"""` string?

docs/multipart-upload.md
```diff
@@ -0,0 +1,206 @@
...
+3. **Complete** — The client calls the complete endpoint with the parent asset ID. The server:
+   - Verifies and completes the multipart upload for each child asset.
+   - Updates the parent directory asset status to `CREATED`.
+
+### Key differences from single file upload
+
+- The `directory_name` must be a single path component (no nesting).
+- File paths within the directory are relative to the directory root and can be nested (e.g., `subdir/file.txt`).
+- Child assets use the label `directory_child` and are not returned in the entity's asset listing (only the parent directory is listed).
+- Content type restrictions are relaxed for directory children — if the content type cannot be determined from the extension, it falls back to `application/octet-stream`.
```
mgeplf: is there a reason we use the extension, and don't allow the user to specify?

docs/multipart-upload.md
```diff
@@ -0,0 +1,206 @@
...
+```
+POST /{entity_type}/{entity_id}/assets/directory/multipart-upload/initiate
+POST /{entity_type}/{entity_id}/assets/{asset_id}/directory/multipart-upload/complete
+```
+
+### Flow
+
+1. **Initiate** — The client sends a request with the directory name, label, and a list of files (each with filename, filesize, sha256_digest, preferred_part_count). The server:
+   - Creates a parent directory asset with status `UPLOADING`.
+   - For each file, creates a child asset with status `UPLOADING` and `parent_id` pointing to the parent.
```
mgeplf, line 97: wrt `pointing to the parent.`; is this document aimed at entitycore devs, or users? if it's the latter, perhaps details like this aren't required?

app/db/events.py
```diff
@@ -36,6 +36,14 @@ def delete_assets_from_storage(session: Session):
     """
     to_delete: set[Asset] = session.info.pop(ASSETS_TO_DELETE_KEY, set())
     for asset in to_delete:
+        if asset.is_directory:
+            # No need to delete a directory key from S3. However, if the files in a directory:
+            # - were uploaded with multipart-upload, they are registered in the database
+            #   and are going to be deleted automatically.
```
mgeplf: `deleted automatically` by s3? or some entitycore process?

app/service/asset.py
```diff
@@ -189,21 +210,155 @@ def create_entity_asset(  # noqa: PLR0913
...
+    storage_type: StorageType,
+    full_path: str | None = None,
+    status: AssetStatus = AssetStatus.CREATED,
+    parent_id: uuid.UUID | None = None,
+) -> AssetRead:
+    """Create an asset for the specified entity id and type.
+
+    Args:
+        repos: Repository group for database access.
+        user_context: User context for authorization and audit.
```
mgeplf: `and audit`?

app/service/asset.py
```diff
@@ -131,12 +138,11 @@ def get_writable_entity_db_asset(
     return asset
 
 
-def create_entity_asset(  # noqa: PLR0913
+def create_entity_asset_unverified(  # noqa: PLR0913
```
mgeplf, line 141: only used by this module, add a `_` prefix?

app/service/asset.py
```diff
@@ -623,3 +724,127 @@ def create_asset_download_redirect(
...
+            content_type = ContentType.other
+        # create the asset entry in the db for each file
+        asset_read = create_entity_asset_unverified(
+            repos=repos,
+            user_context=user_context,
+            entity=entity,
+            filename=str(Path(json_model.directory_name, initiate_payload.filename)),
+            content_type=content_type,
+            size=initiate_payload.filesize,
+            sha256_digest=initiate_payload.sha256_digest,
```
mgeplf, line 780: should we validate the `sha256_digest`?

app/service/asset.py
```diff
@@ -238,95 +399,44 @@ def delete_asset(
     return AssetRead.model_validate(asset)
 
 
-def delete_asset_storage_object(asset: AssetRead, storage_client_factory: StorageClientFactory):
-    """Delete asset storage object."""
-    # TODO: Handle directories. See https://github.com/openbraininstitute/entitycore/issues/256
-    storage = storages[asset.storage_type]
-    # delete the file from S3 only if not using an open data storage
-    if not storage.is_open:
```
mgeplf, line 246: I'm probably not looking right, but where is the equivalent of this check?
are we relying on `delete_asset_storage_object`?

mgeplf (commented):
looks good, I thought it would be harder

app/service/asset.py
```diff
@@ -623,3 +724,127 @@ def create_asset_download_redirect(
...
+            label=initiate_payload.label,  # same as json_model.label
+            is_directory=False,
+            storage_type=storage.type,
+            status=AssetStatus.UPLOADING,
+            parent_id=parent.id,
+        )
+        # create presigned urls using the part count hint and filesize
+        # asset schemas is updated with the upload metadata
+        # Note: User already authorized when creating the asset
+        asset_read_with_upload_meta = entity_asset_multipart_upload_initiate(
```
eleftherioszisis, line 791: What would happen if one of these failed?

app/service/asset.py
```diff
@@ -66,7 +73,7 @@ def get_entity_assets(
     )
     apply_filter_query_operations = lambda q: q.join(Entity, Entity.id == Asset.entity_id).where(
         Asset.entity_id == entity_id,
-        Asset.status != AssetStatus.DELETED,
+        Asset.parent_id.is_(None),
```
eleftherioszisis, line 76: It is intentional that the respective admin function allows fetching all assets, correct?

alembic/versions/20260515_173039_c7490722832a_multipart_upload_for_directories.py
GianlucaFicarelli: The `"""` aren't needed, it's just a habit to wrap custom queries in triple quotes because they often become multi-line.

docs/multipart-upload.md
GianlucaFicarelli: This note isn't correct, because it's possible to specify an optional `content_type` in `MultipartDirectoryFileRequest`.
If it isn't in the list of allowed content-types, it fallbacks to `application/octet-stream`.
I'll update that point.

Separate point: it's also possible to specify an optional `sha256_digest`, that is optional mainly because it can be expensive to compute for big files. Thoughts about making it mandatory?

app/db/events.py
GianlucaFicarelli: The file is deleted automatically by the code in the event listener itself: it calls `delete_asset_storage_object` for each file, in the same way as it's done for regular files.
Possible downside: the deletion of many files could require some time to complete, but we should verify if that's an issue.

docs/multipart-upload.md
GianlucaFicarelli: afaik we don't have separate docs for users and devs, and in this case this point doesn't seem too technical to me.
In general it could be worth to improve and reorganize the full docs, we can discuss it offline.

app/service/asset.py
GianlucaFicarelli: audit in this context juts means that it's possible to know what is the user that created the asset. I can remove it if it can be confused with the full auditing system that was discussed before.

app/service/asset.py
GianlucaFicarelli: Fair, but for reducing conflicts it's taken almost as is from https://github.com/openbraininstitute/entitycore/pull/598/changes where it's called also from `app/service/admin.py`

app/service/asset.py
GianlucaFicarelli: We cannot verify sha256_digest in the initiate endpoint, but we could decide to:
- make it mandatory (see other comment)
- verify it in the complete endpoint (we already check the size, not sure if we can get the sha256 digest of the full size from AWS without downloading it, need s to be verified)

app/service/asset.py
GianlucaFicarelli: Yes, this is dead code. This condition is verified in `delete_asset_storage_object` that is called only from events.py

app/service/asset.py
GianlucaFicarelli: Good point, if some requests succeed and one fails, the init metadata will remain in AWS.
The cost for the storage should be low, since nothing is uploaded, but a possible approach to cleanup all the incomplete multipart uploads could be to configure a lifecycle cleanup rule in S3, that would delete also any other multipart upload initiated but not completed.
An alternative could be to implement the rollback for S3 data in entitycore, but it seems more fragile, and it can still fail in case of unexpected or random errors.

app/service/asset.py
GianlucaFicarelli: It can be decided. Since the admin endpoint for /assets was returning all the assets (including the deleted ones), it seems ok to keep the same approach and return everything.

app/service/asset.py
GianlucaFicarelli: Created https://github.com/openbraininstitute/INFRA/issues/546 for lifecycle cleanup

app/service/asset.py
GianlucaFicarelli: Removed in the last commits to avoid ambiguity

app/db/events.py
GianlucaFicarelli: I rephrased the comment in the last commits

docs/multipart-upload.md
GianlucaFicarelli: Updated the note with the current behaviour in the last commits

alembic/versions/20260515_173039_c7490722832a_multipart_upload_for_directories.py
GianlucaFicarelli: Replaced with single `"`

app/service/asset.py
eleftherioszisis: Yeah, agreed.

app/service/asset.py
mgeplf: ah, I meant just make sure it's the right length, and has the right characters.  Mainly to prevent ppl putting in `this is a placeholder` or something.

docs/multipart-upload.md
mgeplf: > Thoughts about making it mandatory?

I think it would be worthwhile to make it mandatory; yes incorrect ones could be maliciously added, but that's the tradeoff price we'd pay for not having to compute them ourselves.

app/service/asset.py
mgeplf: k, thanks

mgeplf (approved):

app/service/asset.py
GianlucaFicarelli: Good point, it can be validated directly in the pydantic schema if not already

eleftherioszisis (approved):

PR #610, opened 2026-05-18
Turn Derivation into an Identifiable

eleftherioszisis:


alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,152 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "0929d64df63e"
+down_revision: Union[str, None] = "b8ebf9a4e3da"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Promote ``derivation`` to ``Identifiable``: add ``id`` PK plus ``created_by_id``,
```
mgeplf, line 28: I suppose my new AI tell is using double backticks :)

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,152 @@
...
+        "derivation",
+        sa.Column(
+            "id",
+            sa.Uuid(),
+            server_default=sa.text("gen_random_uuid()"),
+            nullable=True,
+        ),
+    )
+
+    # 2. Add ``created_by_id`` / ``updated_by_id`` as nullable so we can backfill them next.
```
mgeplf, line 46: I also assume I'll have to start ignoring obvious comments in AI files; I suppose for migrations, I'll learn

app/service/derivation.py
```diff
@@ -1,34 +1,250 @@
...
+if TYPE_CHECKING:
+    from app.filters.base import Aliases
+    from app.schemas.auth import UserContext
+
+
+def _project_ids(user_context: UserContextDep) -> list[uuid.UUID]:
+    """Project ids used for authorization, matching ``constrain_to_writable_entities``."""
+    if user_context.project_id:
+        return [user_context.project_id]
+    return user_context.user_project_ids
```
mgeplf: strange, why doesn't `user_context.user_project_ids` already handle this?
I guess it's b/c if it's explicitly specified?

I'm unclear as why this endpoint works differently wrt `user_context.project_id` and `user_context.user_project_ids` from other endpoints.

mgeplf (commented):
looks reasonable to me, but I think I'm missing something wrt `user_project_id`

app/service/derivation.py
eleftherioszisis: Yes, the issue is that we differentiate between an explicit project_id and the project ids from the token groups. When a specific project_id context is passed it is expected to restrict things in that project.

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: The numbering too. :)

app/service/derivation.py
mgeplf: Ok, then I think a function for both `constrain_to_writable_entities` and this should be made, so that they use the same logic, and then it'd be clear, imo.

eleftherioszisis: Refactored auth query logic. Improved?

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,154 @@
...
+from sqlalchemy import Text
+import app.db.types
+
+# revision identifiers, used by Alembic.
+revision: str = "1df0fd1302f9"
+down_revision: Union[str, None] = "c7490722832a"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+OBI_ADMIN_PREF_LABEL = "OBI"
```
GianlucaFicarelli, line 24: `OBI_ADMIN_PREF_LABEL` is defined but not used in the queries below

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,154 @@
...
+    #    transient server default so every pre-existing row gets a fresh UUID. The column is
+    #    initially nullable to keep the ``ADD COLUMN`` cheap on large tables; we'll flip it
+    #    to NOT NULL after the implicit backfill from the server default.
+    op.add_column(
+        "derivation",
+        sa.Column(
+            "id",
+            sa.Uuid(),
+            server_default=sa.text("gen_random_uuid()"),
+            nullable=True,
```
GianlucaFicarelli: Cannot the id be generated directly with nullable=False, and the first alter_column below be removed, since there is a server_default?

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,154 @@
...
+        ["created_by_id"],
+        ["id"],
+    )
+    op.create_foreign_key(
+        op.f("fk_derivation_updated_by_id_agent"),
+        "derivation",
+        "agent",
+        ["updated_by_id"],
+        ["id"],
+    )
```
GianlucaFicarelli, line 140: To keep the uniqueness of `(used_id, generated_id)` as before (as it was the primary key), we should add a unique constraint on the same columns.
Alternatively, we could set the unique constraint on `(used_id, generated_id, derivation_type)` if we need to allow multiple associations between the same entities but different derivation types.
cc @darshanmandge @chr-pok could you clarify what should be the unique constraint for the derivation association?

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,154 @@
...
+    op.drop_constraint(op.f("fk_derivation_created_by_id_agent"), "derivation", type_="foreignkey")
+    op.drop_index(op.f("ix_derivation_used_id"), table_name="derivation")
+    op.drop_index(op.f("ix_derivation_updated_by_id"), table_name="derivation")
+    op.drop_index(op.f("ix_derivation_generated_id"), table_name="derivation")
+    op.drop_index(op.f("ix_derivation_creation_date"), table_name="derivation")
+    op.drop_index(op.f("ix_derivation_created_by_id"), table_name="derivation")
+
+    # Restore the old composite primary key on (used_id, generated_id). This will fail if
+    # the table contains duplicate (used_id, generated_id) pairs that became possible after
+    # the upgrade; such rows must be reconciled manually before downgrading.
```
GianlucaFicarelli: See comment above about the uniqueness.

app/db/auth.py
```diff
@@ -25,69 +86,66 @@ def constrain_to_writable_entities[Q: Query | Select](
...
+            is_service_maintainer=user_context.is_service_maintainer,
+        )
     )
 
-    if user_context.is_service_maintainer:
-        return query.where(
-            db_model_class.authorized_project_id.in_(project_ids),
-        )
 
+def constrain_to_accessible_entities[Q: Query | Select](
```
GianlucaFicarelli: The name `accessible_entities` can be ambiguous: the name of a similar function was previously changed from accessible to readable for that reason.

app/db/auth.py
```diff
@@ -25,69 +86,66 @@ def constrain_to_writable_entities[Q: Query | Select](
...
+    *,
+    query: Q,
+    user_context: UserContext,
+    db_model_class: type[HasAuth] = Entity,
+) -> Q:
+    """Ensure a query is filtered to rows that are viewable by the user.
+
+    TODO: Consolidate with constrain_to_readable_entities and adapt read_many calls
```
GianlucaFicarelli: Do you plan to open a separate PR for consolidating it?

app/db/auth.py
```diff
@@ -25,69 +86,66 @@ def constrain_to_writable_entities[Q: Query | Select](
...
         If one is provided query will be constrained within that single project_id.
     """
-    project_ids = (
-        [user_context.project_id] if user_context.project_id else user_context.user_project_ids
+    return query.where(
+        is_writable_by_user(
+            model=db_model_class,
+            project_ids=user_context.authorized_project_ids,
+            is_service_maintainer=user_context.is_service_maintainer,
+        )
```
GianlucaFicarelli, line 94: Nice cleanup

app/db/auth.py
```diff
@@ -1,19 +1,80 @@
...
+def is_public(model: type[HasAuth]) -> SQLColumnExpression[bool]:
+    return model.authorized_public == true()
+
+
+def is_private(model: type[HasAuth]) -> SQLColumnExpression[bool]:
+    return model.authorized_public == false()
+
+
+def is_in_projects(model: type[HasAuth], project_ids: list[UUID4]) -> SQLColumnExpression[bool]:
+    return model.authorized_project_id.in_(project_ids)
```
GianlucaFicarelli: For extra safety it would be better to guard against en empty list of project_ids.
In that case, it should be fine to return a False expression.

app/db/utils.py
GianlucaFicarelli, line 200: Unrelated to the PR but mentioning it here anyway: I wonder if it's possible to avoid the iteration to make this function more efficient, since it's called in many endpoints.

app/queries/entity.py
```diff
@@ -3,9 +3,38 @@
...
+    constrain_entity_query_to_project,
+    constrain_to_accessible_entities,
+    constrain_to_readable_entities,
+)
 from app.db.model import Entity
 from app.errors import ensure_result
+from app.schemas.auth import UserContext
+
+
+def get_accessible_entity[T: Entity](
```
GianlucaFicarelli, line 16: Similar to the comment above, prefer "readable" instead of "accessible".
There is some overlap with `get_readable_entity` that already exists and needs project_id instead of user_context, but the purpose is very similar.
If we need to keep both for the moment, a clearer name should be found.

app/schemas/derivation.py
```diff
@@ -55,7 +57,15 @@ def label_matches_derivation_type(self):
         return self
 
 
-class DerivationRead(DerivationBase):
+DerivationUserUpdate = make_update_schema(DerivationCreate, "DerivationUserUpdate")  # pyright: ignore [reportInvalidTypeForm]
+DerivationAdminUpdate = make_update_schema(
+    DerivationCreate,
+    "DerivationAdminUpdate",
+    excluded_fields=set(),
```
GianlucaFicarelli: I think we should exclude `used_id` and `derived_id` from the update, and probably also `derivation_type`, because they are the core fields of the association, and in those cases it would be clearer if the association is deleted and recreated.

app/service/derivation.py
```diff
@@ -146,80 +349,47 @@ def create_one(
...
-    generated_entity = get_writable_entity(
-        db,
-        Entity,
-        params.generated_id,
-        user_context.project_id,
-    )
-    db_model_class = Derivation
+    id_: uuid.UUID,
+) -> DeleteResponse:
+    query = _load(sa.select(Derivation).where(Derivation.id == id_))
```
GianlucaFicarelli: This query would apply multiple joins for used, generated, created_by, updated_by, but only generated is needed in the following logic.

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: Very good point. It makes sense to include the derivation_type too in the uniqueness constraint.

app/db/auth.py
eleftherioszisis: Yes

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
chr-pok: I am not aware of any examples atm where we would create multiple associations between the same entities with different types. But I guess such cases could exist in the future, so yeah, it makes sense to include the derivation_type in the uniqueness constraint

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # 6. Drop the transient server default on ``id`` so the schema matches a freshly created
+    #    table (the application generates UUIDs in Python via ``app.utils.uuid.create_uuid``).
+    op.execute("ALTER TABLE derivation ALTER COLUMN id DROP DEFAULT")
+
+    # 7. Swap the primary key from (used_id, generated_id) to (id).
+    op.drop_constraint(op.f("pk_derivation"), "derivation", type_="primary")
+    op.create_primary_key(op.f("pk_derivation"), "derivation", ["id"])
+
+    # 8. Indexes and foreign keys, identical to the auto-generated layout.
```
GianlucaFicarelli: It could be useful to add an index on derivation_type, since it's almost always used for filtering the derivations.
Ideally we should test if it's worth, considering the number of records and types, but these are going to increase.

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: Added in query.

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: Indeed. Changed it.

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: Added unique constraint for all three.

app/db/auth.py
eleftherioszisis: Renamed

app/queries/entity.py
eleftherioszisis: Renamed

app/service/derivation.py
eleftherioszisis: Changed to only joining generated

alembic/versions/20260521_111545_1df0fd1302f9_derivation_identifiable.py
eleftherioszisis: Added

app/db/auth.py
eleftherioszisis: I added it.

app/schemas/derivation.py
eleftherioszisis: done

app/db/utils.py
GianlucaFicarelli: Opened https://github.com/openbraininstitute/entitycore/issues/613, I'll open a small PR

app/service/derivation.py
```diff
@@ -1,34 +1,237 @@
...
+
+
+def _load(query: sa.Select):
+    return query.options(
+        joinedload(Derivation.used),
+        joinedload(Derivation.generated),
+        joinedload(Derivation.created_by),
+        joinedload(Derivation.updated_by),
+        raiseload("*"),
+    )
```
GianlucaFicarelli, line 57: All these `joinedload`and also others in the same file should have `innerjoin=True`

GianlucaFicarelli (commented):
One additional comment but it LGTM after that

app/service/derivation.py
eleftherioszisis: Added

GianlucaFicarelli (approved):

PR #611, opened 2026-05-21
Handle empty files in directory upload

GianlucaFicarelli:
Add support for empty files to https://github.com/openbraininstitute/entitycore/pull/608

The `upload_meta` in the db will contain:
`{"part_size": 0, "upload_id": "", "part_count": 0}`

The returned `upload_meta` will contain:
`{"part_size": 0, "parts": [{"part_number": 0, "url": "<presigned_url>"}]}`

In this way the client can process the presigned url in the same way as for multipart uploads.
Alternative: it would be possible to add a type to upload_meta, but it requires a change to the UploadMeta and UploadMetaRead schemas.


app/service/asset.py
```diff
@@ -639,6 +640,127 @@ def entity_asset_multipart_upload_complete(
...
+            is_directory=False,
+        )
+    except Exception as e:
+        raise ApiError(
+            message="Failed to check object.",
+            error_code=ApiErrorCode.GENERIC_ERROR,
+            http_status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+        ) from e
+
+    if check_result["exists"] is False:
```
mgeplf: `is False`? can `check_result["exists"]` be false-y but not `False`?

mgeplf (approved):

app/service/asset.py
GianlucaFicarelli: `exists` can be only True or False.
Do you mean that `if not check_result["exists"]` would be more readable than `if check_result["exists"] is False`?
A small improvement could be to specify the returned type of `check_object` as a TypedDict, so it's clear what are the available fields and their types, although the function is used only internally.

mgeplf: yeah, sorry I wasn't clear, I thought https://docs.astral.sh/ruff/rules/true-false-comparison/ would catch it, but I guess we have that off since comparing to True/False is needed w/ sqlalchemy.

The dict return is ok, I was just surprised.

mgeplf: you don't want to change this? `not check_result["exists"]` seems more normal to me, for the same reasons as in https://docs.astral.sh/ruff/rules/true-false-comparison/

GianlucaFicarelli: Agreed, although the linked rule is "against using the equality operators == and != to compare values to True or False", not against using `is False` or `is True`.

mgeplf (approved):

eleftherioszisis (approved):

PR #612, opened 2026-05-21
Make sha256_digest mandatory in MultipartDirectoryFileRequest, and stricter validation

GianlucaFicarelli:


app/schemas/asset.py
```diff
@@ -67,6 +75,14 @@ def validate_path_component_str(s: str) -> str:
 
 PathComponentStr = Annotated[str, AfterValidator(validate_path_component_str)]
 RelativePathStr = Annotated[str, AfterValidator(validate_relative_path_str)]
+Sha256Digest = Annotated[
+    str,
+    Field(
+        description="SHA256 digest of the file content in hexadecimal format.",
+        pattern=r"^[a-f0-9]{64}$",
```
mgeplf: hex can be upper case, afaik

GianlucaFicarelli: Yes I was considering that, but I was wondering if it was useful to force a conversion to lower case in BeforeValidator for consistency.
The conversion shouldn't be needed because the string is converted to bytes with `bytes.fromhex` in
https://github.com/openbraininstitute/entitycore/blob/d2986f9dd735f08e7f2d3843ba5312bc19e2f16c/app/repository/asset.py#L38
and when the digest is read from the db, it is converted with `x.hex()` that always returns lowercase.

mgeplf (approved):

PR #618, opened 2026-05-27
Avoid iteration in get_declaring_class

### LINKED ISSUE #613: Improve get_declaring_class by avoiding iterations
## Description
In db.utils, improve get_declaring_class by avoiding the iteration over all the registry mappers, to make this function more efficient, since it's called in many endpoints.

Current code:
https://github.com/openbraininstitute/entitycore/blob/01236cecd7b515f666ce984c0593a2ae176a21cf/app/db/utils.py#L190-L204

## Acceptance criteria
- [ ] No iterations

eleftherioszisis:
Fixes #613

### DIFF
--- a/app/db/utils.py
+++ b/app/db/utils.py
@@ -193,11 +193,11 @@ def has_project_id_in_columns(db_model_class) -> bool:
     return "authorized_project_id" in db_model_class.__mapper__.columns
 
 
-def get_declaring_class(
-    db_model_class: type[Identifiable], column_name: str
+def get_authorized_project_id_declaring_class(
+    db_model_class: type[Identifiable],
 ) -> type[HasAuth] | None:
     """Return the class that has a table with project id or None."""
     if not has_project_id_in_columns(db_model_class):
         return None
 
-    return class_mapper(db_model_class).get_property(column_name).parent.class_
+    return class_mapper(db_model_class).get_property("authorized_project_id").parent.class_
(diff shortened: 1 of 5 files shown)


app/db/utils.py
GianlucaFicarelli: Discussed separately, commenting for reference: `column_name` should be removed as it's always passed as `authorized_project_id`, and the call to `has_project_id_in_columns` would be incorrect if `column_name` has a different value.

app/db/utils.py
```diff
@@ -195,11 +200,4 @@ def get_declaring_class(
...
 
-    declaring_table = db_model_class.__mapper__.columns[column_name].table
-
-    # Iterate all mappers registered with the Base registry
-    for mapper in db_model_class.registry.mappers:
-        if mapper.local_table is declaring_table:
-            return mapper.class_
-
-    return None
+    return class_mapper(db_model_class).get_property(column_name).parent.class_
```
GianlucaFicarelli: Discussed separately, commenting for reference:
it would be safer to add explicit tests for models using STI: for example CellMorphologyProtocol and ModifiedReconstructionCellMorphologyProtocol

eleftherioszisis: done

app/db/utils.py
eleftherioszisis: done

GianlucaFicarelli (approved):

PR #620, opened 2026-06-01
Enable is_service_maintainer when auth is disabled

GianlucaFicarelli:
Enable is_service_maintainer when auth is disabled, and a couple of minor fixes.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -73,7 +73,7 @@ class Settings(BaseSettings):
 
     API_ASSET_POST_MAX_SIZE: int = 150 * MB
     PAGINATION_DEFAULT_PAGE_SIZE: int = 30
-    PAGINATION_MAX_PAGE_SIZE: int = 200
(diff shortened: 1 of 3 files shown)


app/config.py
```diff
@@ -73,7 +73,7 @@ class Settings(BaseSettings):
 
     API_ASSET_POST_MAX_SIZE: int = 150 * MB
     PAGINATION_DEFAULT_PAGE_SIZE: int = 30
-    PAGINATION_MAX_PAGE_SIZE: int = 200
+    PAGINATION_MAX_PAGE_SIZE: int = 1000
```
GianlucaFicarelli, line 76: PAGINATION_MAX_PAGE_SIZE=1000 matches the value in staging and production, so it can be set as the default value as well without needing to set the corresponding env variable locally.

mgeplf (approved):

PR #622, opened 2026-06-02
Allow updating measurement kinds in MeasurementAnnotation update

eleftherioszisis:
Issue: https://github.com/openbraininstitute/prod-platform-architecture/issues/210

### DIFF
--- a/app/service/measurement_annotation.py
+++ b/app/service/measurement_annotation.py
@@ -22,7 +22,7 @@ from app.db.model import (
     MeasurementKind,
     MeasurementLabel,
 )
-from app.db.utils import MEASURABLE_ENTITIES
+from app.db.utils import MEASURABLE_ENTITIES, MeasurableEntityType
 from app.dependencies.auth import AdminContextDep, UserContextDep, UserContextWithProjectIdDep
 from app.dependencies.common import PaginationQuery
 from app.dependencies.db import SessionDep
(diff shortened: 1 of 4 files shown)


app/service/measurement_annotation.py
```diff
@@ -77,6 +77,45 @@ def _update_measurement_label_ids(
     return measurement_annotation
 
 
+def _update_measurement_label_ids_on_update(
```
GianlucaFicarelli, line 94: Since the main difference with `_update_measurement_label_ids` is the potential absence of `entity_type` in the update model, wouldn't it be possible to reuse or factor out some common logic between the 2 functions?

app/service/measurement_annotation.py
```diff
@@ -302,7 +344,10 @@ def admin_update_one(
     user_context: AdminContextDep,
     db: SessionDep,
     id_: uuid.UUID,
-    json_model: MeasurementAnnotationAdminUpdate,  # pyright: ignore [reportInvalidTypeForm]
+    json_model: Annotated[
+        MeasurementAnnotationAdminUpdate,
+        Depends(_update_measurement_label_ids_on_update),
+    ],
 ) -> MeasurementAnnotationRead:
     def apply_operations(q):
```
GianlucaFicarelli, line 357: More related to https://github.com/openbraininstitute/entitycore/pull/616 rather than this PR, but if I'm not wrong most update endpoints already allow to update an entity without specifying the project context.
In this case instead it's still required, but it can be handled in a different PR.

app/service/measurement_annotation.py
eleftherioszisis: Refactored

GianlucaFicarelli (approved):

PR #624, opened 2026-06-05
Generic Workflow Result Entity

dkeller9:
### Summary

Adds a new `GenericResult` entity to entitycore, providing a flexible, schema-light result type
for cases where no specific entity fits. Unlike existing result types (`SimulationResult`,
`AnalysisNotebookResult`, etc.), `GenericResult` uses a free-form `data_payload` JSON field and a
`result_type` discriminator string, allowing it to serve as a base for future specialised result
subtypes without requiring new tables.

### Changes

- **`db/model.py`** — new `GenericResult` model inheriting from `Entity` and
  `NameDescriptionVectorMixin`, with `result_type` (polymorphic discriminator) and `data_payload`
  (nullable JSON) columns
- **`db/types.py`** — added `generic_result` to `EntityType` enum and an
  `ALLOWED_ASSET_LABELS_PER_ENTITY` entry permitting `other` assets
- **`filters/generic_result.py`** — `GenericResultFilter` and `NestedGenericResultFilter` with
  standard ordering and ilike search support
- **`schemas/generic_result.py`** — `GenericResultCreate`, `GenericResultRead`,
  `NestedGenericResultRead`, and update schemas
- **`service/generic_result.py`** — standard CRUD service wiring read, create, update, delete,
  and list operations
- **`routers/generic_result.py`** — user and admin routes registered under the `generic_result`
  route
- **`migrations/<timestamp>_<id>_generic_result.py`** — Alembic migration creating the
  `generic_result` table with `name`, `description`, `description_vector`, `result_type`, and
  `data_payload` columns, plus GIN and name indexes

### Notes

- `result_type` defaults to `"generic_result"` and can be used by callers to distinguish subtypes
  without requiring a new table or model
- `data_payload` is nullable in the DB and `dict[str, Any] | None` in the read schema accordingly
- Asset uploads are supported via the `other` label; more specific labels can be added per subtype
  in future


app/db/model.py
```diff
@@ -1611,6 +1612,16 @@ class SimulationExecution(Activity, ExecutionActivityMixin):
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
 
 
+class GenericResult(Entity, NameDescriptionVectorMixin):
+    __tablename__ = EntityType.generic_result.value
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    result_type = sa.Column(sa.String(50), nullable=False, default="generic_result")
```
eleftherioszisis: This is a type, therefore it should be a controlled. StrEnum

app/db/model.py
```diff
@@ -1611,6 +1612,16 @@ class SimulationExecution(Activity, ExecutionActivityMixin):
...
+    data_payload = sa.Column(sa.JSON, nullable=True, default=dict)
```
eleftherioszisis:
```suggestion
    data_payload: Mapped[JSON_DICT | None]
```
It would be better if it wasn't nullable though but rather set by default to an empty dict:
```python
data_payload: Mapped[JSON_DICT] = mapped_column(default={}, nullable=False, servcer_default="{}")
```

Is there no schema constraint for this json dict?

app/db/types.py
```diff
@@ -885,6 +888,15 @@ class LabelRequirements(BaseModel):
             )
         ],
     },
+    EntityType.generic_result: {
+        AssetLabel.generic_result: [
+            LabelRequirements(
+                content_type=ContentType.other,
+                is_directory=False,
+                description="Generic result asset.",
+            )
```
eleftherioszisis: `octet-stream` doesn't seem like the right format for the assets of the generic result. Could you omit the constraint entirely if there are no restrictions?

app/schemas/generic_result.py
```diff
@@ -0,0 +1,49 @@
...
+from app.schemas.utils import make_update_schema
+
+
+class GenericResultBase(BaseModel, NameDescriptionMixin):
+    model_config = ConfigDict(from_attributes=True)
+
+
+class GenericResultCreate(GenericResultBase, AuthorizationOptionalPublicMixin):
+    data_payload: dict[str, Any] = {}
+    result_type: str = "generic_result"
```
eleftherioszisis: Use StrEnum

app/filters/generic_result.py
```diff
@@ -0,0 +1,33 @@
...
+    IdFilterMixin,
+    ILikeSearchFilterMixin,
+    NameFilterMixin,
+)
+from app.filters.entity import EntityFilterMixin
+
+
+class NestedGenericResultFilter(IdFilterMixin, NameFilterMixin, CustomFilter):
+    class Constants(CustomFilter.Constants):
+        model = GenericResult
```
eleftherioszisis: Doesn't seem it is used.

app/db/model.py
dkeller9: Added GenericResultType(StrEnum) to app/db/types.py and updated the model, schema, and filter to use it.

app/db/model.py
dkeller9: Updated to Mapped[JSON_DICT] = mapped_column(default={}, nullable=False, server_default="{}") and removed | None from the read schema.

app/db/types.py
dkeller9: Changed the ALLOWED_ASSET_LABELS_PER_ENTITY entry for generic_result to None.

app/schemas/generic_result.py
dkeller9: Updated GenericResultCreate and GenericResultRead to use GenericResultType instead of str.

app/filters/generic_result.py
dkeller9: Removed NestedGenericResultFilter and NestedGenericResultFilterDep from app/filters/generic_result.py.

app/db/model.py
```diff
@@ -1656,6 +1658,18 @@ class SimulationExecution(Activity, ExecutionActivityMixin):
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
 
 
+class GenericResult(Entity, NameDescriptionVectorMixin):
+    __tablename__ = EntityType.generic_result.value
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    result_type = sa.Column(
+        sa.Enum(GenericResultType), nullable=False, default=GenericResultType.generic_result
+    )
```
eleftherioszisis:
```suggestion
    result_type: Mapped[GenericResultType]
```

app/db/types.py
```diff
@@ -889,6 +896,7 @@ class LabelRequirements(BaseModel):
             )
         ],
     },
+    EntityType.generic_result: None,
```
eleftherioszisis: I am afraid this is not going to work judging by https://github.com/openbraininstitute/entitycore/blob/db73afc917af08c378d630be9ff23b343b0c6aae/app/schemas/asset.py#L187

Would it be possible to explicitly list all the allowed labels/extensions we may allow?

app/db/model.py
```diff
@@ -1656,6 +1657,16 @@ class SimulationExecution(Activity, ExecutionActivityMixin):
     __mapper_args__ = {"polymorphic_identity": __tablename__}  # noqa: RUF012
 
 
+class TaskResult(Entity, NameDescriptionVectorMixin):
+    __tablename__ = EntityType.task_result.value
+
+    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("entity.id"), primary_key=True)
+    result_type: Mapped[TaskResultType]
```
eleftherioszisis:
```suggestion
    task_result_type: Mapped[TaskResultType]
```

app/db/types.py
```diff
@@ -145,6 +147,10 @@ class EntityType(StrEnum):
     task_config = auto()
 
 
+class TaskResultType(StrEnum):
+    task_result = auto()
```
eleftherioszisis: Here we should list result types like circuit_extraction__circuit perhaps?

dkeller9: Added some additional potential result types here.

dkeller9:
The tests are failing because the migration to task_result is incomplete, but that is something we can sort out at the end.

app/db/types.py
```diff
@@ -1047,6 +1069,25 @@ class LabelRequirements(BaseModel):
...
+    TaskResultType.ion_channel_simulation__result: None,
+    TaskResultType.em_synapse_mapping__result: None,
+    TaskResultType.aind_ephys_preprocessing__result: None,
+    TaskResultType.aind_ephys_spikesorting__result: None,
+    TaskResultType.extracellular_recording_weights_calculation__result: None,
+    TaskResultType.mesh_lod_generation__result: None,
+    TaskResultType.efeature_extraction__result: None,
+    TaskResultType.emodel_optimization__result: None,
+    TaskResultType.optimized_emodel_analysis_validation__result: None,
+    TaskResultType.circuit_synaptic_physiology_assignment__result: None,
```
eleftherioszisis: @dkeller9 These need to be updated with respective allowed labels per task result type. I have added a couple. Note that you can add more than just a single entity allowed labels. I just added those as examples.

dkeller9: I've added some extra allowed labels per task result type.

eleftherioszisis (approved):

PR #623, opened 2026-06-05
Add optional exercise_id to AnalysisNotebookTemplate

pgetta:
Adds a nullable, indexed uuid column referencing an exercise owned by a third-party (OBI grading service / customer Moodle).
Exposes `exercise_id` and `exercise_id__in` on the list filter so clients can fetch templates for one or more exercises.

Relates to [Avoid copy pasting token for moodle exercises.](https://github.com/openbraininstitute/PMO/issues/652).

PR #633, opened 2026-06-16
fix: change AnalysisNotebookTemplate exercise_id from UUID to string

pgetta:
## Why

The OBI Grading service identifies exercises with **free-form text ids, not UUIDs**. The original `exercise_id` implementation (#623) typed the column as `UUID`, which is incorrect — this fixes it.

## What

Switch `AnalysisNotebookTemplate.exercise_id` from `UUID` to a bounded string (`VARCHAR(255)`, non-empty, max 255), keeping it **nullable** and **indexed**.

| Layer | Change |
|---|---|
| ORM (`app/db/model.py`) | `Mapped[str \| None] = mapped_column(String(255), index=True)` |
| Schema (`app/schemas/analysis_notebook_template.py`) | `Annotated[str \| None, Field(min_length=1, max_length=255)]` |
| Filter (`app/filters/analysis_notebook_template.py`) | `exercise_id` / `exercise_id__in` → `str` / `list[str]` |
| Migration | new revision `209ed0ccfbd1`, casts `uuid -> varchar` via `exercise_id::text` (downgrade casts back with `::uuid`) |

### DIFF
--- a/alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
+++ b/alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
@@ -12,9 +12,6 @@ from alembic import op
 import sqlalchemy as sa
 
 
-from sqlalchemy import Text
-import app.db.types
-
 # revision identifiers, used by Alembic.
 revision: str = "209ed0ccfbd1"
 down_revision: Union[str, None] = "c9d905433b3e"
@@ -34,6 +31,16 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
+    conn = op.get_bind()
+    bad = conn.execute(
+        sa.text(
+            "SELECT COUNT(*) FROM analysis_notebook_template "
+            "WHERE exercise_id IS NOT NULL AND exercise_id !~* "
+            "'^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'"
+        )
+    ).scalar()
+    if bad:
+        raise RuntimeError(f"Cannot downgrade: {bad} exercise_id value(s) are not valid UUIDs")
     op.alter_column(
         "analysis_notebook_template",
         "exercise_id",
--- a/alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
+++ b/alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
@@ -1,0 +1,44 @@
+"""change analysis notebook template exercise_id to string
+
+Revision ID: 209ed0ccfbd1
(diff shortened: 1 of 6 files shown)


eleftherioszisis:
Do we control the OBI grading service? Does it have to use free-form text ids?

pgetta:
Yes, we do control it, but probably it's best to keep these human readable as they will be used by our clients across multiple places, e.g. definition in one place and copying/pasting into their Learning Management System.

That's the config example:
```json
{
  "tenant_id": "obi-moodle",
  "consumers": {...}
  },
  "exercises": {
    "obi.depolarization_rate.101": {...},
    ...
  }
}
```

There `obi.depolarization_rate.101` is the exercise ID. If copy/pasting went wrong this kind of ID is the best to quickly spot an issue visually.

But tech wise I agree that UUIDs are better.

alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
```diff
@@ -0,0 +1,44 @@
...
+        "analysis_notebook_template",
+        "exercise_id",
+        existing_type=sa.Uuid(),
+        type_=sa.String(length=255),
+        existing_nullable=True,
+        postgresql_using="exercise_id::text",
+    )
+
+
+def downgrade() -> None:
```
eleftherioszisis, line 33: Should this become instead:
```python
def downgrade() -> None:
    raise RuntimeError("Irreversible migration: UUID -> String conversion is not safely reversible")
```
I suppose it won't be possible to downgrade right?

app/schemas/analysis_notebook_template.py
```diff
@@ -42,7 +41,7 @@ class AnalysisNotebookTemplateBase(BaseModel, NameDescriptionMixin):
     model_config = ConfigDict(from_attributes=True)
     specifications: AnalysisNotebookTemplateSpecifications | None = None
     scale: AnalysisScale
-    exercise_id: uuid.UUID | None = None
+    exercise_id: Annotated[str | None, Field(min_length=1, max_length=255)] = None
```
eleftherioszisis, line 44: Would it be useful perhaps to also add a `exercise_id__ilike` for pattern matching?

alembic/versions/20260616_134302_209ed0ccfbd1_change_analysis_notebook_template_.py
pgetta: Good point, I've added a guarded raise, so that if there are only UUID valid values - downgrade will still work.

app/schemas/analysis_notebook_template.py
pgetta: I don't think there is a use case where this can be useful. At least for now.

eleftherioszisis (approved):

pgetta:
Thank you @eleftherioszisis.

PR #635, opened 2026-06-17
Add EModel Building extraction Workflow Result AssetLabels

darshanmandge:
- Adds `AssetLabel` for the extraction workflow result of the EModel Building pipeline.


app/db/types.py
```diff
@@ -509,6 +509,10 @@ class AssetLabel(StrEnum):
     task_config = auto()
     lod_mesh_block = auto()
     electrode_array_weight_matrix = auto()
+    efeature_extraction_features = auto()
+    efeature_extraction_recipe = auto()
```
eleftherioszisis: We shouldn't be storing the recipe of the entire pipeline here. Only the config needed for this step to generate the recipe on the fly.

app/db/types.py
```diff
@@ -1092,7 +1096,39 @@ class LabelRequirements(BaseModel):
...
+                description="Extracted e-features JSON (FitnessCalculatorConfiguration).",
+            ),
+        ],
+        AssetLabel.efeature_extraction_recipe: [
+            LabelRequirements(
+                content_type=ContentType.json,
+                is_directory=False,
+                description="BluePyEModel recipes.json used during extraction.",
+            ),
+        ],
```
eleftherioszisis: Same as above. We should not store this.

app/db/types.py
darshanmandge: I have removed the recipe entry here.

app/db/types.py
darshanmandge: I have removed the recipe AssetLabel.

app/db/types.py
```diff
@@ -1092,7 +1095,32 @@ class LabelRequirements(BaseModel):
...
+                ),
+            ),
+        ],
+        AssetLabel.efeature_extraction_cells: [
+            LabelRequirements(
+                content_type=ContentType.other,
+                is_directory=False,
+                description="BluePyEfe cells pickle (optional, for later conversion to HDF5).",
+            ),
+        ],
```
eleftherioszisis, line 1130: We should not allow pickle files here either.

eleftherioszisis: @darshanmandge could this be replaced with hdf5 files instead?

darshanmandge: Possibly yes. I have created another issue for that.  I will change it.

darshanmandge: I have changed pickle to h5 for the content type. I have also added another asset label with H5 as bluepyefe optionally generates 2 pickle files: cells.pkl and protocols.pkl.

eleftherioszisis (approved):

PR #636, opened 2026-06-17
Add release_version facet and release_version__in filter to EM cell mesh

pgetta:
## Summary

Adds two capabilities to the EM-cell-mesh list endpoint:

- **(a)** aggregates `release_version` into the `facets` response, so clients can see the distribution of versions and their counts
- **(b)** accepts a `release_version__in` query filter, so clients can list meshes for several versions at once

The exact-match `release_version` filter already existed; this only adds the `__in` variant and the facet.

## Implementation notes

`release_version` is the first facet on a **scalar base-table column** rather than a joined entity, which required two non-obvious touches:

- **`app/queries/factory.py`** — registered the facet with `id = EMCellMesh.release_version` and `label = sa.cast(EMCellMesh.release_version, sa.String)` (`Facet.label` is typed `str` and Pydantic v2 will not coerce an `int`), plus a no-op `filter_joins` entry (`lambda q: q`) since `_get_facets` forces a join on each facet key.
- **`app/dependencies/common.py`** — widened `FacetQueryParams.label` to `InstrumentedAttribute[str] | sa.ColumnElement[str]` so the cast type-checks.
- **`app/service/em_cell_mesh.py`** — added `"release_version"` to `facet_keys` (flows into `filter_keys` via the existing `*facet_keys` splat).
- **`app/filters/em_cell_mesh.py`** — one-line `release_version__in: list[int] | None`; fastapi-filter auto-wires the `IN` operator.

### Facet ordering

Facet entries sort by the **string** `label` (`orderby_keys=["label"]` is hardcoded in `_get_facets`), so `release_version` entries sort lexicographically (`1, 10, 2`). This is accepted as-is — no change to shared facet code. It only affects the order of entries inside the `facets.release_version` array; it does **not** affect the paginated `data` rows, the counts, or `release_version__in` filtering. The frontend re-sorts that small array numerically.

## Testing

- `tests/test_em_cell_mesh.py` — fixture rows now have distinct `release_version` values; `test_filtering` asserts the new facet entry and the `release_version__in=[1, 2]` result.
- Full suite: 1041 passed, total coverage 97.58% (gate 90%). `ruff` + `pyright` clean.

## Related issue

[Add column "version" to browsing of EMCellMeshes. Add filter for version](openbraininstitute/prod-explore-functionality#504)

### DIFF
--- a/app/dependencies/common.py
+++ b/app/dependencies/common.py
@@ -66,7 +66,8 @@ def forbid_extra_query_params(
 
 class FacetQueryParams(TypedDict):
     id: InstrumentedAttribute[uuid.UUID] | InstrumentedAttribute[int]
-    label: InstrumentedAttribute[str]
+    # ColumnElement covers cast() expressions for facets on scalar base-table columns
+    label: InstrumentedAttribute[str] | sa.ColumnElement[str]
     type: NotRequired[InstrumentedAttribute[str]]
 
 
--- a/app/filters/em_cell_mesh.py
+++ b/app/filters/em_cell_mesh.py
@@ -19,6 +19,7 @@ class EMCellMeshFilter(
     MeasurableFilterMixin,
 ):
     release_version: int | None = None
+    release_version__in: list[int] | None = None
     dense_reconstruction_cell_id: int | None = None
     level_of_detail: int | None = None
     mesh_type: EMCellMeshType | None = None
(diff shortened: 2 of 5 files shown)


eleftherioszisis (approved):
lgtm

pgetta:
Thanks Eleftherios!

PR #638, opened 2026-06-18
Add lifecycle_status to Entity and enforce schema consistency

eleftherioszisis:
* Add lifecycle_status in Entity (active/draft)
* Refactor schemas to centralize Identifiable/Entity schemas and reuse that to propagate the same fields and enforce consistency
* Fix service/tests accordingly

Issue: https://github.com/openbraininstitute/prod-contribute-fix-data/issues/80


app/schemas/agent.py
```diff
@@ -28,28 +27,32 @@ class PersonCreate(PersonBase):
...
+class CreatedByUpdatedByMixin(Schema):
     created_by: NestedPersonRead
     updated_by: NestedPersonRead
 
 
-class PersonRead(NestedPersonRead, CreationMixin, CreatedByUpdatedByMixin):
-    pass
+from app.schemas.identifiable import IdentifiableRead  # noqa: E402
 
+IdentifiableRead.model_rebuild(_types_namespace={"NestedPersonRead": NestedPersonRead})
```
GianlucaFicarelli: `IdentifiableRead.model_rebuild()` is called also in identifiable.py, so this call in agent.py seems redundant.

app/schemas/identifiable.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class IdentifiableRead(NestedIdentifiableRead):
+    created_by: NestedPersonRead
+    updated_by: NestedPersonRead
+
+
+from app.schemas.agent import NestedPersonRead  # noqa: E402
+
+IdentifiableRead.model_rebuild(_types_namespace={"NestedPersonRead": NestedPersonRead})
```
GianlucaFicarelli: Instead of passing _types_namespace to model_rebuild, it should work to:
- remove the import at the top:
```python
if TYPE_CHECKING:
    from app.schemas.agent import NestedPersonRead
```
- add ignore TC001 in the last import above:
```python
from app.schemas.agent import NestedPersonRead  # noqa: E402, TC001
```

app/schemas/contribution.py
```diff
@@ -32,16 +38,28 @@ class ContributionCreate(ContributionBase):
...
     contributions: list[NestedContributionRead] | None
+
+
+from app.schemas.agent import AgentRead  # noqa: E402
+from app.schemas.entity import NestedEntityRead  # noqa: E402
+
+NestedContributionRead.model_rebuild(_types_namespace={"AgentRead": AgentRead})
+ContributionRead.model_rebuild(
+    _types_namespace={"AgentRead": AgentRead, "NestedEntityRead": NestedEntityRead}
+)
```
GianlucaFicarelli: Instead of passing _types_namespace to model_rebuild, it should work to:
- remove the import at the top:
```python
if TYPE_CHECKING:
    from app.schemas.agent import AgentRead
    from app.schemas.entity import NestedEntityRead
```
- add ignore TC001 in the last import above:
```python
from app.schemas.agent import AgentRead  # noqa: E402, TC001
from app.schemas.entity import NestedEntityRead  # noqa: E402, TC001
```

app/schemas/agent.py
```diff
@@ -1,23 +1,22 @@
...
-    model_config = ConfigDict(from_attributes=True)
-
+class PersonBase(Schema):
     given_name: str | None = None
     family_name: str | None = None
     pref_label: str
 
 
-class PersonCreate(PersonBase):
+from app.schemas.identifiable import IdentifiableCreate, NestedIdentifiableRead  # noqa: E402
```
GianlucaFicarelli: It seems that this import can be moved at the top.

app/schemas/agent.py
```diff
@@ -28,28 +27,32 @@ class PersonCreate(PersonBase):
...
 
-class CreatedByUpdatedByMixin(BaseModel):
+class CreatedByUpdatedByMixin(Schema):
     created_by: NestedPersonRead
     updated_by: NestedPersonRead
 
 
-class PersonRead(NestedPersonRead, CreationMixin, CreatedByUpdatedByMixin):
-    pass
+from app.schemas.identifiable import IdentifiableRead  # noqa: E402
```
GianlucaFicarelli: It seems that this import can be moved at the top.

app/schemas/base.py
```diff
@@ -1,51 +1,47 @@
...
+class Schema(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
 
 
 class ActivityTypeMixin:
     type: ActivityType | None = None
 
 
-class EntityTypeMixin(BaseModel):
+class EntityTypeMixin(Schema):
```
GianlucaFicarelli: Would it work if the mixins are modified to not inherit from either BaseModel or Schema, so that they only define the attributes?
In that way, each model needs to inherit explicitly from Schema and add the required mixins.

app/schemas/identifiable.py
eleftherioszisis: done

app/schemas/agent.py
eleftherioszisis: done

app/schemas/contribution.py
eleftherioszisis: done

app/schemas/agent.py
eleftherioszisis: done

app/schemas/agent.py
eleftherioszisis: done

app/schemas/base.py
eleftherioszisis: It does indeed. I cleaned up and separated, I hope in a more readable way, the mixins from the schemas.

app/schemas/agent.py
GianlucaFicarelli: This import isn't moved yet or the change not pushed.

app/schemas/base.py
GianlucaFicarelli: Looks good, although I see that `EntityTypeMixin` isn't used, so it can be removed.
Also `ProjectContext` isn't used anymore, since only `OptionalProjectContext` is now used.

app/schemas/agent.py
eleftherioszisis: oh, sorry about that. I must have missed it. Moved now.

app/schemas/base.py
eleftherioszisis: I removed them.

app/schemas/agent.py
```diff
@@ -28,28 +25,30 @@ class PersonCreate(PersonBase):
...
 
-class NestedPersonRead(PersonBase, IdentifiableMixin):
-    type: str
+class NestedPersonRead(PersonBase, NestedIdentifiableRead):
+    type: AgentType
     sub_id: uuid.UUID | None
 
 
-class CreatedByUpdatedByMixin(BaseModel):
+class CreatedByUpdatedByMixin(Schema):
```
GianlucaFicarelli: This mixin still inherits from Schema although it shouldn't be needed.
Moreover, it's used only by NestedMEModelCalibrationResultRead so I wonder if it should be removed completely (because nested entities usually don't expose created_by and updated_by)

eleftherioszisis: I didn't want to remove things from schemas because it is likely that the frontend depends on them. I think the right way forward is to create an issue about that and ask if that is needed/used.

app/schemas/cell_morphology_protocol.py
```diff
@@ -46,13 +33,12 @@ class CellMorphologyProtocolMixin:
 # Base Models
 
 
-class CellMorphologyProtocolBase(BaseModel, NameDescriptionMixin):
-    model_config = ConfigDict(from_attributes=True)
-    type: Literal[EntityType.cell_morphology_protocol] = EntityType.cell_morphology_protocol
+class CellMorphologyProtocolBaseMixin(NameDescriptionMixin):
+    type: EntityType = EntityType.cell_morphology_protocol
```
GianlucaFicarelli, line 38: This validation is weaker than the previous one and it would allow a user to override the type in the creation/update endpoint.
Alternatively, is it possible to remove type completely from the create/update schemas? 
It shouldn't be a breaking change, since the type is ignored when not defined in the schema.

app/schemas/derivation.py
```diff
@@ -73,8 +74,8 @@ def label_matches_derivation_type(self):
 )  # pyright: ignore [reportInvalidTypeForm]
 
 
-class DerivationRead(DerivationBase, CreationMixin, CreatedByUpdatedByMixin, IdentifiableMixin):
-    used: BasicEntityRead
-    generated: BasicEntityRead
+class DerivationRead(DerivationBase, IdentifiableRead):
+    used: NestedEntityRead
+    generated: NestedEntityRead
```
GianlucaFicarelli, line 79: This changes from BasicEntityRead to NestedEntityRead and adds other attributes to the nested entities (authorized_project_id, authorized_public, lifecycle_status, creation_date, update_date).
It will increase the size of the payload, but the new fields are probably useful.

eleftherioszisis: yeah, I believe it won't hurt to have those around and they may be useful.

app/service/brain_region.py
```diff
@@ -24,6 +24,8 @@ def _load(select: sa.Select):
     return select.options(
         joinedload(BrainRegion.species),
         joinedload(BrainRegion.strain),
+        joinedload(BrainRegion.created_by, innerjoin=True),
+        joinedload(BrainRegion.updated_by, innerjoin=True),
```
GianlucaFicarelli: It doesn't seem that these additional joins and aliases below are needed with the model BrainRegionRead

app/schemas/agent.py
GianlucaFicarelli: Sure we can keep it as is to not introduce potential breaking changes, and handle it in a separate issue/PR.
Maybe @bilalesi or @pgetta can already confirm if any code in the frontend or in the simulator uses `MEModelRead.calibration_result.created_by` or `MEModelRead.calibration_result.updated_by` in the response to the read_one or read_many endpoint `/memodel`?
The current openapi docs are at:
https://staging.cell-a.openbraininstitute.org/api/entitycore/docs#/memodel/read_many_memodel_get

app/schemas/cell_morphology_protocol.py
eleftherioszisis: Added a validator

app/service/brain_region.py
eleftherioszisis: Indeed. Removed.

GianlucaFicarelli (approved):

app/schemas/agent.py
pgetta: I've checked both Core Web app and Bluenaas / small scale simulator - and none are using those nested properties.

PR #640, opened 2026-06-24
Support ordering circuits by target_simulator

pgetta:
## Summary

Enables ordering the `GET /circuit` list endpoint by `target_simulator`, which previously returned a `422` because the field was not whitelisted for sorting.

## Changes

- **`app/filters/circuit.py`** — add `target_simulator` to `CircuitFilter.Constants.ordering_model_fields`, so `order_by=target_simulator` / `order_by=-target_simulator` are accepted. (`target_simulator` was already a supported *filter*; this adds *sorting*.)
- **`app/db/types.py`** — reorder the `TargetSimulator` enum into alphabetical order (`Brian2, CORENEURON, LearningEngine, NEURON`). Native Postgres enums sort by definition order, so this makes the ascending sort intuitive and lexicographic.
- **`alembic/versions/…9e36650e03a9…`** — migration syncing the DB enum to the new order via `op.sync_enum_values`, with a symmetric downgrade. The `circuit.target_simulator` column and its `'NEURON'` server default are preserved.
- **`tests/test_circuit.py`** — circuits had no ordering test; add `test_ordering` that exercises every sortable field (asc/desc, count-preserving) and spot-checks `target_simulator` order against the enum's definition order. Also varies `target_simulator` across the `models` fixture and updates the affected filter assertion.
- **`scripts/export/build_database_archive.sh`** — refreshed head revision (auto-updated by `make migration`).

## Deployment note

`sync_enum_values` can't reorder a Postgres enum in place — it recreates the type and runs `ALTER TABLE circuit ALTER COLUMN … TYPE … USING …`, which takes an `ACCESS EXCLUSIVE` lock and rewrites the column for the duration. No data is at risk (no values removed/renamed), but it briefly blocks on a large `circuit` table.

## Testing

- `make lint` — ruff + pyright clean.
- `PYTEST_ADDOPTS="-k test_circuit" make test-local` — 52 passed, including the new `test_ordering` and the updated `test_filtering`, against a DB with the new migration applied.


app/db/types.py
```diff
@@ -1262,12 +1262,14 @@ class TargetSimulator(StrEnum):
...
+
+    Ensure the list is in alphabetical order to guarantee correct ordering.
     """
 
-    neuron = "NEURON"
+    brian2 = "Brian2"
     coreneuron = "CORENEURON"
     learning_engine = "LearningEngine"
-    brian2 = "Brian2"
+    neuron = "NEURON"
```
eleftherioszisis, line 1273: Does the ordering of the enum affect the filter ordering?

pgetta: It seems so, that's what I've learned recently: https://www.postgresql.org/docs/current/datatype-enum.html#DATATYPE-ENUM-ORDERING.

There is way to cast to text when doing "order by" query - but it has a performance impact. So re-ordering the enum is the easiest way for the moment.

Had a discussion on that with @GianlucaFicarelli - he suggested this approach.

eleftherioszisis: I see. Thanks for the info!

pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/217.

eleftherioszisis (approved):
LGTM

pgetta:
Thank you Eleftherios.

app/filters/circuit.py
```diff
@@ -66,6 +66,7 @@ class Constants(ScientificArtifactFilter.Constants):
             "name",
             "scale",
             "build_category",
+            "target_simulator",
```
GianlucaFicarelli, line 69: Note that there isn't an index for `target_simulator`, but it's the same also for other fields listed in ordering_model_fields, so we may add the index if we notice poor performances.

GianlucaFicarelli (approved):

app/filters/circuit.py
pgetta: Thank you Gianluca, that is good to know. I'll check the search performance once this is deployed.

pgetta:
Had to rebase and force push due to a conflicting migrations. Could you please re-approve?

pgetta:
Ah, I thought due to the rebase I've lost the approvals, but that's not the case. Merging now.

PR #642, opened 2026-06-25
Support querying and displaying circuits by derivation type

pgetta:
## Summary

Implements the backend for [openbraininstitute/prod-explore-functionality#517](https://github.com/openbraininstitute/prod-explore-functionality/issues/517) — the Circuits overview needs a **"Derivation type"** column and a **filter** by derivation type, general enough to absorb new derivation types.

This is derived entirely from the existing normalized `Derivation` table (`used` → `generated`, indexed `derivation_type`).

## Filter — narrow the circuit list by derivation type

Both directions are exposed, mirroring the `Derivation` model's roles:

- `?generated_derivation__derivation_type=circuit_extraction` — circuits that *were derived* this way ("how it was derived")
- `?used_derivation__derivation_type=circuit_extraction` — circuits that *are a source* of such a derivation ("what was derived from it")
- `__in` variants supported for both.

No source-type restriction: a circuit derived from a non-circuit source (e.g. `emodel_circuit`) surfaces that type too. Implemented with the established `query_params_factory` + `filter_joins` join pattern (enum, so it's a plain filter, not a facet).

## Read — opt-in derivation lists via `expand`

Derivation data loads only when requested, per direction, via the `expand` query param (same pattern as `cell_morphology` / `em_cell_mesh`):

- `?expand=generated_derivations` → each entry carries the source `used` entity
- `?expand=used_derivations` → each entry carries the resulting `generated` entity
- both can be requested together (repeated params: `?expand=generated_derivations&expand=used_derivations`)

Backed by two **view-only** relationships on `Circuit` plus load-aware properties: an un-expanded direction serializes as `null` (no extra query, never trips `raiseload`); an expanded-but-empty direction serializes as `[]`. Nothing loads on the default `GET /circuit`.

## Pagination fix (one-to-many filter joins)

The derivation filter join is one-to-many (a circuit can match several `Derivation` rows). The shared paginated id subquery (`_with_subquery`) selected ids without `DISTINCT`, so `OFFSET/LIMIT` paged over duplicated rows while `total_items` used `count(distinct id)` — a matching circuit could repeat across pages or be skipped. Fixed by selecting `DISTINCT` ids in the subquery, which also hardens the pre-existing `contribution` / `mtype` / `used` / `generated` one-to-many filters. `DISTINCT` is valid because every `ORDER BY` element is already in the subquery select list.


app/service/circuit.py
```diff
@@ -58,6 +70,15 @@ def _load(query: sa.Select):
...
     )
+    if expand and ExpandableAttribute.generated_derivations in expand:
+        query = query.options(
+            selectinload(Circuit.derivations_as_generated).joinedload(Derivation.used)
+        )
+    if expand and ExpandableAttribute.used_derivations in expand:
+        query = query.options(
+            selectinload(Circuit.derivations_as_used).joinedload(Derivation.generated)
+        )
+    return query
```
GianlucaFicarelli, line 81: For discussion: it would be nice if the possibility to filter by derivation type and get the list of derivations is made available to any entity.
However, it would require changes:
- to the _load function in every entity (change to many files, with many repetitions, so I would avoid that)
- or some changes in queries/common.py for read_many/read_one

Thoughts? Could it be useful already for other derivation types?
If the change adds too much complexity, the generalization can be postponed.

app/service/circuit.py
GianlucaFicarelli, line 87: Can it be useful to add the expand parameter to read_one as well, not only read_many?
For cell morphologies, it's present in read_one and not in read_many (for performance reasons), but it would be cleaner to be consistent across endpoints.

pgetta: Good idea, I've extended also the `read_one`

pgetta:
Closing in favour of [Generalize derivation filters and properties from Circuit to all Entities](https://github.com/openbraininstitute/entitycore/pull/647).

PR #647, opened 2026-06-25
Generalize derivation filters and properties from Circuit to all Entities

pgetta:
Alternative implementation of #642.

Where #642 adds the derivation-backed filters and expand-read properties to **Circuit** only, this PR lifts them up to the **`Entity`** base so **every** entity subclass inherits them. It also adds an entity-id filter (`used_id` / `generated_id`, + `__in`) alongside `derivation_type`.

- **Filter** — centralized in `EntityFilterMixin` + `factory.py` (joins auto-injected per entity model, applied only when the filter is set), so no per-service filter wiring.
- **Expand-read** — `generated_derivations` / `used_derivations` live on `EntityRead` / `EntityReadWoutAssets`, always present and `null` until `?expand=` is used; `expand` is wired into `read_one`, `admin_read_one`, `read_many`, `admin_read_many` across all entity services.
- No DB migration: the `derivation` table already exists and the new relationships are `viewonly`.


app/queries/common.py
```diff
@@ -81,6 +85,7 @@ def router_read_one[T: Schema, I: Identifiable](
         )
     if apply_operations:
         query = apply_operations(query)
+    query = apply_derivation_expand(query, db_model_class, expand)
```
eleftherioszisis, line 88: Shouldn't this be applied only to entities?

app/filters/base.py
```diff
@@ -240,7 +240,12 @@ def has_nested_filtering_field(self, name: str) -> bool:
             name: The name of the nested filtering field. It's possible to specify deeply nested
             filtering fields using the dot notation, e.g. "measurement_kind.pref_label".
         """
-        attr = attrgetter(name)(self)
+        try:
+            attr = attrgetter(name)(self)
+        except AttributeError:
+            # A join may be registered (e.g. the derivation joins, added for every entity model)
+            # without this filter declaring the matching field; treat it as not filtering.
+            return False
```
eleftherioszisis: I am not sure I understand why is this needed. Isn't it possible to avoid silently skipping errors? If nevertheless is essential could you cover it in tests so that it is clear what is the expected behavior of this bit?

app/filters/base.py
```diff
@@ -251,7 +256,10 @@ def get_nested_filter(self, name: str) -> "CustomFilter[T] | None":
             name: The name of the nested filter. It's possible to specify deeply nested filters
             using the dot notation, e.g. "measurement_annotation.measurement_kind".
         """
-        attr = attrgetter(name)(self)
+        try:
+            attr = attrgetter(name)(self)
+        except AttributeError:
+            return None
```
eleftherioszisis: Same here

app/filters/entity.py
```diff
@@ -1,15 +1,36 @@
...
+    Exposed on every entity filter as ``generated_derivation`` / ``used_derivation`` (see
+    EntityFilterMixin). Besides ``derivation_type``, the related-entity ids are filterable:
+    on ``generated_derivation`` the meaningful side is ``used_id`` ("derived from entity X"),
+    on ``used_derivation`` it is ``generated_id`` ("source of entity Y").
+    """
+
+    derivation_type: DerivationType | None = None
+    derivation_type__in: list[DerivationType] | None = None
+    used_id: uuid.UUID | None = None
+    used_id__in: list[uuid.UUID] | None = None
```
GianlucaFicarelli, line 26: Nice, this can be useful for selecting cell morphologies derived from one or more `EMDenseReconstructionDataset` with derivation_type `em_dense_reconstruction_dataset_cell_morphology` and support pagination.

See [this comment](https://github.com/openbraininstitute/prod-build-circuit/issues/29#issuecomment-4389847949)

app/db/model.py
```diff
@@ -634,6 +634,39 @@ def __table_args__(cls):  # noqa: D105, PLW3201
...
+    derivations_as_used: Mapped[list["Derivation"]] = relationship(
+        "Derivation",
+        primaryjoin=lambda: Entity.id == Derivation.used_id,
+        foreign_keys=lambda: [Derivation.used_id],
+        viewonly=True,
+        overlaps="used",
+    )
+
+    @property
+    def generated_derivations(self) -> list["Derivation"] | None:
```
GianlucaFicarelli: Thinking about naming, what about calling the properties:
`generated_from_derivations` and `used_by_derivations`?
Not much longer, but it looks slightly clearer to me.
Other ideas?

app/filters/circuit.py
```diff
@@ -56,6 +56,9 @@ class CircuitFilter(
     number_connections__lte: int | None = None
     number_connections__gte: int | None = None
 
+    # generated_derivation / used_derivation are inherited from EntityFilterMixin via
+    # ScientificArtifactFilter, so every entity filter exposes them.
+
```
GianlucaFicarelli: This comment isn't very useful since it's the same for every entity, so it can be removed,

app/queries/factory.py
```diff
@@ -47,6 +48,11 @@
 from app.queries.types import ApplyOperations
 
 
+def _is_entity_model(db_model_class: Any) -> bool:
+    """Whether the model is an Entity subclass (kept separate to avoid narrowing the caller)."""
+    return isinstance(db_model_class, type) and issubclass(db_model_class, Entity)
```
eleftherioszisis, line 53: Is it indeed the case that the type needs to be as broad as "Any"?

app/db/model.py
```diff
@@ -634,6 +634,39 @@ def __table_args__(cls):  # noqa: D105, PLW3201
...
+        """Derivations where this entity is the generated entity, or None if not expanded."""
+        if "derivations_as_generated" in sa.inspect(self).unloaded:
```
GianlucaFicarelli, line 667: Do we need that guard?
If the property is accessed without loading the derivations, it could be better to fail rather than returning  None.
Removing the guard can be acceptable only if we are sure that the property is accessed only when required.

app/queries/factory.py
```diff
@@ -74,6 +80,18 @@ def _get_alias[T: type[DeclarativeBase]](db_cls: T, name: str | None = None) ->
...
+    # Derivation aliases in `aliases` (mutating the dict the caller also hands to router_read_many)
+    # so the join lambdas below and `filter_model.filter(...)` resolve the *same* alias objects,
+    # and so the two directions don't collide on a single un-aliased Derivation table.
+    is_entity_model = _is_entity_model(db_model_class)
+    if is_entity_model and Derivation not in aliases:
+        aliases[Derivation] = {
+            "generated_derivation": aliased(Derivation, flat=True),
+            "used_derivation": aliased(Derivation, flat=True),
+        }
+
```
GianlucaFicarelli, line 94: Mutating the aliases dict seems a bit fragile, and unexpected by the caller.
An alternative could be to provide a function to build the aliases, that is called in each `read_many` instead of initializing the aliases dict directly.
However, it requires more changes and it would be easier to review if it's done in a separate PR (that can be also done after this one, I can have a look as well).

app/service/analysis_notebook_result.py
```diff
@@ -52,6 +53,7 @@ def read_one(
     user_context: UserContextDep,
     db: SessionDep,
     id_: uuid.UUID,
+    expand: ExpandDep = None,
```
GianlucaFicarelli, line 56: It's a bit annoying to add expand to each endpoint, but I don't see a cleaner way to avoid those repetitions 🤔

app/db/model.py
GianlucaFicarelli: I checked better the code and these properties are accessed even when `expand` is empty, so the guard is needed.

app/queries/common.py
pgetta: It already is - `apply_derivation_expand` early-returns unless `issubclass(db_model_class, Entity)` (and when expand is empty), so it's a no-op for non-entity models.

app/db/model.py
pgetta: Agreed, clearer. Renamed: `generated_derivations -> generated_from_derivations` and `used_derivations -> used_by_derivations` (properties, response fields, and the ?expand= values).

app/filters/base.py
pgetta: Good catch - these were effectively unreachable, I've removed the try/except block.

app/filters/base.py
pgetta: Should be fixed as well.

app/queries/factory.py
pgetta: Nice, I've narrowed the param to object.

app/filters/circuit.py
pgetta: Removed.

app/queries/factory.py
pgetta: Thank you, Gianluca, I'll note this down - can look a bit later (in case you don't address it in the meantime).

app/queries/factory.py
GianlucaFicarelli: I don't see advantages in using object instead of Any here.
`db_model_class` is defined as Any in the parameters of `query_params_factory`, so it should be narrowed there if possible, or left as Any in this PR.

pgetta:
Let me merge changes from `main` real quick.

app/queries/factory.py
pgetta: For now I've reverted that back to `Any`.

pgetta:
Ok, code coverage seems to be better now as well.

GianlucaFicarelli (commented):
LGTM, I'll let @eleftherioszisis to have a look at the last changes

eleftherioszisis (approved):
lgtm

pgetta:
Thank you all.

PR #650, opened 2026-06-26
Add `authorized_project_id` column to `task_activity`

dkeller9:
## Summary

Add `authorized_project_id` column to the `task_activity` table so that the generic `router_update_activity_one` authorization check can resolve the project ID for `TaskActivity` rows.

## Problem

When obi-one completes a `mesh_lod_generation` task and calls `PATCH /task-activity/{id}` with `generated_ids`, the request fails with a 500 Internal Server Error. The root cause is in `router_update_activity_one`:

```python
id_model_class = get_authorized_project_id_declaring_class(db_model_class)
if id_model_class is None:
    raise ApiError(..., http_status_code=HTTPStatus.INTERNAL_SERVER_ERROR)
```

`get_authorized_project_id_declaring_class` calls `has_project_id_in_columns`, which checks `db_model_class.__mapper__.columns` — only columns defined in the model's **own** table, not inherited ones. `TaskActivity`'s table only had `id` and `task_activity_type`; `authorized_project_id` lives in the parent `activity` table and was therefore invisible to this check, causing it to return `None` and raise the 500.

## Solution

- Add `authorized_project_id` as an explicit column on the `task_activity` table in `db/model.py`
- Add a migration that creates the column, adds an index, and back-fills existing rows from the parent `activity` table

## Changes

- `app/db/model.py` — add `authorized_project_id: Mapped[uuid.UUID | None]` to `TaskActivity`
- `alembic/versions/91eb89d1293e_add_authorized_project_id_to_task_activity.py` — new migration

## Testing

Verified locally by running the full alembic migration chain from a clean DB. End-to-end test: run a `mesh_lod_generation` task via obi-one against staging and confirm the final `PATCH /task-activity/{id}` returns 200 instead of 500.


eleftherioszisis:
> get_authorized_project_id_declaring_class calls has_project_id_in_columns, which checks db_model_class.__mapper__.columns — only columns defined in the model's own table, not inherited ones. 

Incorrect:
```python
(Pdb) db_model_class
<class 'app.db.model.TaskActivity'>
(Pdb) has_project_id_in_columns(db_model_class)
True
```

Can you please open an issue with the error so that we can look into it instead of proposing fixes that have not been verified first? Thanks.

PR #653, opened 2026-06-29
Allow users register publications

eleftherioszisis:
Issue: https://github.com/openbraininstitute/obi-one/issues/867


mgeplf (approved):

app/service/publication.py
```diff
@@ -62,7 +62,7 @@ def admin_read_one(db: SessionDep, id_: uuid.UUID) -> PublicationRead:
 def create_one(
     db: SessionDep,
     json_model: PublicationCreate,
-    user_context: AdminContextDep,
+    user_context: UserContextDep,
```
GianlucaFicarelli: It could be useful to add a comment or a link to explain why the create endpoint allows any user, while update and delete endpoints still require AdminContextDep, to prevent accidental changes in the future.

eleftherioszisis: Added link to issue

GianlucaFicarelli (approved):

PR #654, opened 2026-06-29
Add Person ORCID Organization ROR_ID

eleftherioszisis:
Issue: https://github.com/openbraininstitute/obi-one/issues/793


app/utils/pydantic_validators.py
```diff
@@ -0,0 +1,47 @@
+import re
+from typing import Annotated
+
+from pydantic import AfterValidator
+
+ORCID_REGEX = re.compile(r"^\d{4}-\d{4}-\d{4}-\d{4}$")
```
chr-pok: Looking at the specs, seems like valid ORCIDs may have an "X" character as well at the last position (checksum), see https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier.

Also, would be great if we could even check the checksum as additional after-validation.

app/utils/pydantic_validators.py
```diff
@@ -0,0 +1,47 @@
...
+ROR_REGEX = re.compile(r"^0[a-z0-9]{8}$")
```
chr-pok: According to the specs, the regular expression to validate should be: `^0[a-hj-km-np-tv-z|0-9]{6}[0-9]{2}$`, see https://ror.readme.io/docs/identifier.

The last two digits are also a checksum which could be validated I suppose.

app/db/model.py
```diff
@@ -333,6 +333,8 @@ class Person(Agent):
     family_name: Mapped[str | None]
     sub_id: Mapped[uuid.UUID | None] = mapped_column(unique=True, index=True)
 
+    orcid: Mapped[str] = mapped_column(String(19), nullable=True, unique=True, index=True)
```
GianlucaFicarelli: orcid defined as `Mapped[str]` but has `nullable=True`.
If it's nullable, it's enough to define it as
`Mapped[str | None]` and remove `nullable=True`

app/db/model.py
```diff
@@ -346,6 +348,13 @@ class Organization(Agent):
     # what is the difference between name and label here ?
     alternative_name: Mapped[str]
 
+    ror_id: Mapped[str] = mapped_column(
```
GianlucaFicarelli: ror_id is defined as `Mapped[str]` but has `nullable=True`.
If it's nullable, it's enough to define it as
`Mapped[str | None]` and remove `nullable=True`

app/utils/pydantic_validators.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def validate_orcid(value: str) -> str:
+    if value is None:
+        return value
```
GianlucaFicarelli: value cannot be None according to the type annotation, that seems correct because ORCID is validated as str.

app/utils/pydantic_validators.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return value
+
+
+ORCID = Annotated[str, AfterValidator(validate_orcid)]
+
+
+def validate_ror(value: str) -> str:
+    if value is None:
+        return value
```
GianlucaFicarelli: value cannot be None according to the type annotation, that seems correct because ROR_ID is validated as str.

jplanasc:
Other than the comments above, LGTM.

@chr-pok I can do the ORCID / ROR ID checks on obi-one side.

eleftherioszisis:
> Other than the comments above, LGTM.
> 
> @chr-pok I can do the ORCID / ROR ID checks on obi-one side.

Note that they will still be added here in the application layer.

chr-pok:
> @chr-pok I can do the ORCID / ROR ID checks on obi-one side.

Sure, I guess the earlier we check the better. But I suppose if the user provides a wrong ORCID / ROR ID, you won't be able to fetch any metadata anyway.

app/db/model.py
eleftherioszisis: done

app/db/model.py
eleftherioszisis: done

app/utils/pydantic_validators.py
eleftherioszisis: removed

app/utils/pydantic_validators.py
eleftherioszisis: removed

app/utils/pydantic_validators.py
eleftherioszisis: Fixed and added checksum validation

app/utils/pydantic_validators.py
eleftherioszisis: Fixed and added checksum validation

app/utils/pydantic_validators.py
GianlucaFicarelli: This is still unchanged

eleftherioszisis: Oh man. Sorry about that. Removed

GianlucaFicarelli (approved):

PR #664, opened 2026-07-10
Build entitycore container image for linux/arm64

danifr:
for https://github.com/openbraininstitute/INFRA/issues/600

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,6 +3,7 @@ name: Build and publish the Docker image
 permissions:
   id-token: write # Required for OIDC authentication
   contents: read # Standard permission for GitHub Actions
+  actions: write # Required for GHA build cache (cache-from/cache-to)
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -31,5 +30,23 @@ jobs:
...
 
-      - name: Build and publish the Docker image to AWS ECR
-        run: make publish
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Build and push
```
eleftherioszisis, line 39: Could this be refactored to use the Makefile and docker-compose? It is better to have one way to build things. For the cache I made similar changes in launch-system:
https://github.com/openbraininstitute/launch-system/blob/4403d67d4deb75597e5c46af1e70ef7d6ec2b834/docker-compose.run.yml#L15-L18

danifr: I thought of that too but the problem is we want to have multi-arch builds (x86 + arm64). 

AFAIK docker compose build + docker compose push doesn't support them. For that we will need to migrate to `docker buildx` in the Makefile. But that would break local builds.

launch-system only builds for x86 (but it is in my TODO to add support for multi-arch builds too)

eleftherioszisis: I see. ok, let's leave it as it is for now.

GianlucaFicarelli (approved):

PR #667, opened 2026-07-14
Add Sentry integration

pgetta:
## What

Adds a Sentry integration to entitycore, mirroring [openbraininstitute/accounting-service#53](https://github.com/openbraininstitute/accounting-service/pull/53) with entitycore's conventions. Uses `sentry-sdk>=2.65.0` (latest release).

## Changes

- **`app/sentry.py`** (new): `init_sentry()` calls `sentry_sdk.init(...)` with the DSN, environment, release (`APP_VERSION`), tracing + profiling sample rates, and `profile_lifecycle="trace"`.
- **`app/config.py`**: new settings, all env-var driven with defaults —
  - `DEPLOYMENT_ENV: Literal["local", "staging", "production"] = "local"` → Sentry `environment`
  - `SENTRY_DSN: str | None = None` (unset → nothing is sent)
  - `SENTRY_TRACES_SAMPLE_RATE` / `SENTRY_PROFILE_SESSION_SAMPLE_RATE` — both `Annotated[float, Field(ge=0, le=1)] = 1`
- **`app/__main__.py`**: calls `init_sentry()` after `configure_logging()` (so the loguru handler isn't stripped) and before the app is imported (so the FastAPI/Starlette auto-integrations attach).
- **`pyproject.toml` + `uv.lock`**: add `sentry-sdk>=2.65.0`.
- **Tests**: `tests/test_config.py` asserts out-of-range sample rates raise `ValidationError`; `tests/test_sentry.py` asserts `init_sentry()` produces an active client with `transport is None` (no DSN) and options matching settings.

With `SENTRY_DSN` unset (local dev, tests, CI) the SDK initializes as a no-op. Production supplies `SENTRY_DSN` / `DEPLOYMENT_ENV` / sample rates as runtime env vars; the DSN is deliberately kept out of the Docker build ARGs.

### DIFF
--- a/app/sentry.py
+++ b/app/sentry.py
@@ -1,0 +1,21 @@
+"""Sentry integration."""
+
+import sentry_sdk
+
+from app.config import settings
+
+
+def init_sentry() -> None:
+    """Initialize the Sentry SDK.
+
+    Must be called after ``configure_logging``, which removes all the loguru handlers,
+    including the ones registered by the Sentry loguru integration.
+    """
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        environment=settings.DEPLOYMENT_ENV,
+        release=settings.APP_VERSION,
+        traces_sample_rate=settings.SENTRY_TRACES_SAMPLE_RATE,
+        profile_session_sample_rate=settings.SENTRY_PROFILE_SESSION_SAMPLE_RATE,
+        profile_lifecycle="trace",
+    )
--- a/app/__main__.py
+++ b/app/__main__.py
@@ -3,6 +3,7 @@
 import click
 
 from app.logger import configure_logging, configure_warnings
+from app.sentry import init_sentry
 from app.utils.uvicorn import run_server
 
 
@@ -22,4 +23,5 @@ def run(*, host: str, port: int, reload: bool) -> None:
 
 configure_logging()
 configure_warnings()
+init_sentry()
 cli()
--- a/app/config.py
+++ b/app/config.py
@@ -34,7 +34,7 @@ class Settings(BaseSettings):
 
     SENTRY_DSN: str | None = None
     SENTRY_TRACES_SAMPLE_RATE: Annotated[float, Field(ge=0, le=1)] = 0.1
-    SENTRY_PROFILE_SESSION_SAMPLE_RATE: Annotated[float, Field(ge=0, le=1)] = 1
+    SENTRY_PROFILE_SESSION_SAMPLE_RATE: Annotated[float, Field(ge=0, le=1)] = 1.0
 
     ROOT_PATH: str = ""
     CORS_ORIGINS: list[str] = ["*"]
--- a/app/config.py
+++ b/app/config.py
@@ -33,8 +33,8 @@ class Settings(BaseSettings):
     DEPLOYMENT_ENV: Literal["local", "staging", "production"] = "local"
 
     SENTRY_DSN: str | None = None
-    SENTRY_TRACES_SAMPLE_RATE: Annotated[float, Field(ge=0, le=1)] = 1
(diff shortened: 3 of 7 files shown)


pgetta:
Relates to [Sentry integration for backend services](https://github.com/openbraininstitute/prod-platform-architecture/issues/220).

app/sentry.py
```diff
@@ -0,0 +1,21 @@
...
+    Must be called after ``configure_logging``, which removes all the loguru handlers,
+    including the ones registered by the Sentry loguru integration.
+    """
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        environment=settings.DEPLOYMENT_ENV,
+        release=settings.APP_VERSION,
+        traces_sample_rate=settings.SENTRY_TRACES_SAMPLE_RATE,
+        profile_session_sample_rate=settings.SENTRY_PROFILE_SESSION_SAMPLE_RATE,
+        profile_lifecycle="trace",
```
GianlucaFicarelli, line 20: Could you motivate the choice of the values for these parameters, in particular `traces_sample_rate` and `profile_session_sample_rate`?
How much overhead should we expect?
Is it ok to use the same values in staging and production, or should we differentiate them?

pgetta: I'm adding links to the documentation:

* [profile_session_sample_rate](https://docs.sentry.io/platforms/python/integrations/fastapi/#profile_session_sample_rate)
* [traces_sample_rate](https://docs.sentry.io/platforms/python/integrations/fastapi/#traces_sample_rate)

Some explanation from LLM:

### traces_sample_rate

This governs tracing / performance monitoring. It's a float from 0.0 to 1.0 representing the fraction of transactions captured. Setting it to 1.0 captures 100% of transactions; 0.1 would capture roughly 10%. This parameter needs to be set when initializing the SDK for performance measurements to be recorded at all—if you leave it unset, you get error monitoring but no traces. The sampling decision is made per transaction (each incoming request, in a FastAPI app), so it's re-evaluated continuously as traffic flows. (There's a sibling option, traces_sampler, that lets you compute the rate dynamically per-transaction instead of using a fixed number.)

### profile_session_sample_rate

This governs Continuous Profiling, the newer profiling model that replaced the old profiles_sample_rate / profiles_sampler options. It's also a float from 0.0 to 1.0, but the default is 0.0, meaning profiling is off unless you explicitly turn it on.

The critical difference from traces_sample_rate is the sampling scope. It samples whole sessions, not individual transactions. Sampling is evaluated once at the beginning of each session, not every time profiling data is collected. On a backend service, the session starts when the Sentry SDK is initialized and stops when the service or process is terminated, so sampling is only evaluated once during the process lifetime. In practice that means if you run N instances of a service (containers, pods, etc.), this rate controls the percentage of those instances that will have profiling enabled—not the percentage of requests. Because the scope is different, you can't directly translate an old profiles_sample_rate value into this one.

### How they interact in the FastAPI default config

The FastAPI snippet pairs profile_session_sample_rate with profile_lifecycle="trace". That combination is where the two rates connect. In trace mode, the profiler is automatically started and stopped based on the trace lifecycle: it starts when there's at least one root span/transaction active and stops when there are none—so you don't call any start/stop functions yourself. [Sentry Help Center](https://sentry.zendesk.com/hc/en-us/articles/36024865135131-Continuous-UI-Profiling-FAQ)

But trace-lifecycle profiling only fires when both gates open: the profile session must be sampled (via profile_session_sample_rate) and the trace must be sampled (via traces_sample_rate or traces_sampler). So with profile_lifecycle="trace", your traces_sample_rate effectively also gates profiling. If traces_sample_rate=0, no traces are sampled, so no profiling happens regardless of the profile rate. Setting both to 1.0, as the docs' example does, means every instance is profiled and profiling runs whenever a transaction is active.

GianlucaFicarelli (approved):

PR #678, opened 2026-07-22
Fix logger name misattribution by reading caller info from LogRecord

pgetta:
## Problem

Since the Sentry integration (#667), every log emitted by third-party libraries through stdlib `logging` (httpx, uvicorn, ...) is attributed to `sentry_sdk.integrations.logging` instead of the real caller:

```json
{"time":"...","level":"INFO","name":"sentry_sdk.integrations.logging","message":"HTTP Request: GET .../userinfo \"HTTP/1.1 200 OK\"","extra":{...}}
```

## Cause

sentry-sdk's default `LoggingIntegration` monkeypatches `logging.Logger.callHandlers`, inserting a `sentry_sdk` frame into the call stack. `InterceptHandler.emit` re-derives the caller by walking frames and skipping only frames from the `logging` module, so the walk now stops at Sentry's wrapper frame, and loguru derives `name`/`function`/`line` from it. Same upstream report: Delgan/loguru#509.

## Fix

Stop walking frames altogether. `logging.Logger._log` stamps `record.name`, `record.funcName`, and `record.lineno` via `findCaller()` **before** any handler — patched or not — runs, so `emit` now reads the caller off the `LogRecord` and applies it with `L.patch(...)`:

- Immune to any current or future patching of the handler chain (sentry-sdk, ddtrace, OpenTelemetry log instrumentation, ...), not just `sentry_sdk.*`.
- Removes a per-record Python-level frame walk from the hot path of every intercepted log line.
- No import-ordering constraint between `configure_logging()` and `init_sentry()`.

The request-context patcher from `configure_logging` is unaffected: loguru applies the core patcher first and `.patch()` composes on top of it.

## Tests

`tests/test_logger.py` adds regression coverage. Per review feedback, the tests do not call `sentry_sdk.init` (its `callHandlers` monkeypatch is global and never reverted, and sentry is already covered by `test_sentry.py`); instead, an autouse fixture wraps `logging.Logger.callHandlers` with an extra frame via pytest's `monkeypatch`, simulating exactly what sentry's `LoggingIntegration.setup_once` does. The simulated patch reproduces the misattribution under the old frame-walking implementation, so the tests fail on pre-fix code.

## Alternative to #677 (closed in favor of this PR)

#677 fixed the same bug by extending the frame walk to also skip `sentry_sdk.*` frames. The tradeoff that led to picking this PR:

- #677 kept the exact pre-#667 semantics (`name` = calling module's `__name__`, e.g. `httpx._client`) but hardcodes one library into the frame walk; any other library that wraps the handler chain reproduces the bug and needs another clause.
- This PR is robust by construction, but changes `name` to the stdlib **logger name** (e.g. `httpx` instead of `httpx._client`). The two coincide for the common `logging.getLogger(__name__)` idiom; the logger name is also the key used in `LOG_STANDARD_LOGGER` and log filtering. If you have saved log queries or alerts matching exact module-level `name` values, they may need updating.


pgetta:
@GianlucaFicarelli this is an alternative fix for the logging issue. Please let me know which one makes more sense to apply.

tests/test_logger.py
```diff
@@ -0,0 +1,42 @@
+import logging
+
+import sentry_sdk
+from loguru import logger
+
+from app.logger import InterceptHandler
+
+
+def _intercept_one_message(logger_name):
+    sentry_sdk.init(dsn=None)
```
GianlucaFicarelli: `sentry_sdk.init` has a global effect that could affect other tests, and it's called also in `test_sentry.py`.
Is there a recommended approach for running tests with sentry enabled?
If it should be always enabled, it can be set up with a session fixture in `conftest.py` and the invocation removed from `test_sentry.py`.
Alternatively, sentry should be tested in `test_sentry.py` only.
In this test, instead, it should be possible to mock `logging.Logger.callHandlers` to simulate sentry monkey-patch, without actually initializing the sentrysdk.

GianlucaFicarelli (commented):
Thanks Pavlo, this PR looks more generic than https://github.com/openbraininstitute/entitycore/pull/677 so I think we can go with this.
The new behaviour for name described in the PR description looks good to me, maybe even clearer than the previous one.
I just commented about the tests and the global effect of sentry initialization.

tests/test_logger.py
pgetta: Good point, went with the last option: an autouse fixture now wraps `logging.Logger.callHandlers` with `monkeypatch`, without initializing the SDK, so the patch is reverted after each test. It simulates exactly what sentry's `LoggingIntegration.setup_once` does - replacing `callHandlers` with a wrapper defined outside the `logging` module, and that extra frame is the whole trigger of the bug. I checked that the simulated patch still reproduces the misattribution with the old frame-walking implementation, so the test keeps failing on the pre-fix code. Sentry itself stays tested in `test_sentry.py` only.

GianlucaFicarelli (approved):

pgetta:
Thank you Gianluca

PR #679, opened 2026-07-22
Refactor asset router and service to extract reusable logic

GianlucaFicarelli:
This PR moves the remaining business logic for assets from routers to service.
It's needed for later adding the missing admin endpoints for assets in a separate PR, without duplicating code.


app/routers/admin.py
```diff
@@ -23,6 +22,7 @@
     asset as asset_service,
     publish as publish_service,
 )
+from app.types import EntityRoute
```
GianlucaFicarelli, line 25: app/routers/types.py has been moved to app/types.py to avoid circular imports, since it's imported in both routers and services.

eleftherioszisis (approved):

PR #687, opened 2026-07-28
Strict update schemas: forbid extras and merge exclusions

### LINKED ISSUE #614: Improve update schemas
* Do not allow to silently ignore excluded values. Let the user know they are updating forbidden attributes.
* Merge default keys in the make_update_schema instead of using them as defaults to avoid skipping the default keys by mistake

eleftherioszisis:
## Summary
- Update `make_update_schema` so non-empty `excluded_fields` merge with `DEFAULT_EXCLUDED_FIELDS`; `set()` still means no exclusions (admin).
- Generated update models use `extra="forbid"`, so unknown or excluded PATCH fields return 422 instead of being dropped.
- Cover the behavior in shared update helpers and derivation/unit tests.

Closes #614

## Test plan
- [x] `tests/test_schemas_utils.py` (merge / empty / forbid)
- [x] `tests/test_derivation.py::test_update_one`
- [x] `check_entity_update_one` / `check_global_update_one` via subject, species, circuit, measurement-label update tests
- [x] Full suite: `make test-docker` (or CI)

### DIFF
--- a/app/schemas/utils.py
+++ b/app/schemas/utils.py
@@ -1,6 +1,6 @@
 from typing import Annotated
 
-from pydantic import Field, create_model
+from pydantic import ConfigDict, Field, create_model
 
 from app.schemas.base import Schema
 
@@ -10,24 +10,40 @@ DEFAULT_EXCLUDED_FIELDS = {
 DEFAULT_PRESERVED_FIELDS = set()
 
 
+class UpdateSchema(Schema):
+    model_config = ConfigDict(from_attributes=True, extra="forbid")
+
+
 def make_update_schema(
     schema: type[Schema],
     new_schema_name: str | None = None,
-    excluded_fields: set = DEFAULT_EXCLUDED_FIELDS,
+    excluded_fields: set | None = None,
     preserved_fields: set = DEFAULT_PRESERVED_FIELDS,
 ):
     """Create a new pydantic schema from current schema where all fields are optional.
 
     When the api payload instantiates the schema model, the user can explicitly set an update value
     to None.
+
+    ``excluded_fields`` semantics:
+    - ``None`` (omitted): use ``DEFAULT_EXCLUDED_FIELDS``
+    - empty set: no exclusions (admin opt-out)
+    - non-empty set: ``DEFAULT_EXCLUDED_FIELDS | excluded_fields``
     """
 
     def make_optional(field):
         return Annotated[field.annotation | None, Field(default=None)]
 
+    if excluded_fields is None:
+        excluded = set(DEFAULT_EXCLUDED_FIELDS)
+    elif not excluded_fields:
+        excluded = set()
+    else:
+        excluded = DEFAULT_EXCLUDED_FIELDS | excluded_fields
+
     fields = {}
     for name, field in schema.model_fields.items():
-        if name in excluded_fields:
+        if name in excluded:
             continue
 
         if name in preserved_fields:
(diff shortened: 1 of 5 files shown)


mgeplf (approved):

app/schemas/utils.py
```diff
@@ -10,24 +10,40 @@
...
+    - ``None`` (omitted): use ``DEFAULT_EXCLUDED_FIELDS``
+    - empty set: no exclusions (admin opt-out)
+    - non-empty set: ``DEFAULT_EXCLUDED_FIELDS | excluded_fields``
     """
 
     def make_optional(field):
         return Annotated[field.annotation | None, Field(default=None)]
 
+    if excluded_fields is None:
+        excluded = set(DEFAULT_EXCLUDED_FIELDS)
```
GianlucaFicarelli: Minor, DEFAULT_EXCLUDED_FIELDS is already a set, the wrapping `set()` can be removed.

app/schemas/utils.py
```diff
@@ -10,24 +10,40 @@
...
+class UpdateSchema(Schema):
+    model_config = ConfigDict(from_attributes=True, extra="forbid")
+
+
 def make_update_schema(
     schema: type[Schema],
     new_schema_name: str | None = None,
-    excluded_fields: set = DEFAULT_EXCLUDED_FIELDS,
+    excluded_fields: set | None = None,
     preserved_fields: set = DEFAULT_PRESERVED_FIELDS,
```
GianlucaFicarelli, line 21: Unrelated to this PR, but could you add also an explanation for preserved_fields in the docstring?

app/schemas/utils.py
```diff
@@ -10,24 +10,40 @@
 DEFAULT_PRESERVED_FIELDS = set()
 
 
+class UpdateSchema(Schema):
+    model_config = ConfigDict(from_attributes=True, extra="forbid")
```
GianlucaFicarelli, line 14: Should entitysdk be updated as well to not pass extra fields in the update calls?

eleftherioszisis: I would prefer not duplicating the logic if possible. Why not just let the user send an update query and let the server say no?

app/schemas/utils.py
eleftherioszisis: done

app/schemas/utils.py
eleftherioszisis: done

app/schemas/utils.py
GianlucaFicarelli: What will happen if the user passes an entity to `client.update_entity()` [here](https://github.com/openbraininstitute/entitysdk/blob/fbb31dd7973e0c962c4ca74073841b5ee9858ccd/src/entitysdk/client.py#L320)?

eleftherioszisis: Needs to be changed. I'll check if there is a good way to conform with the strict schema or remove it given that the majority of users update fields directly.

GianlucaFicarelli: Looking for update_entity in the internal repos https://github.com/search?q=org%3Aopenbraininstitute+%22update_entity%22&type=code&p=1 gives a reasonable amount of matches that can be checked, in case any of those passes an entity.

eleftherioszisis: Indeed. There is only one notebook that seems to be passing an entity. The rest seem to be using the attrs route.

GianlucaFicarelli (approved):

PR #689, opened 2026-07-29
Add recording_arrays relationship to Simulation

### LINKED ISSUE #680: Relate a Simulation to the SimulatableExtracellularRecordingArray(s) it records with
One or multiple SimulatableExtracellularRecordingArray entities can be used by a Simulation

The Simulation should represent which SimulatableExtracellularRecordingArray entities it uses so that entitysdk.stage_simulation can appropriately stage the electrode_array_weight_matrix asset of these SimulatableExtracellularRecordingArray entities. 

Each array entity will be used in one or multiple sonata report blocks, meaning that we will also need a mechanism to make the correct correspondence . 

This information is in the obi-one simulation config (which could therefore be used in the staging function), but we could also put information into the stored simulation_config which helps us make this association, such as the entityid of the SimulatableExtracellularRecordingArray in the electrode_file entry of the sonata config

eleftherioszisis:
## Summary
- Adds a many-to-many `recording_arrays` relationship on `Simulation` linking to `SimulatableExtracellularRecordingArray` via a new junction table.
- Wires create/update schemas, nested association logic, service load path, and migration.
- Enables `entitysdk.stage_simulation` to discover and stage `electrode_array_weight_matrix` assets.

Closes #680

## Test plan
- [x] `test_create_one` verifies empty `recording_arrays` by default
- [x] `test_create_one_with_recording_arrays` verifies association with 2 arrays
- [x] All 56 simulation-related tests pass


james-isbister (commented):
LGTM, but I'll let @GianlucaFicarelli review the details

GianlucaFicarelli (approved):
LGTM

PR #695, opened 2026-07-31
Add circuit_simplification types and NEST target simulator

darshanmandge:
Add the following enum values needed by the obi-one circuit simplification task (PR #897):

TargetSimulator:
  - nest = 'NEST'

DerivationType:
  - circuit_simplification (with docstring explaining semantic meaning)

TaskActivityType:
  - circuit_simplification__config_generation
  - circuit_simplification__execution

TaskConfigType:
  - circuit_simplification__campaign
  - circuit_simplification__config


app/db/types.py
```diff
@@ -1307,6 +1314,7 @@ class TargetSimulator(StrEnum):
     brian2 = "Brian2"
     coreneuron = "CORENEURON"
     learning_engine = "LearningEngine"
+    nest = "NEST"
```
GianlucaFicarelli, line 1317: For confirmation, where is NEST defined?
The other values have a correspondent value in libsonata:
https://github.com/openbraininstitute/libsonata/blob/b4b0498087696dcbe7b5c31ce8ad276daddd5387/python/bindings.cpp#L748-L753
Does libsonata need to be updated as well?

darshanmandge: NEST simulations are not yet supported by the simulator/platform and will come later. This will probably need changes in libsonata (as you mentioned), simulator and obi-one. 

I followed the same pattern of adding new simualtor and made the circuit registration so that in future when the NEST simulations are possible, we can use it.

GianlucaFicarelli (approved):
LGTM

PR #696, opened 2026-08-03
Validate ElectricalRecordingStimulus has the same visibility of recording

### LINKED ISSUE #693: Validate that stimuli have the same project and public attributes as the recording when created
## Description
Validate that stimuli have the same project and public attributes as the recording when created.

It should be done for all the recording types (ion_channel_recording and electrical_cell_recording).

## Technical implementation & specs
It can be done in the app logic if simpler, or as a trigger in the db.

## Additional context
Related issue: https://github.com/openbraininstitute/prod-build-emodel/issues/209

## Acceptance criteria
- [ ] it's not possible to register a private stimulus for a public recording
- [ ] it's not possible to register a public stimulus for a private recording (this should be validated already)
- [ ] it's not possible to register a stimulus for a recording if the project isn't the same

GianlucaFicarelli:
Fix https://github.com/openbraininstitute/entitycore/issues/693


app/db/triggers.py
```diff
@@ -161,6 +161,69 @@ def unauthorized_private_reference_trigger(model: type[Entity], field_name: str)
...
+def _get_must_match_trigger_name(table: str, field_name: str) -> str:
+    name = _truncate_identifier(f"match_trg_{table}_{field_name}")
+    return _check_name_length(name)
+
+
+def must_match_reference_function(model: type[Entity], field_name: str) -> PGFunction:
+    """Return a PGFunction that checks the model's visibility mirrors the referenced entity.
+
+    Enforces that authorized_public and authorized_project_id on the new row are
+    identical to those of the referenced entity. Use this for containment relationships
```
mgeplf: `containment relationships` that's a new term to me, doesn't it say that the child must not have a NULL field for its parent?

mgeplf (approved):

app/db/triggers.py
GianlucaFicarelli: It means that the child entity (e.g. stimulus) is part of the other entity (e.g. recording) so it should share the same visibility (authorized_public and authorized_project_id).
Do you have any suggestion to make it clearer?
We could use just "Use this for relationships..."

mgeplf: > We could use just "Use this for relationships..."

Makes sense.

It may be my interpretation of `containment relationship` is off, but it seems to be a term that sort of applies, but isn't quite right?

mgeplf:
LGTM

PR #701, opened 2026-08-11
Allow pagination_request.page_size to be 0

GianlucaFicarelli:
Currently the UI calls the backend with parameters `with_facets=true` and `page_size=1` to get the facets once.
Since the data are not needed in this case, entitycore can relax the validation of `page_size` to be >= 0, and skip the execution of the data query completely.

A request with page_size=0 won't fail anymore and will always return an empty list for data.

Other: the code for retrieving data has been moved to `_retrieve_rows` without changes, except  for the addition of the check for page_size.

cc @bilalesi for the update of page_size parameter in the UI

### DIFF
--- a/app/queries/common.py
+++ b/app/queries/common.py
@@ -1,5 +1,5 @@
 import uuid
-from collections.abc import Set as AbstractSet
+from collections.abc import Iterable, Set as AbstractSet
 from http import HTTPStatus
 
 import sqlalchemy as sa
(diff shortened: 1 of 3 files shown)


mgeplf (approved):

bilalesi:
Cool!

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #703, opened 2026-08-12
Analysis Notebook Template unique name per project

g-bar:
Enforce that two `AnalysisNotebookTemplate` entities within the same project cannot share the same name.

For the Courses feature, each course has a master project and a synchronization mechanism. When a notebook is patched or deleted in the master project, the same operation is applied to the notebook with the same name in all child projects.

The initial suggestion was to use a derivation link from each child notebook to its corresponding master notebook. However, this is not possible because the notebooks are private entities belonging to different projects within the same virtual lab.

Instead, notebooks are matched by name. This requires notebook names to be unique within each project. Uniqueness is currently enforced only by the frontend, so backend validation is being added to enforce this constraint.

See original discussion here: https://github.com/openbraininstitute/prod-launch-notebook/issues/55#issuecomment-4622228788

**No database constraint**
`name` and `authorized_project_id` live in different tables (joined-table inheritance), so a DB-level unique constraint across them is not possible without denormalizing the schema.

## Changes

- `app/service/analysis_notebook_template.py`: `_check_unique_name` called in `create_one` and `update_one`, returns `409 ENTITY_DUPLICATED` on conflict.
- `tests/test_analysis_notebook_template.py`: tests for same-project conflict (create + update), cross-project allowed, and successful rename.

### DIFF
--- a/app/service/analysis_notebook_template.py
+++ b/app/service/analysis_notebook_template.py
@@ -1,18 +1,21 @@
 import uuid
+from http import HTTPStatus
 from typing import TYPE_CHECKING
 
 import sqlalchemy as sa
-from sqlalchemy.orm import aliased, joinedload, raiseload, selectinload
+from sqlalchemy.orm import Session, aliased, joinedload, raiseload, selectinload
 
 from app.db.model import (
     Agent,
     AnalysisNotebookTemplate,
     Contribution,
+    Entity,
     PlatformUser,
 )
 from app.dependencies.auth import AdminContextDep, UserContextDep, UserContextWithProjectIdDep
 from app.dependencies.common import ExpandDep, FacetsDep, PaginationQuery, SearchDep
 from app.dependencies.db import SessionDep
+from app.errors import ApiError, ApiErrorCode
 from app.filters.analysis_notebook_template import AnalysisNotebookTemplateFilterDep
 from app.queries.common import (
     router_create_one,
@@ -34,6 +37,24 @@ from app.schemas.types import ListResponse
 
 if TYPE_CHECKING:
     from app.filters.base import Aliases
+
+
+def _check_unique_name(
+    db: Session, name: str, project_id: uuid.UUID, exclude_id: uuid.UUID | None = None
+) -> None:
+    query = sa.select(AnalysisNotebookTemplate.id).where(
+        AnalysisNotebookTemplate.name == name,
+        Entity.authorized_project_id == project_id,
+        AnalysisNotebookTemplate.id == Entity.id,
+    )
+    if exclude_id is not None:
+        query = query.where(AnalysisNotebookTemplate.id != exclude_id)
+    if db.execute(query).first() is not None:
+        raise ApiError(
+            message="AnalysisNotebookTemplate already exists or breaks unique constraints",
+            error_code=ApiErrorCode.ENTITY_DUPLICATED,
+            http_status_code=HTTPStatus.CONFLICT,
+        )
 
 
 def _load(query: sa.Select):
@@ -85,6 +106,7 @@ def create_one(
     json_model: AnalysisNotebookTemplateCreate,
     user_context: UserContextWithProjectIdDep,
 ) -> AnalysisNotebookTemplateRead:
+    _check_unique_name(db, json_model.name, user_context.project_id)
     return router_create_one(
         db=db,
         json_model=json_model,
@@ -101,6 +123,8 @@ def update_one(
     id_: uuid.UUID,
     json_model: AnalysisNotebookTemplateUpdate,  # pyright: ignore [reportInvalidTypeForm]
 ) -> AnalysisNotebookTemplateRead:
+    if json_model.name is not None and user_context.project_id is not None:
+        _check_unique_name(db, json_model.name, user_context.project_id, exclude_id=id_)
     return router_update_one(
         id_=id_,
         db=db,
--- a/tests/test_analysis_notebook_template.py
+++ b/tests/test_analysis_notebook_template.py
@@ -65,10 +65,8 @@ def test_update_one(clients, json_data):
         admin_route=ADMIN_ROUTE,
         clients=clients,
         json_data=json_data,
-        patch_payload={
-            "name": "name",
-            "description": "description",
-        },
+        private_json_data=json_data | {"name": "other name"},
+        patch_payload={"description": "new description"},
         optional_payload=None,
     )
 
--- a/tests/test_analysis_notebook_template.py
+++ b/tests/test_analysis_notebook_template.py
@@ -2,6 +2,7 @@ import pytest
 
 from app.db.model import AnalysisNotebookTemplate
 from app.db.types import EntityType
+from app.errors import ApiErrorCode
 
 from .utils import (
     assert_request,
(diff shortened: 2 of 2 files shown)


eleftherioszisis:
Instead of adding that constraint on the service side, wouldn't it be preferable to add a composite unique constraint on name + authorized_project_id on the dB model layer for the analysis notebook?

app/service/analysis_notebook_template.py
```diff
@@ -101,6 +123,8 @@ def update_one(
     id_: uuid.UUID,
     json_model: AnalysisNotebookTemplateUpdate,  # pyright: ignore [reportInvalidTypeForm]
 ) -> AnalysisNotebookTemplateRead:
+    if json_model.name is not None and user_context.project_id is not None:
+        _check_unique_name(db, json_model.name, user_context.project_id, exclude_id=id_)
```
GianlucaFicarelli: Now that project_id is optional in update endpoints, this check can be bypassed by the user by not providing a project_id.
Since we already use triggers for other checks, I think it could be cleaner to create a trigger with a generic function in triggers.py. It would also have the advantage of being reusable in case this check becomes useful also for other entities.

g-bar:
> Instead of adding that constraint on the service side, wouldn't it be preferable to add a composite unique constraint on name + authorized_project_id on the dB model layer for the analysis notebook?

Ah yes! I initially wanted to avoid denormalizing the `authorized_project_id` but I think this can be done cleanly with a composite foreign key and the unique constraint you suggest. And that also avoids the application check or triggers.

```py
class Entity:
    id: UUID
    authorized_project_id: UUID

# Required by the composite FK
    __table_args__ = (
        UniqueConstraint('id', 'authorized_project_id', name='uq_entity_id_project'),
    )
    
```
    
  ```py  
   class UniqueNamePerProjectMixin(NameDescriptionVectorMixin):
        __abstract__ = True
        name: Mapped[str] = mapped_column(index=True, unique=True)

        # Denormalize the authorized_project_id
        authorized_project_id: Mapped[uuid.UUID] = mapped_column(Uuid)
        
        _table_args__ = (
        # Enforces the ids and authorized project_ids pairs are always consistent
        ForeignKeyConstraint(
            ['id', 'authorized_project_id'],
            ['entity.id', 'entity.authorized_project_id'],
            name='fk_model_entity_composite'
        ),
        
        # Enforces per project name unique constraint
        UniqueConstraint('authorized_project_id', 'name', name='uq_model_project_name'),
    )
```

Only minor downside is we have to duplicate of the authorized project id (we already do it for the id), and the composite FK ensures the pair (id, authorized_project_id) never differs between the parent `entity` and the child (eg analysis notebook template)

This would be easily reusable for other entity types, and would be completely transparent, we can continue to access `entity.id` and `entity.authorized_project_id` as before.

Thoughts? @eleftherioszisis @GianlucaFicarelli

File: app/service/analysis_notebook_template.py (hunk 1 above)
Comment (g-bar) [reply app/service/analysis_notebook_template.py]: Ah yeah, I think we can avoid the trigger and fully eliminate this check if we implement this:
https://github.com/openbraininstitute/entitycore/pull/703#issuecomment-5317491588

So Postgres takes care of the unique constrain check, and the router_create_one, router_update_one functions already take care of the auth checks.

**Comment by g-bar** [comment]
I went for the triggers, seems to be working well.

File: alembic/versions/20260821_134130_410a4832aa63_update_triggers.py [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "410a4832aa63"
+down_revision: Union[str, None] = "8953d8ad7437"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_table("_person_sub_id_mapping")
```
GianlucaFicarelli: This drop table shouldn't be executed, but it's caused by a previous migration and it's unrelated to the changes in this PR.
I can open a separate PR to tell alembic to ignore the tables starting with `_`.

alembic/versions/20260821_134130_410a4832aa63_update_triggers.py
```diff
@@ -0,0 +1,98 @@
...
+        schema="public",
+        signature="unique_name_fnc_analysis_notebook_template()",
+        definition="RETURNS TRIGGER AS $$\n            DECLARE\n                lock_key bigint;\n                project_id uuid;\n            BEGIN\n                SELECT authorized_project_id INTO project_id\n                FROM entity WHERE id = NEW.id;\n\n                lock_key := (\n                    ('x' || substring(md5(project_id::text || '\\:' || NEW.name), 1, 16))::bit(64)::bigint\n                ) >> 1;\n                PERFORM pg_advisory_xact_lock(lock_key);\n\n                IF EXISTS (\n                    SELECT 1 FROM entity e\n                    JOIN analysis_notebook_template t ON t.id = e.id\n                    WHERE e.authorized_project_id = project_id\n                    AND t.name = NEW.name\n                    AND t.id != NEW.id\n                ) THEN\n                    RAISE EXCEPTION 'duplicate name % in project %', NEW.name, project_id\n                        USING ERRCODE = '23505';\n                END IF;\n\n                RETURN NEW;\n            END;\n            $$ LANGUAGE plpgsql",
```
GianlucaFicarelli: In trigger.py here is a newline before `::bit(64)::bigint` that causes the function to be replaced if `make migration` is run again.

app/queries/common.py
```diff
@@ -478,29 +478,33 @@ def router_update_one[T: Schema, I: Identifiable](
...
+        create_associations_to_entities(
+            db=db,
+            parent=db_model_instance,
+            json_model=json_model,
+            nested_relationships=nested_relationships,
+            project_id=getattr(db_model_instance, "authorized_project_id", None),
+            action="update",
+        )
+
+        db.flush()
```
GianlucaFicarelli: Is there a reason to apply both the context managers `ensure_result` and `ensure_uniqueness` to the full block?

alembic/versions/20260821_134130_410a4832aa63_update_triggers.py
GianlucaFicarelli: Done in https://github.com/openbraininstitute/entitycore/pull/716

alembic/versions/20260821_134130_410a4832aa63_update_triggers.py
g-bar: Ah good catch fixed!

app/queries/common.py
g-bar: No good reason, changed to only wrap the relevant lines.

alembic/versions/20260821_134130_410a4832aa63_update_triggers.py
GianlucaFicarelli: The references to `_person_sub_id_mapping` need to be removed also from `downgrade`.

g-bar: Done

GianlucaFicarelli (approved):
LGTM, the trigger approach doesn't look too complex, it doesn't require changes to the table schemas and the read endpoints are unaffected.

